PR #4, opened 2017-11-21
add mirror info

paul-hammant:


koppor:
Is it also possible to enable "Issues" here? Seeing [teletype](https://teletype.atom.io/) getting more and more attraction, users will surely find the IntelliJ issue https://youtrack.jetbrains.com/issue/IDEABKL-708, which links to the saros project.

fzieris:
@koppor Yes, we will enable GitHub's issue feature soon. We currently prepare the migration of our [Sourceforge bug tracker](https://sourceforge.net/p/dpp/bugs).

koppor:
@fzieris Thank you for the backlink. For [JabRef](https://github.com/JabRef/jabref) we tried some sourceforge migration script and it looked good. However, a new repository was created. We also did not want to migrate all rusty bugs from https://sourceforge.net/p/jabref/bugs/. And thus decided against it.

m273d15:
As of now GitHub hosts the main repository. Therefore the mirror info is not needed.

PR #125, opened 2018-07-12
[FEATURE][CORE] Cancelable Input Streams

stefaus:
This patch introduces a class, wrapping Input Streams, to check for user
cancellations at each read access and adds this behavior to the
DecompressArchiveTask.

It shortens the response time, till user cancelations are proceded,
while decompressing of large files and removes an old FIXME.


stefaus:
@tobous You're right! In my initial version I extended BoundedInputStream, where this was enough. So I forgot to check this for InputStream...
Now added overrides for all methods to fix this.

PR #126, opened 2018-07-12
[FEATURE][CORE] Negotiate Transfer Type Instant

stefaus:
In preparation for the new feature 'Instant Session Start', this patch
introduces a new Transfer Type INSTANT and extends the Project
Negotiation Type Hook to negotiate according to the User Preferences.

While users can already select the feature, it has no effect and Project
Negotiation defaults to ARCHIVE.

### DIFF
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/TransferType.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/TransferType.java
@@ -7,5 +7,5 @@ import com.thoughtworks.xstream.annotations.XStreamAlias;
  */
 @XStreamAlias("TT")
 public enum TransferType {
-    ARCHIVE
+    ARCHIVE, INSTANT
 }
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/NegotiationFactory.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/NegotiationFactory.java
@@ -119,13 +119,16 @@ public final class NegotiationFactory {
         }
 
         switch (transferType) {
+        case INSTANT:
+            /* $FALL-THROUGH$ - implemented in follow-up patch */
         case ARCHIVE:
-            return new ArchiveOutgoingProjectNegotiation(remoteAddress, resources,
-                sessionManager, session, /* editorManager */
+            return new ArchiveOutgoingProjectNegotiation(remoteAddress,
+                resources, sessionManager, session, /* editorManager */
                 context.getComponent(IEditorManager.class), workspace,
                 checksumCache, connectionService, transmitter, receiver);
         default:
-            throw new UnsupportedOperationException("transferType not implemented");
+            throw new UnsupportedOperationException(
+                "transferType not implemented");
         }
     }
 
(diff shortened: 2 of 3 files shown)


de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/TransferType.java
```diff
@@ -7,5 +7,5 @@
  */
 @XStreamAlias("TT")
 public enum TransferType {
-    ARCHIVE
-}
\ No newline at end of file
+    ARCHIVE, INSTANT
```
Drakulix: Similar to my comment in #128, I am not sure the name is fitting. Theoretically also the archive transfer could be made 'instant' by queuing activities. And you implementation is currently not instant/non-blocking. I would prefer a different name. But that does not need to block this PR.

stefaus: > Theoretically also the archive transfer could be made 'instant' by queuing activities.

How do you mean that?

From my point of view, what happens at Archive mode is:
- create a Session, file list, ...
- block the whole session
- create a archive
- unblock and resume the session
- start sending actives to the new participant
- send the archive file (meanwhile the new participant queues incoming activities)
- after receiving the whole archive and unpack, apply queued activities
- done, user can open the first file

What happens at Instant mode is:
- create a Session, file list, ...
- for now: block the session
- send file by file (and later activities for every file already transmitted/available)
- participant receives file by file and can directly open them (and later only queues activities if file is not yet transmitted)
- unblock

> And you implementation is currently not instant/non-blocking.

To give a practical value, in my Benchmark environment (Bandwidth and Latency limited), it takes 34 seconds for a complete Project transfer in both modes, but only 4 seconds till the first open file in Instant mode, measured after the client choose which project to use.

On the road map for appending patches is:
- remove edit lock for existing participants
- afterwards remove for new participant

> I would prefer a different name.

I think 'Stream', like you mentioned in #128 wouldn't correspond to the complete feature which was already named before my work as 'Instant Session Start'.
Nethertheless I would prefer just one word across all Parts like Setting, UI, TransferType, ProjectNegotiation and I think Instant fits best.

stefaus:
@fzieris @Drakulix please reapprove, no changes in the patch itself, just rebase and merge of master.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/TransferType.java
Drakulix: > How do you mean that?

What I view as 'instant' is the fact, that the UI will not be blocked. Theoretically this could be achieved by queuing for the archive transfer type as well, although it is not very well suited indeed.

> I think 'Stream', like you mentioned in #128 wouldn't correspond to the complete feature which was already named before my work as 'Instant Session Start'.

Yes but the feature was mostly referring to getting rid of the blocking property of the archive transfer type. It was obvious, that this would require a new transfer type as well to be useful.

Whats my biggest problem is that one type's name now does describe a mechanism (sort of), while the other does describe a property of the connection. Both are descriptive, but not in the same way.

stefaus: > What I view as 'instant' is the fact, that the UI will not be blocked.

Ok, from my Point of View, according to [Saros Ongoing Work Page](http://www.saros-project.org/ongoing-work#dormant_work),
> Instead of starting a Saros session with the transfer of a ZIP file  containing the baseline for the following activities, a Session could  start empty and be filled file-by-file (transferred through activities)  so developers can start working on their code almost immediately.

this is not the main reason for this Feature, but indeed a byproduct, which is part of the next related patch.

>[...] one type's name now does describe a mechanism (sort of), while the other does describe a property of the connection. 

I think Instant describes that a participant receives Project Files file-by-file. Like Archive is hiding the Zip Mechanism, Instant is hiding if it would be by Stream or Activities.
And if the feature would be about the Blocking Part should be named Nonblocking.

Drakulix: > Like Archive is hiding the Zip Mechanism, Instant is hiding if it would be by Stream or Activities.

Alright. I do not feel like `Instant` is the best way to express this, but `FileByFile` or something along those lines would be too long and I cannot think of a better way.

PR #130, opened 2018-07-16
[BUILD] Enable travis notifications on failure

m273d15:
Activate that Travis CI sends mails if a build fails


m273d15:
@tobous Changed the config as proposed: "Only send mail if the build fails"

PR #132, opened 2018-07-16
[BUILD] Deploy master branch nightly to GitHub Releases

m273d15:
* The script release.sh provides functions to
  create/remove releases and git tags. This is
  implemented via the GitHub Release API.
* The script release_daily.sh uses this functions
  to create a (pre)release master-daily that currently
  contains only the intellij zip. If a release with
  the name master-daily already exists, the existing
  release will be removed.
* In order to release from another branch, add the branch
  name to the list of allowed branches in "deploy.on.condition"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,11 @@ git:
 
 services: docker
 
-# Sonar token is encrypted via travis encrypt
-addons:
-  sonarcloud:
-    organization: "saros"
-    token:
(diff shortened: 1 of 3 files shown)


m273d15:
See [my fork](https://github.com/m273d15/saros/releases/tag/master-daily) to see an example release that is created with this change.

m273d15:
Re-open if required.

PR #136, opened 2018-07-16
[UI][I] Remove contact form address from messages

tobous:
The contact form on out website saros-project.org is no longer
activated. Add contact should be done through our mailing lists.

Also removes the separate devel mailing list address as it is currently
not used.


m273d15:
I checked that the removed properties are not still referenced in code.
LGTM

PR #147, opened 2018-07-17
[FEATURE][SERVER] Add ActivityExecutors

Drakulix:
Applies Activities onto the local filesystem of the Server. (Finally does not trigger the `ConsistencyWatchdog` anymore).

### DIFF
--- a/de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/FileActivityExecutor.java
+++ b/de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/FileActivityExecutor.java
@@ -1,0 +1,102 @@
+package de.fu_berlin.inf.dpp.server.session;
+
+import de.fu_berlin.inf.dpp.activities.FileActivity;
+import de.fu_berlin.inf.dpp.activities.SPath;
+import de.fu_berlin.inf.dpp.filesystem.IFile;
+import de.fu_berlin.inf.dpp.filesystem.IResource;
+import de.fu_berlin.inf.dpp.server.editor.ServerEditorManager;
+import de.fu_berlin.inf.dpp.session.AbstractActivityConsumer;
+import de.fu_berlin.inf.dpp.session.ISarosSession;
+import org.apache.log4j.Logger;
+import org.picocontainer.Startable;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+
+/**
+ * Executes FileActivities and performs actual filesystem operations.
+ */
+public class FileActivityExecutor extends AbstractActivityConsumer implements
+    Startable {
+
+    private static final Logger LOG = Logger
+        .getLogger(FileActivityExecutor.class);
+
+    private final ISarosSession session;
+    private final ServerEditorManager editorManager;
+
+    /**
+     * Creates a FileActivityExecutor.
+     *
+     * @param session
+     *            the current session
+     * @param editorManager
+     *            the editor manager to update the file mapping
+     *            on a file move
+     */
+    public FileActivityExecutor(ISarosSession session,
+        ServerEditorManager editorManager) {
+
+        this.session = session;
+        this.editorManager = editorManager;
+    }
+
+    @Override
+    public void start() {
+        session.addActivityConsumer(this, Priority.ACTIVE);
+    }
+
+    @Override
+    public void stop() {
+        session.removeActivityConsumer(this);
+    }
+
+    @Override
+    public void receive(FileActivity activity) {
+        try {
+            switch (activity.getType()) {
+            case CREATED:
+                executeFileCreation(activity);
+                break;
+            case MOVED:
+                executeFileMove(activity);
+                break;
+            case REMOVED:
+                executeFileRemoval(activity);
+                break;
+            default:
+                LOG.warn("Unknown file activity type " + activity.getType());
+                break;
+            }
+        } catch (IOException e) {
+            LOG.error("Could not execute "+activity, e);
+        }
+    }
+
+    private void executeFileCreation(FileActivity activity) throws IOException {
+        IFile file = activity.getPath().getFile();
+        file.create(new ByteArrayInputStream(activity.getContent()), true);
+    }
+
+    private void executeFileMove(FileActivity activity) throws IOException {
+        SPath oldPath = activity.getOldPath();
+        IFile oldFile = oldPath.getFile();
+        SPath newPath = activity.getPath();
+        IFile newFile = newPath.getFile();
+        oldFile.move(activity.getPath().getFullPath(), true);
+        byte[] content = activity.getContent();
+        if (content != null) {
+            newFile.setContents(new ByteArrayInputStream(content),
+                true, true);
+        }
+        // only update if all previous operations are successful 
+        editorManager.updateMapping(oldPath, newPath);
+    }
+
+    private void executeFileRemoval(FileActivity activity) throws IOException {
+        SPath path = activity.getPath();
+        IFile file = path.getFile();
+        editorManager.closeEditor(path);
+        file.delete(IResource.NONE);
+    }
+}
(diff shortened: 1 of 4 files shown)


de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/FileActivityExecutor.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private void executeFileMove(FileActivity activity) throws IOException {
+        SPath oldPath = activity.getOldPath();
+        IFile oldFile = oldPath.getFile();
+        SPath newPath = activity.getPath();
+        IFile newFile = newPath.getFile();
+        editorManager.moveFile(oldPath, newPath);
+        oldFile.move(activity.getPath().getFullPath(), true);
+        newFile.setContents(new ByteArrayInputStream(activity.getContent()), true, true);
```
srossbach: getContent can be null if the file was only moved

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/FileActivityExecutor.java
```diff
@@ -0,0 +1,91 @@
...
+import de.fu_berlin.inf.dpp.server.editor.ServerEditorManager;
+import de.fu_berlin.inf.dpp.session.AbstractActivityConsumer;
+import de.fu_berlin.inf.dpp.session.ISarosSession;
+import org.apache.log4j.Logger;
+import org.picocontainer.Startable;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+
+public class FileActivityExecutor extends AbstractActivityConsumer implements
```
srossbach: A little description what this class is supposed to do would be nice.

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/FileActivityExecutor.java
```diff
@@ -0,0 +1,91 @@
...
+    Startable {
+
+    private static final Logger LOG = Logger
+        .getLogger(FileActivityExecutor.class);
+
+    ISarosSession session;
+    ServerEditorManager editorManager;
```
srossbach: Any reason the variables should be package private instead of just private ?

Drakulix:
@srossbach Anymore issues with this PR?

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/TextEditActivityExecutor.java
```diff
@@ -0,0 +1,45 @@
...
+import org.picocontainer.Startable;
+
+/**
+ * Executes TextEditActivities through the {@link ServerEditorManager}.
+ */
+public class TextEditActivityExecutor extends AbstractActivityConsumer implements
+    Startable {
+
+    ISarosSession session;
+    ServerEditorManager editorManager;
```
srossbach: private final

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/FolderActivityExecutor.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Executes FolderActivities and performs actual filesystem operations.
+ */
+public class FolderActivityExecutor extends AbstractActivityConsumer implements
+    Startable {
+
+    private static final Logger LOG = Logger
+        .getLogger(FolderActivityExecutor.class);
+
+    private ISarosSession session;
```
srossbach: final

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/FolderActivityExecutor.java
```diff
@@ -0,0 +1,84 @@
...
+    private ServerEditorManager editorManager;
```
srossbach: final

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/TextEditActivityExecutor.java
```diff
@@ -0,0 +1,45 @@
...
+        session.addActivityConsumer(this, Priority.ACTIVE);
+    }
+
+    @Override public void stop() {
+        session.removeActivityConsumer(this);
+    }
+
+    @Override
+    public void receive(TextEditActivity activity) {
+        editorManager.applyTextEdit(activity);
```
srossbach: much boiler code for actually 1 line of actual logic. And the worst part is that that consumer is a complete different package (I know that there is no common rule where classes should be placed).

Why not move this code directly to the EditorManager ?

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/FolderActivityExecutor.java
```diff
@@ -0,0 +1,84 @@
...
+        IFolder folder = activity.getPath().getFolder();
+        folder.create(IResource.NONE, true);
+    }
+
+    private void executeFolderRemoval(FolderDeletedActivity activity)
+        throws IOException {
+
+        SPath path = activity.getPath();
+        IFolder folder = path.getFolder();
+        this.editorManager.closeEditorsInFolder(path);
```
srossbach: this keyword is unnecessary

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/FileActivityExecutor.java
```diff
@@ -0,0 +1,98 @@
...
+    private ISarosSession session;
+    private ServerEditorManager editorManager;
+
+    /**
+     * Creates a FileActivityExecutor.
+     *
+     * @param session
+     *            the current session
+     * @param editorManager
+     *            the editor manager
```
srossbach: Not very helpful.

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/TextEditActivityExecutor.java
Drakulix: In that case the `ServerEditorManager` would also need to extend the `AbstractActivityConsumer` or are you talking about changing the core interface?

Either way this is how it is handled in all other plugins and the `EditorManager` is only involved, because ultimately the Editor is responsible for applying the text edit, but you cannot let each editor listen for activities easily. I feel this is the cleanest way to split the code and responsibilities, although this indeed looks like boilerplate.

All you other comments should be resolved, please check again. :)

srossbach: https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/editor/EditorManager.java#L161 

?

Drakulix: Okay, you are absolutely right, my apologies.
However this would also involve quite some boilerplate, as I would need to setup an `ISessionLifetimeListener` to add and remove the `IActivityExecutor` (https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/editor/EditorManager.java#L356).

Looks like almost the same amount of code to me and not really worth for just one type of Activity. Especially because I think this version is easier to understand.

If you insist, I will change it, but you approved the PR, so I assume this is not that important, right?

Drakulix:
> I already approved your changes but for the server part you do not need a Lifecycle listener. It has an historical reason why the Saros/E EditorManager uses this pattern because its lifetime is bound to the application although it should be bound to the session.

Another PR I have in preparation just starts a session in the beginning of the main function as the server is supposed to only handle one session anyway. As a result the lifetime of the session equals the lifetime of the program and therefor the `EditorManager` in the special case of the server.

I was assuming I would not need to handle this given the special circumstances. In testing this seems to work out okay, or do you see any other problems with this approach?

Issue #98, opened 2018-06-07
Core Test UPnPTest#testDispose() Fails if Assertions are Enabled


tobous:
When running the Saros core test suite with assertions enabled, the test UPnPTest#testDispose() fails with an assertion error:
<pre>
java.lang.AssertionError
	at de.fu_berlin.inf.dpp.net.upnp.internal.UPnPServiceImpl.createPortMapping(UPnPServiceImpl.java:168)
	at de.fu_berlin.inf.dpp.net.UPnPTest.testDispose(UPnPTest.java:163)
</pre>

### Steps To Reproduce:
1. Enable assertions (include "-EA" in your vm options for the core test runconfig)
2. Run the core test suite (de.fu_berlin.inf.dpp.SarosCoreTestSuite)

m273d15:
Enable assertions during build process after this issue is fixed.

PR #154, opened 2018-07-23
[FEATURE][CORE] Replaceable Queue Mechanism

stefaus:
This patch extracts a Interface out of the project based ActivityQueuer,
used on client-side, which handle activities in SarosSession, during the
project negotiation.
This moves the queuing responsibility into the transfer implementation
and is the preperation for a resource based queuing implementation.

### DIFF
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractIncomingProjectNegotiation.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractIncomingProjectNegotiation.java
@@ -213,13 +213,7 @@ public abstract class AbstractIncomingProjectNegotiation extends ProjectNegotiat
   protected void cleanup(IProgressMonitor monitor, Map<String, IProject> projectMapping) {
     fileReplacementInProgressObservable.replacementDone();
 
-    /*
-     * TODO Queuing responsibility should be moved to Project
-     * Negotiation, since its the only consumer of queuing
-     * functionality. This will enable a specific Queuing mechanism per
-     * TransferType (see github issue #137).
-     */
-    for (IProject project : projectMapping.values()) session.disableQueuing(project);
+    session.flushQueue();
 
     if (fileTransferManager != null)
       fileTransferManager.removeFileTransferListener(transferListener);
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ArchiveIncomingProjectNegotiation.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ArchiveIncomingProjectNegotiation.java
@@ -17,6 +17,7 @@ import de.fu_berlin.inf.dpp.net.xmpp.XMPPConnectionService;
 import de.fu_berlin.inf.dpp.observables.FileReplacementInProgressObservable;
 import de.fu_berlin.inf.dpp.session.ISarosSession;
 import de.fu_berlin.inf.dpp.session.ISarosSessionManager;
+import de.fu_berlin.inf.dpp.session.internal.ActivityQueuer;
 import de.fu_berlin.inf.dpp.util.CoreUtils;
 import java.io.File;
 import java.io.IOException;
@@ -35,6 +36,9 @@ import org.jivesoftware.smackx.filetransfer.IncomingFileTransfer;
 public class ArchiveIncomingProjectNegotiation extends AbstractIncomingProjectNegotiation {
 
   private static final Logger LOG = Logger.getLogger(ArchiveIncomingProjectNegotiation.class);
+
+  /** the one queuer used for all incoming project negotiations */
+  private static final ActivityQueuer activityQueuer = new ActivityQueuer();
 
   public ArchiveIncomingProjectNegotiation(
       final JID peer, //
@@ -62,6 +66,9 @@ public class ArchiveIncomingProjectNegotiation extends AbstractIncomingProjectNe
         connectionService,
         transmitter,
         receiver);
+
+    // Always registers the same, basically a NOP after the first call
+    session.registerActivityQueuer(activityQueuer);
   }
 
   @Override
@@ -80,14 +87,9 @@ public class ArchiveIncomingProjectNegotiation extends AbstractIncomingProjectNe
 
       final String projectID = entry.getKey();
       final IProject project = entry.getValue();
-      /*
-       * TODO Queuing responsibility should be moved to Project
-       * Negotiation, since its the only consumer of queuing
-       * functionality. This will enable a specific Queuing mechanism per
-       * TransferType (see github issue #137).
-       */
+
       session.addProjectMapping(projectID, project);
-      session.enableQueuing(project);
+      activityQueuer.enableQueuing(project);
     }
 
     transmitter.send(
@@ -106,6 +108,14 @@ public class ArchiveIncomingProjectNegotiation extends AbstractIncomingProjectNe
     if (filesMissing) {
       receiveAndUnpackArchive(projectMapping, transferListener, monitor);
     }
+  }
+
+  @Override
+  protected void cleanup(IProgressMonitor monitor, Map<String, IProject> projectMapping) {
+
+    for (IProject project : projectMapping.values()) activityQueuer.disableQueuing(project);
+
+    super.cleanup(monitor, projectMapping);
   }
 
   /** Receives the archive with all missing files and unpacks it. */
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/InstantIncomingProjectNegotiation.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/InstantIncomingProjectNegotiation.java
@@ -16,6 +16,7 @@ import de.fu_berlin.inf.dpp.net.xmpp.XMPPConnectionService;
 import de.fu_berlin.inf.dpp.observables.FileReplacementInProgressObservable;
 import de.fu_berlin.inf.dpp.session.ISarosSession;
 import de.fu_berlin.inf.dpp.session.ISarosSessionManager;
+import de.fu_berlin.inf.dpp.session.internal.ActivityQueuer;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.List;
@@ -30,6 +31,13 @@ import org.jivesoftware.smackx.filetransfer.IncomingFileTransfer;
 public class InstantIncomingProjectNegotiation extends AbstractIncomingProjectNegotiation {
 
   private static final Logger log = Logger.getLogger(InstantIncomingProjectNegotiation.class);
+
+  /**
+   * the one queuer used for all incoming project negotiations
+   *
+   * <p>TODO change to resource based queuer
+   */
+  private static final ActivityQueuer activityQueuer = new ActivityQueuer();
 
   public InstantIncomingProjectNegotiation(
       final JID peer, //
@@ -57,6 +65,9 @@ public class InstantIncomingProjectNegotiation extends AbstractIncomingProjectNe
         connectionService,
         transmitter,
         receiver);
+
+    // Always registers the same, basically a NOP after the first call
+    session.registerActivityQueuer(activityQueuer);
   }
 
   @Override
@@ -76,7 +87,7 @@ public class InstantIncomingProjectNegotiation extends AbstractIncomingProjectNe
 
       session.addProjectMapping(projectID, project);
       /* TODO change queuing to resource based queuing */
-      session.enableQueuing(project);
+      activityQueuer.enableQueuing(project);
     }
 
     transmitter.send(
@@ -93,6 +104,14 @@ public class InstantIncomingProjectNegotiation extends AbstractIncomingProjectNe
     for (FileList list : missingFiles) filesMissing += list.getPaths().size();
 
     if (filesMissing > 0) receiveStream(monitor, filesMissing);
+  }
+
+  @Override
+  protected void cleanup(IProgressMonitor monitor, Map<String, IProject> projectMapping) {
+
+    for (IProject project : projectMapping.values()) activityQueuer.disableQueuing(project);
+
+    super.cleanup(monitor, projectMapping);
   }
 
   private void receiveStream(IProgressMonitor monitor, int fileCount)
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/IActivityQueuer.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/IActivityQueuer.java
@@ -1,0 +1,19 @@
+package de.fu_berlin.inf.dpp.session;
+
+import de.fu_berlin.inf.dpp.activities.IActivity;
+import de.fu_berlin.inf.dpp.activities.IResourceActivity;
+import java.util.List;
+
+/** Interface for queuing incoming {@linkplain IActivity activities}. */
+public interface IActivityQueuer {
+
+  /**
+   * Processes the incoming {@linkplain IActivity activities} and decides which {@linkplain
+   * IResourceActivity resource related activities} should be queued.
+   *
+   * @param activities
+   * @return dequeued {@linkplain IResourceActivity IResourceActivities} and not queued {@linkplain
+   *     IActivity IActivities} from input {@code activities} list
+   */
+  public List<IActivity> process(final List<IActivity> activities);
+}
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ISarosSession.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ISarosSession.java
@@ -387,26 +387,15 @@ public interface ISarosSession {
   public void changeColor(int colorID);
 
   /**
-   * FOR INTERNAL USE ONLY !
+   * Register a activity queuer. Effective only at first call as all following calls are ignored.
    *
-   * <p>Starts queuing of incoming {@linkplain IResourceActivity project-related activities}, since
-   * they cannot be applied before their corresponding project is received and extracted.
-   *
-   * <p>That queuing relies on an existing project-to-projectID mapping (see {@link
-   * #addProjectMapping(String, IProject)}), otherwise incoming activities cannot be queued and will
-   * be lost.
-   *
-   * @param project the project for which project-related activities should be queued
-   * @see #disableQueuing
+   * @param activityQueuer
+   * @return true if the activityQueuer was registered successfully
    */
-  public void enableQueuing(IProject project);
+  public boolean registerActivityQueuer(IActivityQueuer activityQueuer);
 
-  /**
-   * FOR INTERNAL USE ONLY !
-   *
-   * <p>Disables queuing for the given project and flushes all queued activities.
-   */
-  public void disableQueuing(IProject project);
+  /** Send a NOPActivity to trigger queue processing. */
+  public void flushQueue();
 
   /**
    * Returns the id of the current session.
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/SarosSession.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/SarosSession.java
@@ -46,6 +46,7 @@ import de.fu_berlin.inf.dpp.session.IActivityConsumer.Priority;
 import de.fu_berlin.inf.dpp.session.IActivityHandlerCallback;
 import de.fu_berlin.inf.dpp.session.IActivityListener;
 import de.fu_berlin.inf.dpp.session.IActivityProducer;
+import de.fu_berlin.inf.dpp.session.IActivityQueuer;
 import de.fu_berlin.inf.dpp.session.ISarosSession;
 import de.fu_berlin.inf.dpp.session.ISarosSessionContextFactory;
 import de.fu_berlin.inf.dpp.session.ISessionListener;
@@ -66,6 +67,7 @@ import java.util.Set;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.log4j.Logger;
 import org.picocontainer.MutablePicoContainer;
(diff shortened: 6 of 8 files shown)


de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ArchiveIncomingProjectNegotiation.java
```diff
@@ -77,14 +82,9 @@ protected void transfer(IProgressMonitor monitor,
 
             final String projectID = entry.getKey();
             final IProject project = entry.getValue();
-            /*
-             * TODO Move enable (and disable) queuing responsibility to
-             * SarosSession, since the second call relies on the first one, and
-             * the first one is never done without the second. (see also TODO in
-             * {@link AbstractIncomingProjectNegotiation#cleanup}).
-             */
```
fzieris: Why did you delete this TODO?

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ArchiveIncomingProjectNegotiation.java
```diff
@@ -40,6 +41,8 @@
     private static final Logger LOG = Logger
         .getLogger(ArchiveIncomingProjectNegotiation.class);
 
+    private final ActivityQueuer activityQueuer = new ActivityQueuer();
```
fzieris: This changes many things at once, and I'm skeptic about them:

1. There is not one ActivityQueuer instance, but new queuers are created for each IPN.
2. Instead of better hiding the ActivityQueuer in the SarosSession (see deleted TODO below), the queuing becomes a shared responsibility of SarosSession and IPN.
3. There is an abstract super class, yet two implementations do very similar things with their respective queuers, which was before mostly part of the super class.

What are your reasons behind these changes?

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ArchiveIncomingProjectNegotiation.java
stefaus: answer in comment of line 44

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ArchiveIncomingProjectNegotiation.java
stefaus: In general, because my changes do the opposite of the old TODOs and move queuing into the negotiation process.
In #156 I created a ResourceActivityQueuer which is called all the time from the incoming negotiation class.

regarding 3: I could have moved it to the abstract and than move it in the next patch and I know this patch would be cleaner this way, but it would have created new diff lines in the next patch, more work in rethinking and probably errors.
Thats the problem with stacked PRs..., I created a big prototype and just splitted what is somehow independent from each other.

regarding 1: I create new Queues for every IPN, because they could change when the next incoming negotiation uses a different type. I could make them static but didn't thought more about this and don't know if/when the class loader would drop the class and GC would take the queue.

regarding 2: For the resource queuing, I need to call queuing with SPaths as parameter. So instead of adding new methods to SarosSession, I decided to make it part of the only process which uses it and keep the declared methods amount of ISarosSession the same. This thereby makes it easier to add new functionality to the queuing implementation.

fzieris: Alright, I'm fine with that.

(Although I still find the way GitHub dispalys Pull Requests confusing, and all the more "stacked" Pull Requests. It's hard to get an overview.)

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/SarosSession.java
```diff
@@ -1039,13 +1040,12 @@ public void changeColor(int colorID) {
     }
 
     @Override
-    public void enableQueuing(IProject project) {
-        activityQueuer.enableQueuing(project);
+    public void registerQueuingHandler(IActivityQueuer activityQueuer) {
+        this.activityQueuer = activityQueuer;
```
tobous: Check if activityQueuer is null?

I am not sure how to handle this in general. Our coderules don't specify if we generally expect the caller or the called methods to do the null checks. I generally prefer doing it in the called method, but without annotations for these kinds of checks, it adds a lot of "unnecessary" bulk to the code. ("Unnecessary" in the sens that it is not important for the reader as it usually just throws an IllegalArgumentException.)

@fzieris Any input on how to handle null checks for passed arguments in general?

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/IActivityQueuer.java
```diff
@@ -0,0 +1,22 @@
...
+     * which {@linkplain IResourceActivity resource related activities} should
+     * be queued. The method returns all other activities.
+     * <p>
+     * If flushing of the queue was previously requested, than this method will
+     * additionally return activities from the queue.
+     * 
+     * @param activities
+     * @return activities that are not queued
+     */
+    public List<IActivity> process(final List<IActivity> activities);
```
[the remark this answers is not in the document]
tobous: On the other hand, the addition of another "ActivityQueuer" implementation makes it seem like maybe there should be a general IActivityQueuer implementation.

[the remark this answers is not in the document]
stefaus: Yes, another Impl is the reason for this and the process method is the only common for both.
To avoid the naming confusion I renamed the ActivityQueuer in #155 and In #156 I add an resource based queuing mechanism, which hopefully explains the reason more practical.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/SarosSession.java
stefaus: Yes, Checking on called side, especially public methods is a good idea. How about adding a short assert statement for this?

tobous: I don't know. I generally don't work with assert statements, especially since they are not always enabled. (The tests for the Saros core were always run on our build server without asserts enabled - which lead to nobody noticing the problem described in #98 - so I would assume that normal builds are run with assertions disabled as well.

srossbach: IRC you can have multiple project negotiations for a user concurrently.

See https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ProjectNegotiationObservable.java

stefaus:
> the queuing stuff is project based. even with 1 neg per user this is still needed as you can still receive a negotiation on the client side while a former one is still running

@srossbach I'm not sure what you're trying to tell me, because this PR doesn't remove queuing or changes something on the project based ActivityQueuer.

I hope it becomes clearer what this PR does after reading the [PR message](https://github.com/saros-project/saros/pull/154#issue-203273978) and the context of it via related PRs. 

So this patch creates a Interface for queuing implementations, which is for this patch just the project based ActivityQueuer; a method registering a queue impl for a session and call it from the project negotiation impl. This is preparation for following PRs:
* #155 renames ActivityQueuer to ProjectActivityQueuer to make its behaviour clear
* #156 adds a resource based activity queuer and use it in instant session start
* #157 extracts existing logic to a new method and calls it from the transfer implementation
* #159 remove the session lock for existing session participants at instant session start

So the goal stays that the archive based project negotiaiton uses the project based queue and changes effect only instant session.

If you're interested, my thesis about instant session start is [online](https://www.inf.fu-berlin.de/inst/ag-se/theses/Moll18-saros-session-start.pdf) (please excuse my language and some minor errors in it, it was a hot summer...).

The other PRs are not rebased and need some changes but I will do this the next days.

srossbach:
Add your logic (resource based queuing) to the ActivityQueuer. I will not approve any logic that overtake / delegate resource activity queuing to a random class.

stefaus:
What are your reasons to dislike a separate queuing mechanism only used in instant session start?
I argue that this gives the possibility to leave the archive transfer with project based queuing as a stable fallback.

And what do you mean exactly by random class?

srossbach:
1. you could rename your interface to iactivitymodificator because this is the true nature what this interface offers. and this is not going to happen
2. the old archive transfer can benefit from resource based queuing when partial sharing is used because for now it blocks the whole project.

stefaus:
to 1. could you please be more specific about your concern?
to 2. Yes it would benefit from it, but increases the requirements for a first iteration of a resource based queuer, this way it leaves more options open for optimizations with instant session start and the usage of archive transfer as a stable working fallback. Like you said "95% of the use cases are = full sharing". When the instant session start is proven stable the archive transfer can be deprecated anyway, as it has not advantage in usage.

srossbach:
Just look at the hack inside of the ActivityQueuer and you will get an idea what you can do with a custom ActivityQueuer. You will always find some "smart" people that could abuse this for "fancy" stuff.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractIncomingProjectNegotiation.java
```diff
@@ -213,13 +213,7 @@ protected abstract void transfer(
...
     fileReplacementInProgressObservable.replacementDone();
 
-    /*
-     * TODO Queuing responsibility should be moved to Project
-     * Negotiation, since its the only consumer of queuing
-     * functionality. This will enable a specific Queuing mechanism per
-     * TransferType (see github issue #137).
-     */
-    for (IProject project : projectMapping.values()) session.disableQueuing(project);
+    session.flushQueue();
```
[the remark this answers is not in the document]
stefaus: From my commit message:
> This moves the queuing responsibility into the transfer implementation

this means it moving from the abstract class to the implementation

https://github.com/saros-project/saros/blob/e6403c823eaa34928f2411f084541108665f998d/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/InstantIncomingProjectNegotiation.java#L104-L109

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ArchiveIncomingProjectNegotiation.java
```diff
@@ -36,6 +37,8 @@
 
   private static final Logger LOG = Logger.getLogger(ArchiveIncomingProjectNegotiation.class);
 
+  private static final ActivityQueuer activityQueuer = new ActivityQueuer();
```
srossbach: Please add a comment that this is not the best place to instantiate this object here.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ArchiveIncomingProjectNegotiation.java
```diff
@@ -62,6 +65,8 @@ public ArchiveIncomingProjectNegotiation(
         connectionService,
         transmitter,
         receiver);
+
+    session.registerActivityQueueHandler(activityQueuer);
```
srossbach: Can you add a comment that this is basically a NOP after the first call ?

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ISarosSession.java
```diff
@@ -387,26 +387,17 @@ public void addSharedResources(
   public void changeColor(int colorID);
 
   /**
-   * FOR INTERNAL USE ONLY !
-   *
-   * <p>Starts queuing of incoming {@linkplain IResourceActivity project-related activities}, since
-   * they cannot be applied before their corresponding project is received and extracted.
+   * Register a queue handler for an incoming project negotiation. Should be called only once or
```
srossbach: Great now you glue the Negotiation stuff even harder into the session management. While we tried to make the old methods look like some general generic API methods this new description also tells you the only use case.

Also note there is no way you can "safely" change the handler as you have to do two calls. It does not matter on how often you call flushQueue as activities can arrive at any time and so it is possible that you replace a queuer that still has pending activities.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ISarosSession.java
```diff
@@ -387,26 +387,17 @@ public void addSharedResources(
...
+   * always with the same <code>activityQueue</code>, otherwise the caller is responsible for making
+   * sure that the handler can be safely changed (empty old queue).
    *
-   * <p>That queuing relies on an existing project-to-projectID mapping (see {@link
-   * #addProjectMapping(String, IProject)}), otherwise incoming activities cannot be queued and will
-   * be lost.
-   *
-   * @param project the project for which project-related activities should be queued
-   * @see #disableQueuing
+   * @client
```
srossbach: This is not a valid JavaDoc annotation and we do not have any JavaDoc preprocessor.

If you find "pearls" like "@host, @client, @non-swt ... etc" just deleted them and add an accurate description to the existing JavaDoc.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ISarosSession.java
```diff
@@ -387,26 +387,17 @@ public void addSharedResources(
...
-  public void enableQueuing(IProject project);
+  public void registerActivityQueueHandler(IActivityQueuer activityQueuer);
 
-  /**
-   * FOR INTERNAL USE ONLY !
-   *
-   * <p>Disables queuing for the given project and flushes all queued activities.
-   */
-  public void disableQueuing(IProject project);
+  /** Ensure queued activities get processed. */
```
srossbach: It does not. It only ensure ensures that the process method of the currently used ActivityQueuer is called somewhere in the future.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/IActivityQueuer.java
```diff
@@ -0,0 +1,21 @@
...
+import de.fu_berlin.inf.dpp.activities.IActivity;
+import de.fu_berlin.inf.dpp.activities.IResourceActivity;
+import java.util.List;
+
+public interface IActivityQueuer {
+
+  /**
+   * Processes the incoming {@linkplain IActivity activities} and decides which {@linkplain
+   * IResourceActivity resource related activities} should be queued. The method returns all other
+   * activities.
```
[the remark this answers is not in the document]
stefaus: copy and paste from the ActivityQueuer, i change it.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/IActivityQueuer.java
```diff
@@ -0,0 +1,21 @@
...
+   *
+   * <p>If flushing of the queue was previously requested, than this method will additionally return
```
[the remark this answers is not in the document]
stefaus: copy and paste from the ActivityQueuer, i change it.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ISarosSession.java
stefaus: Ok, I remove the remark to incoming project negotiation.

Regarding the replacement: you would have to ensure that the Queuer you want to replace has a empty queue, how you do it is none of my business (tell the Queuer to dequeue all on next process call and don't queue new ones; flush; check if empty).

srossbach: Have you read my comment ? It is NOT possible without hacking further into the session stuff. So either you add logic to make this possible or restrict the setQueuer call to only be allowed once during a session so you cannot replace it anymore.

The process method is called by a network thread, the offered session methods are called by some other threads.

Didn't you already noticed we added the queuing in such a way that you normally cannot control it ? The SarosSession is no playground for new developers as this is plagued by multiple thread access.

stefaus: I added a check to only allow one queue for a session

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ISarosSession.java
```diff
@@ -387,26 +387,14 @@ public void addSharedResources(
   public void changeColor(int colorID);
 
   /**
-   * FOR INTERNAL USE ONLY !
-   *
-   * <p>Starts queuing of incoming {@linkplain IResourceActivity project-related activities}, since
-   * they cannot be applied before their corresponding project is received and extracted.
+   * Register a activity queuer. Effective only at first call as all following calls are ignored.
    *
-   * <p>That queuing relies on an existing project-to-projectID mapping (see {@link
```
tobous: Is this comment no longer applicable? It is not mentioned in the interface documentation. If it still contains valuable information, you could use `{@inheritDoc}` to extend the interface javadoc for this method.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/SarosSession.java
```diff
@@ -1037,13 +1040,22 @@ public void changeColor(int colorID) {
...
-    activityQueuer.enableQueuing(project);
+  public void registerActivityQueuer(IActivityQueuer activityQueuer) {
+    if (activityQueuer == null || !activityQueuerChangeAllowed) {
+      return;
+    }
+
+    synchronized (this.activityQueuer) {
+      if (activityQueuerChangeAllowed) {
+        log.info("register a new Activity Queue Handler: " + activityQueuer);
+        this.activityQueuer = activityQueuer;
```
tobous: You replace the `activityQueuer` object that you are using as a lock. The way I would understand locking, this would allow another thread to enter the synchronized block before the current thread has left it as you have replaced the object used for locking. Or am I wrong on this?

If this is the case, the current logic could lead to concurrency issues as you replace the lock-object before setting the if condition. This would allow another thread to enter the synchronized block after the `activityQueuer` was replaced but before `activityQueuerChangeAllowed` is set to false by the current thread, meaning the second thread could proceed into the if block and also replace the `activityQueuer` object.

To avoid this problem (if it is a problem), you have to set `activityQueuerChangeAllowed` before replacing the object or use an unrelated object as the lock.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/InstantIncomingProjectNegotiation.java
```diff
@@ -31,6 +32,13 @@
 
   private static final Logger log = Logger.getLogger(InstantIncomingProjectNegotiation.class);
 
+  /**
+   * the one queuer used for all incoming project negotiations
+   *
+   * <p>TODO change to resource based queuer
```
tobous: I am a bit confused by this comment. Could you explain what it means (in this context)?

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/SarosSession.java
srossbach: I do not think it has to be thread safe, as it will be not called concurrently anyway

What you can indeed do is use the default Queuer (I think you had one in the former patch set) and then do

registerActivityQeuer(IActivityQueuer queuer) if (queuer == null) throw ... if (currentQueuer != DEFAULT_QUEUER) throw ... currentQueuer = queuer

However as we already know it is set over and over again and the code now looks like this. And this is bad API design. 

You can use a static block inside the IPNs or add a custom component that registers the required upon session startup.

If you do not want to throw an exception return at least true or false so the caller could get a feedback.

And as Tobias already mentioned synchronizing on the passed in activityQueuer object does not make any sense.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/InstantIncomingProjectNegotiation.java
stefaus: only one queuer exists during the whole session for all incoming negotiations

the TODO is a remark to change this queue, part of #156

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ISarosSession.java
stefaus: Calling `addProjectMapping` is relevant for any resource-activity processing, but I'm not sure if this must mentioned explicitly.
I moved it to `ActivityQueuer`, because other impls could use a different approach (like just queue everything).

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/SarosSession.java
stefaus: Yes sry this sync block was dump, I changed it to an AtomicBoolean and added a return if the call was successful.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ISarosSession.java
srossbach: The method states it is for internal use because you should now the internals when calling this. Although I do not know why we added this comment instead of throwing an exception. The mapping is necessary otherwise the marshalled SPath cannot be unmarshalled. The marshalled SPath looks like  i="3234823" p="some/relativePath/foo.txt"

See https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/misc/xstream/SPathConverter.java

Moving the comment to the ActivityQueuer is not helpful, as the activities are not lost there but here

https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/SarosSession.java#L709

See also https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/activities/IActivity.java#L84

stefaus: So are you ok with just deleting it?

stefaus:
Since this approach conflicts with #355, it is closed.
The new goal will be to completely move the queue mechanism to resource-based.

PR #155, opened 2018-07-23
[FEATURE][CORE] Rename ActivityQueuer

stefaus:
This patch renames ActivityQueuer to ProjectActivityQueuer, to emphasize
its purpose.

extends #154


srossbach:
BTW a refactor is a not a "FEATURE"

stefaus:
Since this approach conflicts with #355, it is closed.
The new goal will be to completely move the queue mechanism to resource-based.

PR #156, opened 2018-07-23
[FEATURE][CORE] Add ResourceActivityQueuer

stefaus:
This patch adds a resource based activity queuer and uses it in Instant
Session start.

extends #155

### DIFF
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/InstantIncomingProjectNegotiation.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/InstantIncomingProjectNegotiation.java
@@ -1,5 +1,6 @@
 package de.fu_berlin.inf.dpp.negotiation;
 
+import de.fu_berlin.inf.dpp.activities.SPath;
 import de.fu_berlin.inf.dpp.communication.extensions.StartActivityQueuingResponse;
 import de.fu_berlin.inf.dpp.exceptions.LocalCancellationException;
 import de.fu_berlin.inf.dpp.exceptions.SarosCancellationException;
@@ -16,12 +17,14 @@ import de.fu_berlin.inf.dpp.net.xmpp.XMPPConnectionService;
 import de.fu_berlin.inf.dpp.observables.FileReplacementInProgressObservable;
 import de.fu_berlin.inf.dpp.session.ISarosSession;
 import de.fu_berlin.inf.dpp.session.ISarosSessionManager;
-import de.fu_berlin.inf.dpp.session.internal.ProjectActivityQueuer;
+import de.fu_berlin.inf.dpp.session.internal.ResourceActivityQueuer;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import org.apache.commons.io.IOUtils;
 import org.apache.log4j.Logger;
 import org.jivesoftware.smack.XMPPException;
@@ -32,12 +35,8 @@ public class InstantIncomingProjectNegotiation extends AbstractIncomingProjectNe
 
   private static final Logger log = Logger.getLogger(InstantIncomingProjectNegotiation.class);
 
-  /**
-   * the one queuer used for all incoming project negotiations
-   *
-   * <p>TODO change to resource based queuer
-   */
-  private static final ProjectActivityQueuer activityQueuer = new ProjectActivityQueuer();
+  /** the one queuer used for all incoming project negotiations */
+  private static final ResourceActivityQueuer activityQueuer = new ResourceActivityQueuer();
 
   public InstantIncomingProjectNegotiation(
       final JID peer, //
@@ -74,9 +73,6 @@ public class InstantIncomingProjectNegotiation extends AbstractIncomingProjectNe
   protected void transfer(
       IProgressMonitor monitor, Map<String, IProject> projectMapping, List<FileList> missingFiles)
       throws IOException, SarosCancellationException {
-
-    awaitActivityQueueingActivation(monitor);
-
     /*
      * the user who sends this ProjectNegotiation is now responsible for the
      * resources of the contained projects
@@ -86,39 +82,40 @@ public class InstantIncomingProjectNegotiation extends AbstractIncomingProjectNe
       final IProject project = entry.getValue();
 
       session.addProjectMapping(projectID, project);
-      /* TODO change queuing to resource based queuing */
-      activityQueuer.enableQueuing(project);
     }
 
+    int missingFilesCount = missingFiles.stream().mapToInt(f -> f.getPaths().size()).sum();
+
+    /* generate file lists */
+    Set<SPath> files = new HashSet<>(missingFilesCount * 2);
+    for (FileList list : missingFiles) {
+      IProject project = session.getProject(list.getProjectID());
+      for (String file : list.getPaths()) {
+        files.add(new SPath(project.getFile(file)));
+      }
+    }
+
+    /* enable queuing for missing files */
+    awaitActivityQueueingActivation(monitor);
+    activityQueuer.enableQueuing(files);
+
+    /* notify host about queuing */
     transmitter.send(
         ISarosSession.SESSION_CONNECTION_ID,
-        getPeer(), //
-        StartActivityQueuingResponse.PROVIDER //
-            .create( //
+        getPeer(),
+        StartActivityQueuingResponse.PROVIDER.create(
             new StartActivityQueuingResponse(getSessionID(), getID())));
 
     checkCancellation(CancelOption.NOTIFY_PEER);
 
-    /* only get files, if something is missing */
-    int filesMissing = 0;
-    for (FileList list : missingFiles) filesMissing += list.getPaths().size();
-
-    if (filesMissing > 0) receiveStream(monitor, filesMissing);
-  }
-
-  @Override
-  protected void cleanup(IProgressMonitor monitor, Map<String, IProject> projectMapping) {
-
-    for (IProject project : projectMapping.values()) activityQueuer.disableQueuing(project);
-
-    super.cleanup(monitor, projectMapping);
+    if (missingFilesCount > 0) {
+      receiveStream(monitor, missingFilesCount);
+    }
   }
 
   private void receiveStream(IProgressMonitor monitor, int fileCount)
       throws SarosCancellationException, IOException {
-
-    String message = "Receiving files from " + getPeer().getName() + "...";
-    monitor.beginTask(message, fileCount);
+    monitor.beginTask("Receiving files from " + getPeer().getName() + "...", fileCount);
     monitor.subTask("Waiting for Host to start...");
 
     awaitTransferRequest();
@@ -133,7 +130,7 @@ public class InstantIncomingProjectNegotiation extends AbstractIncomingProjectNe
 
       IncomingStreamProtocol isp;
       isp = new IncomingStreamProtocol(in, session, monitor);
-      isp.receiveStream();
+      isp.receiveStream(activityQueuer);
     } catch (XMPPException e) {
       throw new LocalCancellationException(e.getMessage(), CancelOption.NOTIFY_PEER);
     } finally {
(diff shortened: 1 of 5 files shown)


de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/ResourceActivityQueuer.java
```diff
@@ -0,0 +1,85 @@
...
+import de.fu_berlin.inf.dpp.session.IActivityQueuer;
+
+/**
+ * This Resource based Queue gets a {@code Set} of {@link SPath}s, which should
+ * be queued and flushes them when the SPath entry was removed by
+ * {@link #disableQueuing(SPath)}.
+ * <p>
+ * To keep it simple, the execution order of queued activities stays unchanged.
+ * </p>
+ * TODO Discard obsolete Editor Events
```
saros-infrastructure: Code Smell Code Smell: Complete the task associated to this TODO comment. ([squid:S1135](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3AS1135))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=156&issues=AWTIBaMkxXt_iLiSqxxk&open=AWTIBaMkxXt_iLiSqxxk)

[ISSUE_KEY]: AWTIBaMkxXt_iLiSqxxk

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/InstantIncomingProjectNegotiation.java
```diff
@@ -58,16 +63,19 @@ public InstantIncomingProjectNegotiation(
...
     @Override
     protected void transfer(IProgressMonitor monitor,
         Map<String, IProject> projectMapping, List<FileList> missingFiles)
         throws IOException, SarosCancellationException {
 
-        awaitActivityQueueingActivation(monitor);
+        try {
+            stoppedLocalEditor = session.getStopManager().stop(
+                session.getLocalUser(), "Read-only while negotiation!");
+        } catch (InterruptedException e) {
```
saros-infrastructure: Bug Bug: Either re-interrupt this method or rethrow the "InterruptedException". ([squid:S2142](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3AS2142))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=156&issues=AWTIBaM2xXt_iLiSqxxl&open=AWTIBaM2xXt_iLiSqxxl)

[ISSUE_KEY]: AWTIBaM2xXt_iLiSqxxl

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/ResourceActivityQueuer.java
saros-infrastructure: Code Smell Code Smell: Complete the task associated to this TODO comment. ([squid:S1135](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3AS1135))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=156&issues=AWTIBaMkxXt_iLiSqxxk&open=AWTIBaMkxXt_iLiSqxxk)

[ISSUE_KEY]: AWTIBaMkxXt_iLiSqxxk

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/ResourceActivityQueuer.java
```diff
@@ -0,0 +1,85 @@
...
+            return activities;
+
+        final List<IActivity> activitiesToExecute = new LinkedList<IActivity>();
+
+        /* check if queue activities are flushable */
+        for (IActivity activity : queueActivities) {
+            if (isAvailable(activity))
+                activitiesToExecute.add(activity);
+            else
+                break;
```
tobous: Why do you use a `break` here? Isn't it possible that the resources for the first activities in the queue are not available, but the resources for activities later in the queue are?
You mention in the javadoc that activities are executed in the order they arrive in to keep it "simple". What purpose does it actually server to keep activities for files queued only because activities for another file could not be executed yet?

They way I understand it, activities for different files should not be able to influence each other, meaning it is only important that the execution order for each resource stays intact.
They could only influence whether the current state between the files is synchronized (when a method is moved from class A to class B while B is not loaded, the method is removed from A but not yet added to B), which should not matter since the "desynchronized" files are not present locally and will be synchronized right after they are loaded locally.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/ResourceActivityQueuer.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import de.fu_berlin.inf.dpp.activities.IActivity;
+import de.fu_berlin.inf.dpp.activities.IResourceActivity;
+import de.fu_berlin.inf.dpp.activities.SPath;
+import de.fu_berlin.inf.dpp.session.IActivityQueuer;
+
+/**
+ * This Resource based Queue gets a {@code Set} of {@link SPath}s, which should
```
tobous: Please mention that activities for the set of SPaths are queued. (Activities are never mentioned in this part of the javadoc.)

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/ResourceActivityQueuer.java
```diff
@@ -0,0 +1,85 @@
...
+    public List<IActivity> process(List<IActivity> activities) {
+        if (queueRessources.isEmpty())
+            return activities;
+
+        final List<IActivity> activitiesToExecute = new LinkedList<IActivity>();
+
+        /* check if queue activities are flushable */
+        for (IActivity activity : queueActivities) {
+            if (isAvailable(activity))
+                activitiesToExecute.add(activity);
```
tobous: Subsequently remove the activity from `queueActivities`?

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/ResourceActivityQueuer.java
```diff
@@ -0,0 +1,85 @@
...
+ * {@link #disableQueuing(SPath)}.
+ * <p>
+ * To keep it simple, the execution order of queued activities stays unchanged.
+ * </p>
+ * TODO Discard obsolete Editor Events
+ */
+public class ResourceActivityQueuer implements IActivityQueuer {
+
+    private Set<SPath> queueRessources;
+    private List<IActivity> queueActivities = new LinkedList<IActivity>();
```
tobous: Rename to `queuedActivities`?

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/ResourceActivityQueuer.java
```diff
@@ -0,0 +1,85 @@
...
+ * be queued and flushes them when the SPath entry was removed by
+ * {@link #disableQueuing(SPath)}.
+ * <p>
+ * To keep it simple, the execution order of queued activities stays unchanged.
+ * </p>
+ * TODO Discard obsolete Editor Events
+ */
+public class ResourceActivityQueuer implements IActivityQueuer {
+
+    private Set<SPath> queueRessources;
```
tobous: Rename to `missingResources` or `unavailableResources`? `queueResources` seems weird. These are the resources activities are queued for, but `resourcesToQueueActivitiesFor` would be way to long.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/InstantIncomingProjectNegotiation.java
```diff
@@ -39,7 +43,8 @@
     private static final Logger log = Logger
         .getLogger(InstantIncomingProjectNegotiation.class);
 
-    private final ProjectActivityQueuer activityQueuer = new ProjectActivityQueuer();
+    private ResourceActivityQueuer activityQueuer;
+    private StartHandle stoppedLocalEditor;
```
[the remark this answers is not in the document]
tobous: Ok, then the confusion stems from me working with IntelliJ. There, an "editor" is a window/tab in the main project view that can be used to modify a specific file. Each editor belongs to exactly one file. There can be any number of editors open at once.

Reading it from this perspective is a bit confusing as you are not blocking a specific editor but the entire IDE/any edits. This is why I would prefer to avoid the term "editor".

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/ResourceActivityQueuer.java
tobous: Ok, I understand. Should be fine for now, but I would still prefer a better solution in the future. Could you add an issue for this?

srossbach: So basically it is not that instant. imagine a make a modification to a file that is written as last file on the receiving side.

as a side note:

file a, b, c

when i cycle through the editors in order c, a, b the result after the negotiation should be: i see file b open in the session overview, neither a or c. i guess that is not trivial if you want to release all activities at once for a resource that got available.

stefaus: Please think this in the context of instant session start in which this scenario is not going to happen / has no impact.
ISS doesn't use a fixed file order. Every opened and especially files opened during the negotiation are transmitted first. Even when you switch between opened files, the active file is transmitted next or probably is already transmitted as this is happening fast.

As a result (in normal / most cases) there is nearly no latency in applying of activities, because the transmission of even a big file like `SarosSession.java` (around 1100 LOC, compressed ~ 9 kB) is fast enough transmitted.

If you like to see a basic example of ISS in action [your'e welcome](https://stefaus.de/saros/instant.mkv)

---

My idea for the next iterations of this queuer is to queue things like open files, selected lines, ... different. Like if you already have a `open file activity` in the queue and get a new one discard the old (easier and probably sufficient, but possible side effects?! - there is some hack in the project based queuer...) or queue it for opening it in background.
But the impact of these cases is low, no data corruption and the next user input would already fix the wrong viewport...

So this queuer is only the first iteration and will get some improvements. 
After all ISS is for now an optional Transfer Type marked as experimental.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/ResourceActivityQueuer.java
```diff
@@ -0,0 +1,85 @@
...
+     * Should be called after Resource is completely written to disk.
+     *
+     * @param resource
+     */
+    public void disableQueuing(SPath resource) {
+        queueRessources.remove(resource);
+    }
+
+    @Override
+    public List<IActivity> process(List<IActivity> activities) {
```
stefaus: missing `synchronized`

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/internal/ResourceActivityQueuer.java
srossbach: This was a theoretical concern what could happen. If I would look at a more realistic scenario I would came to the conclusion that the major time of the negotiation is spent in calculating checksums and so in reality the difference between ISS and the old way is almost zero if you always reuse your projects.

stefaus:
Since the base approach conflicts with #355, this is suspended till further refactorings are done.

PR #157, opened 2018-07-23
[FEATURE][CORE] Move Projects Addition

stefaus:
This patch extracts existing logic to addProjectsToSession() and moves
it call, into the transfer implementation.

extends #156

### DIFF
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractIncomingProjectNegotiation.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractIncomingProjectNegotiation.java
@@ -138,33 +138,6 @@ public abstract class AbstractIncomingProjectNegotiation extends ProjectNegotiat
       transfer(monitor, projectMapping, missingFiles);
 
       checkCancellation(CancelOption.NOTIFY_PEER);
-
-      /*
-       * We are finished with the negotiation. Add all projects resources
-       * to the session.
-       */
-      for (Entry<String, IProject> entry : projectMapping.entrySet()) {
-
-        final String projectID = entry.getKey();
-        final IProject project = entry.getValue();
-
-        final boolean isPartialRemoteProject = getProjectNegotiationData(projectID).isPartial();
-
-        final FileList remoteFileList = getProjectNegotiationData(projectID).getFileList();
-
-        List<IResource> resources = null;
-
-        if (isPartialRemoteProject) {
-
-          final List<String> paths = remoteFileList.getPaths();
-
-          resources = new ArrayList<IResource>(paths.size());
-
-          for (final String path : paths) resources.add(getResource(project, path));
-        }
-
-        session.addSharedResources(project, projectID, resources);
-      }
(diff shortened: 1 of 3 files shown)


de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ArchiveIncomingProjectNegotiation.java
```diff
@@ -108,6 +108,7 @@ protected void transfer(IProgressMonitor monitor,
     @Override
     protected void cleanup(IProgressMonitor monitor,
         Map<String, IProject> projectMapping) {
+        addProjectsToSession(projectMapping);
```
tobous: Why is this called in the cleanup and not at the end of the `transfer(...)` method?

It should basically be the same order (only `checkCancellation(...)` would be called afterwards instead of beforehand). The only big difference would be that with the current implementation, the `addprojectToSession(...)` call is located in the `finally` block, meaning it also gets executed when the incoming project negotiation fails. This is probably the wrong behavior as it does not make much sense to add the "shared" resources to the session when the negotiation/resource sharing failed.

stefaus: You are right, it should move into transfer. Somehow I thought cleanup is called right after transfer, but even then it should move there.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractIncomingProjectNegotiation.java
```diff
@@ -174,6 +147,31 @@ public Status run(Map<String, IProject> projectMapping, final IProgressMonitor m
     return terminate(exception);
   }
 
+  /**
+   * Adds all projects and resources to the session. This notifies all listener about the completed
```
srossbach: I cannot spot any listener call here.

stefaus: listeners of `resourcesAdded` in `session.addSharedResources(...)`

srossbach: I am aware of the broadcasting. However you have no control of what is broadcasted in this method call. In this case you can argue that the broadcast is a result of a finished negotiation. I still find the comment confusing as there is no listener invoked which actually announce a completed negotiation. This is in fact done when you call the terminate method.

And btw, addProjectsToSession is also not correct. It adds resources which may be projects but in case of partial sharing it can really only be resources that are added to an already shared project.

stefaus: removed the listener comment and renamed it

stefaus:
Since the base approach conflicts with #355, this is suspended till further refactorings are done.

PR #165, opened 2018-07-25
[INTERNAL][WB] Added IWhiteboardBrowser and its Eclipse implementation

AbdBarho:
this implementation will be later used in the bridge between the sxe controller of the whiteboard and the browser.
this interface should allow for an eclipse independent implementation of the whiteboard, if for example the whiteboard would be ported to intellij, only the view and an instance of the IWhiteboardbrowser would be needed.

original Gerrit commit: 
[https://saros-build.imp.fu-berlin.de/gerrit/#/c/3550/](https://saros-build.imp.fu-berlin.de/gerrit/#/c/3550/)

### DIFF
--- a/de.fu_berlin.inf.dpp.whiteboard/src/de/fu_berlin/inf/dpp/whiteboard/ui/browser/BrowserRunnable.java
+++ b/de.fu_berlin.inf.dpp.whiteboard/src/de/fu_berlin/inf/dpp/whiteboard/ui/browser/BrowserRunnable.java
@@ -1,0 +1,19 @@
+package de.fu_berlin.inf.dpp.whiteboard.ui.browser;
+
+/**
+ * Contains the code that will be run when a browser function is called.
+ * <p>
+ * is used to overcome the problem that the {@link Runnable} interface doesn't
+ * accept parameters in its {@link Runnable#run()} method.
+ */
+public interface BrowserRunnable {
+    /**
+     * runs the java code when the JavaScript function is called.
+     * 
+     * @param arguments
+     *            the parameters which have been given in browser while calling
+     *            the function
+     * @return result of the executed code
+     */
+    Object run(Object[] arguments);
+}
--- a/de.fu_berlin.inf.dpp.whiteboard/src/de/fu_berlin/inf/dpp/whiteboard/ui/browser/EclipseWhiteboardBrowser.java
+++ b/de.fu_berlin.inf.dpp.whiteboard/src/de/fu_berlin/inf/dpp/whiteboard/ui/browser/EclipseWhiteboardBrowser.java
@@ -1,0 +1,52 @@
+package de.fu_berlin.inf.dpp.whiteboard.ui.browser;
+
+import org.apache.log4j.Logger;
+import org.eclipse.swt.browser.Browser;
+import org.eclipse.swt.browser.BrowserFunction;
+
+import de.fu_berlin.inf.dpp.ui.util.SWTUtils;
+import de.fu_berlin.inf.dpp.whiteboard.ui.HTMLWhiteboardView;
+
+/**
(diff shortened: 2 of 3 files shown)


de.fu_berlin.inf.dpp.whiteboard/src/de/fu_berlin/inf/dpp/whiteboard/ui/browser/BrowserRunnable.java
```diff
@@ -0,0 +1,19 @@
+package de.fu_berlin.inf.dpp.whiteboard.ui.browser;
+
+/**
+ * Contains the code that will be run when a browser function is called.
+ * <p>
+ * is used to overcome the problem that the {@link Runnable} interface doesn't
+ * accept parameters in its {@link Runnable#run()} method.
+ */
+public interface BrowserRunnable {
```
fzieris: Is there a reason to not use the existing browser function infrastructure we already have for the HTML UI? See e.g. 
[TypedJavascriptFunction](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.ui/src/de/fu_berlin/inf/dpp/ui/browser_functions/TypedJavascriptFunction.java) and [TypeRefinery](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.ui/src/de/fu_berlin/inf/dpp/ui/browser_functions/TypeRefinery.java). Seems like reinventing the wheel.

AbdBarho: I went through `TypedJavascriptFunction` and it has exactly what I needed, with it I could definitely avoid reinventing the wheel.

I found out that this function extends [SelfRegisteringJavascriptFunction](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.ui/src/de/fu_berlin/inf/dpp/ui/browser_functions/SelfRegisteringJavascriptFunction.java) which internally register the function in a list called `browserFunctions` in the [BrowserCreator ](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.ui/src/de/fu_berlin/inf/dpp/ui/ide_embedding/BrowserCreator.java) class from the ui package. 

Now if I wanted to use the `BrowserCreator` class to create browser instances with automatically registered functions, I would have to provide an `IPage` which references the whiteboard-html file, and the [IUIResourceLocator](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.ui/src/de/fu_berlin/inf/dpp/ui/ide_embedding/IUIResourceLocator.java) will resolve the relative path to the document and open it in the newly-created browser instance, so far so good. 

Now in this case, the implementation of  `IUIResourceLocator` is [EclipseResourceLocator](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/browser/EclipseResourceLocator.java) which has the BundleID hard coded in it  (which points to the ui package/bundle).

The only way `EclipseResourceLocator` can find the whiteboard is if I change the BundleID, which seems like a very bad idea to me, it would most likely break the ui. 

I might extend `IUIResourceLocator` to resolve resources for the browser, but with this solution I`m solving a problem by doing the same mistake again.

Another possible solution is to move the whiteboard to the ui package, which seems like an overkill.

It should not be ignored that if I use `TypedJavascriptFunction`, the created function will also be registered in all browser instances created for the ui itself, but the ui does not need my whiteboard function(s).

I totally agree that code duplication is bad and should not be allowed, but this commit has everything that I was not able to salvage from the ui package and had to re-create. I think in this case it is not a problem.

Anyway I am open for any suggestions.

PR #169, opened 2018-07-26
[INTERNAL][I] Move content root check to separate method

tobous:
Moves the check whether the given module has exactly one content root in
the CTOR of IntelliJProjectImplV2 to a separate method.

Also improves the message of the thrown exception if there is not
exactly one content root and adds @throws tags for the exception to the
CTOR javadoc.

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/filesystem/IntelliJProjectImplV2.java
+++ b/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/filesystem/IntelliJProjectImplV2.java
@@ -4,6 +4,7 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import com.intellij.openapi.module.ModuleManager;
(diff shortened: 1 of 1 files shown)


tobous:
Clean up for following changes.
Build through Gerrit: https://saros-build.imp.fu-berlin.de/gerrit/#/c/3600/

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/filesystem/IntelliJProjectImplV2.java
```diff
@@ -103,6 +93,41 @@ public IntelliJProjectImplV2(@NotNull final Module module) {
...
+        ModuleRootManager moduleRootManager = ModuleRootManager
+            .getInstance(module);
+
+        VirtualFile[] contentRoots = moduleRootManager.getContentRoots();
+
+        int numberOfContentRoots = contentRoots.length;
+
+        if (numberOfContentRoots != 1) {
+            throw new IllegalArgumentException(
+                "Modules shared with Saros currently must contain exactly one "
```
srossbach: If you try to build a user friendly error message you should not rely on the toString() method of the objects you are using to build the string. In the worst case you get just pointer references which are not helpful for a user.

tobous: I generally try to test whether an object has a useful `toString()` implementation before including it in a log message or exception. (`Module.toString()` just return the module name, `VirtualFile.toString()` returns the paths.)

But you are right. At some point I will forget to check if there is a `toString()` implementation (or worse: an implementation gets removed afterwards). I will try to avoid it in the future.

PR #170, opened 2018-07-26
[INTERNAL][I] Adjust module structure check

tobous:
Adjusts the module structure check in IntelliJProjectImplV2 to allow
more cases that actually work with the current implementation.
Subsequently moves the checks to separate methods to clean up the CTOR
and adds @throws tags for the exceptions to the CTOR javadoc.

Previously, the restrictions were that the content root of all shareable
modules needed to be located in the project root directory. There were
no restrictions concerning the location of the module file. However, if
the module file is not located in the base directory of the content
root, the creation of the new module as part of the project negotiation
fails on the client side.

The new restrictions for shareable modules are now that
 - the content root needs to be located below the project root.
 - the module file needs to be located in the base directory of the
   content root.

The first restriction was added to ensure that shared modules only
contain paths relative to the project root. Otherwise the module file
could contain absolute paths, which would be problematic when trying to
load it on the client side.

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/filesystem/IntelliJProjectImplV2.java
+++ b/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/filesystem/IntelliJProjectImplV2.java
@@ -69,7 +69,14 @@ public final class IntelliJProjectImplV2 extends IntelliJResourceImplV2
      * @param module
      *            an IntelliJ <i>module</i>
      * @throws IllegalArgumentException if the given module does not have
-     *                                  exactly one content root
+     *                                  exactly one content root, the content
+     *                                  root is not located under the project
+     *                                  root or the module file is not located
+     *                                  in the base directory of the content
+     *                                  root
+     * @throws IllegalStateException    if the project base dir, the module file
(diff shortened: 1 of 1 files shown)


tobous:
I still don't fully get how to do stacked pull requests correctly. Had to force push after rebasing on the new head of release/intellij-0.1.0 as the commit from the previous PR was still shown as part of this PR.

I did not change the commit of this PR in any way.

tobous:
OK, then I also had the wrong idea on how to handle stacked PRs. On @jafey PRs I suggested to create one commit pull requests to avoid force-pushing. Otherwise we would have to squash PRs with multiple initial commits (if changes were made through new commits), removing separation of concerns.

tobous:
Original question by @m273d15 

> Why is the IllegalStateException thrown and not a custom exception that is more specific? I found the exception ModuleNotFoundException.
>
>Is there a rule for using custom exceptions / java standard exceptions?

I don't know of any rules concerning the usage of custom exceptions.

In general, this `IllegalStateException` could be replaced with something like a `ResourceNotFoundException`. But I don't know if this carries that much more information. 

The `IllegalStateException` gets thrown if the project root directory, the module file, or the directory containing the module file could not be found in the IntelliJ virtual file system. In most cases, we won't even know the reason why they could not be found. We can only generally assume that the module is currently not configured correctly.

PR #171, opened 2018-07-27
[INTERNAL][I] Add catch block for IllegalStateException

tobous:
Adds a catch block for IllegalStateExceptions in the ContactPopMenu.
IllegalStateExceptions can be thrown by the IntelliJProjectImplV2 CTOR
when the module structure checks can't find certain needed files. As
these exceptions point towards an error in the local module
configuration that is independent of Saros, a warning is shown to the
user for every module that failed with an IllegalStateException.

Beforehand, these exceptions were uncaught and stopped the contact pop
menu from opening if a faulty module existed.

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/ui/Messages.java
+++ b/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/ui/Messages.java
@@ -93,6 +93,8 @@ public class Messages {
     public static String ContactPopMenu_module_not_found_message_condition;
     public static String ContactPopMenu_invalid_module_title;
     public static String ContactPopMenu_invalid_module_message_condition;
+    public static String ContactPopMenu_error_creating_module_object_title;
+    public static String ContactPopMenu_error_creating_module_object_message;
 
     private Messages() {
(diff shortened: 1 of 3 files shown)


tobous:
The wording for the message for the user is a bit clumsy, but I could not think of a better one.
Build through Gerrit: https://saros-build.imp.fu-berlin.de/gerrit/#/c/3602/

tobous:
Force push to clean up after merging base PR. Did not change the commit of this PR.

m273d15:
@tobous The PR itself LGTM, but I have a more or less general question. Why is the IllegalStateException thrown and not a custom exception that is more specific? I found the exception ModuleNotFoundException.

Is there a rule for using custom exceptions / java standard exceptions?

tobous:
I think this discussion should be continued on PR #170, where these exceptions are actually thrown. In this PR, I would simply add a multi-catch for the new exception types if I were to add different exceptions. The general user message should still work.
(While we are on the topic of the general user message: I am still not completely happy with the title/message, so I am open to suggestions for improvements.)

tobous:
This change is not not directly dependent on the PR #170, so I am going to rebase it onto release/intellij-0.1.0 and treat it as a separate PR.

Issue #133, opened 2018-07-16
How to handle stacked pull requests?


m273d15:
### Assumption:
We have three feature changes F1,F2,F3 and F2 depends on F1 and F3 depends on F2.
This leads to a stack/chain of changes. The old gerrit system was able to handle such chains.

How do we handle these chains in GitHub?

### My first proposal was to:
* chain the branches corresponding to the dependencies of the changes
  * Create a branch `branch-f1` that bases on `master`
  * Create a branch `branch-f2` that bases on `branch-f1`
  * Create a branch `branch-f3` that bases on `branch-f2`
* Create corresponding pull requests
* Merge the pull requests in the order
  * Merge `branch-f1`
  * Change target of F2 (`branch-f2`) pull request to `master`
  * Rebase `branch-f2`
  * Merge `branch-f2`
  * Change target of F3 (`branch-f3`) pull request to `master`
  * Rebase `branch-f3`
  * Merge `branch-f3`

#### Advantages
* Each pull request contains the changes of the specific feature (dependent commits are not shown)

#### Disadvantages
* The rebase process removes existing approvals

See [here](https://graysonkoonce.com/stacked-pull-requests-keeping-github-diffs-small/) fore another approach that uses another merge order.

Are there other proposals how this process can be simplified?

stefaus:
The other approach, has the disadvantage of waiting till every PR in line is approved (in the review time, new PRs could be created and this could take forever). 

This could reduce some work and maybe keep the reviews untouched, but have to try it out first: [Managing dependent pull requests](https://wchargin.github.io/posts/managing-dependent-pull-requests/)

m273d15:
In order to simplify the reviewing process of stacked PRs it can be useful to use the GitHub filter function with the arguments `is:pr is:open base:master`.

PR #172, opened 2018-07-27
[STF] Add RemoteHTMLWidgets

jafey:
This pull request contains all classes the htmlBot uses for interacting with the HTML-GUI.

Also there are some basic tests which uses the htmlBot (AddContactTest & StartSessionWizardTest)


tobous:
Why are some commits present twice in this PR? For example, "
[STF] Add RemoteHTMLCheckbox and RemoteHTMLRadioGroup" is present with the commit id 8ecd14f93700f805b94234556d4cc709935e52e9 and a2d2e0d2b82528f50f32a939ded7198585c1df13. Please clean up the commit history for this PR.

jafey:
@tobus I followed the git instructions to resolve merge conflicts and after I pushed the merge to this branch git created new commits in the PR automatically.

What I did is:
```
git fetch origin
git checkout -b pr/stf-project origin/pr/stf-project
git merge master
git push
```

tobous:
Commits for the Saros repository should only contain Unix line endings. It looks like all of your changes have Windows line endings.
Please add the correct `core.autocrlf` setting to your git config. You can have a look at the corresponding section on https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration to figure out the needed setting for your operating system. (For Windows, this should be "core.autocrlf=true".)

After adding this setting, you will have to to adjust all of your commits. The same applies to your commits in #173, #174, #175.

Please use this opportunity to
- split #172 (this PR) into multiple stacked PRs. Currently, this is way to much code to review properly.
- split #175 into multiple PRs. See my comment in the PR about separating changes from refactorings. The base PR of this PR-chain/stack should point to the last PR of the stack for #172.
- split #173 into two (stacked) PRs.

A basic explanation on how we currently handle stacked PRs is given in the initial comment of the ticket #133. If you don't agree with this approach or have a better one, please add to the discussion.

tobous:
As a side note when creating new PRs:

If you are of the opinion that multiple commits should be part of the same PR, you will have to keep the list of commits clean until committing. This means you can't add new commits for fixing stuff noted during the review or for merging with the master, etc.

We don't want to squash multiple commits as this would remove the separation of concerns and worsen readability, which were the reasons to create multiple commits in the first place, meaning we would have to rebase the commits during the merging of the PR.

This means all changes during the review process will have to be done by amending an existing commit and then cherry-picking any other commits in the PR (in the right order) that are based on the changed commit. After all changes are done locally, the changes will have to be force-pushed onto the GitHub branch. Force-pushing is risky as it will override the existing commits/history of the PR, meaning mistakes while amending/cherry-picking commits are hard to revert.

This is why, with our current workflow, I would generally recommend to create PRs only containing single commits (initially). This way, any changes can be done with additional commits that are then squashed when merging the PR.
Furthermore, force-pushing messes up the PR overview a bit, making it harder to get an overview what happened during the review.

jafey:
@tobus please look at the new branches and pull request i made. Is it okay now.

tobous:
PR #186 still has issues with line endings (in IRemoteHTMLInputField and RemoteHTMLInputField). There also seem to be problems with line endings in the following changes in the chain (I only looked at a few at random).

Also, as noted in #186, please choose the current master as the base of all of your PR chains as Kelvin recently added a commit that should enforce LF endings for new commits. I don't know what happens when you rebase existing commits, so you might have to adjust these yourself.

PR #175, opened 2018-07-27
[BUILD] Build new stf and stf.test project

jafey:
In this pull request I added a new de.fu_berlin.inf.dpp.stf project which includes all classes for the
Saros-Test-Framework. The source folder is separated in /html, /eclipse and /intellij. Also this project includes the STF-Self-Tests.

In addition I added a new de.fu_berlin.inf.dpp.stf.test project which includes all STF-Tests
The core, ui, eclipse & intellij projects are now independent from the STF.

Commits bevor Jul 18, 2018 are part of another pr (https://github.com/saros-project/saros/pull/172) but this pr has it as base


tobous:
Please choose the branch of your first pull-request as the target for this pull-request. That way, the changes of the other pull-request won't be shown here.

After changing the target, please don't merge this PR until the base PR is merged.
After the base PR is mergerd, you can again change the target of this pull-request to the master. At that point, you will have to rebase this branch/pull-request on the newest state of the master.

tobous:
Does this patch contain any code changes or is it a pure refactoring (+ the setup of the new project)?

jafey:
@tobous For now it is only Refactoring and some changes to build the new projects. But today I found an issue that breaks everything. So actually no review necessary until I fix it.

tobous:
That is fine, but you should be careful what you include in the PR. You should generally not mix refactorings with other changes. Refactorings can change huge chunks of code but usually don't require an in-depth code review as the work is done automatically by the IDE (or some other tool). This makes it very easy for real changes to fall through the gaps.

As is, looking through all of the "code changes" in this change is very impractical (especially since GitHub for whatever reason marks some of the renamed files as completely changed). If possible, I would even prefer if you could split this into two PRs, one for the refactoring (moving the code) and one introducing the project and setting up the build environment.

As a general rule of thumb, I would treat pull requests just like Gerrit changes. Each PR should be self contained and as small as possible while still being sensible.

PR #180, opened 2018-08-02
Enforce LF in all text files

m273d15:
Use .gitattributes in order to enforce LF. Therefore all CRLFs are overwritten with LF during git operations (as merge, commit, add, checkout,..).


.gitattributes
```diff
@@ -0,0 +1,11 @@
+* text eol=lf
+
+*.icns binary
+*.zip binary
+*.wav binary
+*.swp binary
```
fzieris: Do we actually have files of these types in our repo?

m273d15: My first search included some build results, therefore `*.swp` is superfluous, but the remaining types occur (even if only one time) in our repo.
* `./de.fu_berlin.inf.dpp.ui.widgetGallery/icons/alt_launcher.icns`
* `./de.fu_berlin.inf.dpp.ui/libs/swt-browser-improved-src.zip`
* `./de.fu_berlin.inf.dpp/assets/sounds/*.wav`

m273d15:
I removed the type `*.swp` in the new rebased version

PR #181, opened 2018-08-03
Fix context menu user name

tobous:
Fixes the names being displayed incorrectly when sharing a module through the context menu. This is done by numbering the available users and using the number as the shortcut. More information can be found in the commit message of b7ad18b982e53584d76eeb2fa1c5779270df2e8b.

25235b28deb7767023c784e2592c3c8e258f0594 and 3493f1fc3c1484e032f1bd116a2a353461c52905 subsequently improve the message of the context menu entry and add shortcuts and descriptions to Saros menu entries.

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/META-INF/plugin.xml
+++ b/de.fu_berlin.inf.dpp.intellij/META-INF/plugin.xml
@@ -35,16 +35,17 @@
 
     <actions>
         <!-- Add your actions here -->
-        <group id="saros.MainMenu" text="Saros" description="Saros menu">
+        <group id="saros.MainMenu" text="S_aros" description="Saros menu">
             <add-to-group group-id="MainMenu" relative-to-action="HelpMenu" anchor="before"/>
             <action id="saros.about"
                     class="de.fu_berlin.inf.dpp.intellij.ui.menu.AboutSarosHandler"
-                    text="About Saros"/>
+                    text="_About Saros"/>
         </group>
 
         <group keep-content="true" compact="false" popup="true"
                id="saros.ShareWith"
-               icon="/icons/famfamfam/session_tsk.png" text="Share With..."
+               icon="/icons/famfamfam/session_tsk.png" text="_Share With..."
+               description="Share this module using Saros"
                class="de.fu_berlin.inf.dpp.intellij.ui.menu.SarosFileShareGroup">
             <add-to-group group-id="ProjectViewPopupMenu" anchor="last"/>
         </group>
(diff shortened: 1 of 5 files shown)


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/ui/menu/SarosFileShareGroup.java
```diff
@@ -68,12 +68,16 @@ public void actionPerformed(AnActionEvent e) {
             return new AnAction[0];
         }
 
+        int userCount = 0;
```
fzieris: Classic: Start counting at zero 😄

de.fu_berlin.inf.dpp.intellij/META-INF/plugin.xml
```diff
@@ -35,16 +35,17 @@
...
                     class="de.fu_berlin.inf.dpp.intellij.ui.menu.AboutSarosHandler"
-                    text="About Saros"/>
+                    text="_About Saros"/>
         </group>
 
         <group keep-content="true" compact="false" popup="true"
                id="saros.ShareWith"
-               icon="/icons/famfamfam/session_tsk.png" text="Share With..."
+               icon="/icons/famfamfam/session_tsk.png" text="_Share With..."
+               description="Share this modules through Saros"
```
fzieris: Grammar: Either "theses modules" or "this module"
Wording: "with Saros"?

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/ui/menu/SarosFileShareGroup.java
tobous: I didn't think counting from zero is a problem here. But I can change it if you would prefer.

de.fu_berlin.inf.dpp.intellij/META-INF/plugin.xml
tobous: Fixes, changed to "this module".
I don't like the wording "with Saros" as I always here it in the context of "Share the module with user XYZ", which would suggest to me that we are trying to share the module with a user named "Saros".

How about "Share this module using Saros"?

fzieris: If I'm not mistaken, that's the difference between *ablative of manner* and *ablative of accompaniment* (language nerds, [this way][1]).

Your suggestion "using Saros" is fine.

[1]: https://latin.stackexchange.com/questions/608/fighting-with-someone-and-the-different-uses-of-with-in-latin

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/ui/menu/SarosFileShareGroup.java
fzieris: Some users may be confused. Imagine you have one single contact, and this appears as your only entry: "0: bob".

This if fine for now. We can conduct user tests in the future, if we really want to know what works best.

tobous: User count now starts at 1.

PR #185, opened 2018-08-06
[FIX][CORE] Only notify participants if the user left the session

tobous:
When a local session ends, the local Saros instance tries to notify the
other participants of the session. If the local user is the host of the
session, all other participants will be notified. Otherwise, just the
host will be notified.

Saros now only notifies other participants that the local session has
ended if the reason was "Local_USER_LEFT", meaning the local user left
the session manually.

For the other possible reasons for a local session end, this
notification will always fail:

HOST_LEFT: When the host leaves the session, the session ends.
Furthermore, the host closes all its connections when leaving the
session, meaning the other participants would not be able to notify the
host anyway.

CONNECTION_LOST: Without a working connection, the participant won't be
able to notify other participants/the host-

KICKED: The host removed the participant from the session and
subsequently closed the connection to the participant, meaning the
kicked participant can not notify the host.


de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -333,7 +333,7 @@ public void stopSession(SessionEndReason reason) {
             sessionEnding(session);
 
             try {
-                session.stop();
+                session.stop(reason);
```
tobous: Another option would be to do the check which reason caused the session end here and then just pass a boolean `notifyParticipants` to `SarosSession#stop()`.

I'm split on which option would be better.

PR #186, opened 2018-08-07
[STF] Add RemoteHTMLInputField and use it in AddContactTest

jafey:
In addition I added a method to get the contactlist from the HTML-GUI.
To reduce code duplication I build a method to get list items by any
selector. I build it in a new BotUtils class.

Change-Id: Ia9e158a32d772593cbd74eb7a215252d7af2889d


tobous:
This PR (and probably others along the chain) still contains windows line endings.

Also, please use the current master as the base for this PR chain as it now contains the settings to automatically only commit LF endings to avoid this problem in the future.

jafey:
I just saw that this commit is still in the master.

PR #187, opened 2018-08-07
[STF] Open HTML view via SarosApi and InputField improvement

jafey:
In this commit I added a function to open a view directly in the
browser see RemoteHTMLView.open().

Change-Id: I277dec2a7275a82d51966a79610734e20d03c11b


jafey:
Already in master

PR #189, opened 2018-08-07
[STF] Add checkbox, radio group, select, multi select and progressbar

jafey:
All five components are very similar


jafey:
I want to re-open this pr

PR #193, opened 2018-08-07
[UI][FIX] Set size of EclipseBrowserDialog

jafey:
The problem was, that the dialog size was bigger that the window of the
actual eclipse instance. I wrote a method which set the size of the
dialog to 80% of the actual parent size. This quick fix will be enough
for the moment.

Change-Id: Ib20f202d30d2671ee73ccd6b3cf102633280a9b2


de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/browser/EclipseBrowserDialog.java
```diff
@@ -46,4 +47,25 @@ private void centerShellRelativeToParent() {
             + (parentShellBounds.width - shellSize.x) / 2, parentShellBounds.y
             + (parentShellBounds.height - shellSize.y) / 2);
     }
+
+    private void sizeShellNotBiggerThanParent() {
```
m273d15: Why not `sizeShellSmallerThanParent` instead of the negated version?

tobous: The name still seems overly "wordy" to me. My suggestion would be to call it `adjustShellSize` and add a more detailed explanation as javadoc for the method.

de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/browser/EclipseBrowserDialog.java
```diff
@@ -48,7 +48,11 @@ private void centerShellRelativeToParent() {
             + (parentShellBounds.height - shellSize.y) / 2);
     }
 
-    private void sizeShellNotBiggerThanParent() {
+    /**
+     * If the shell is bigger that the parent, the shell size is set to 80% of
```
m273d15: is bigger tha[n] the parent

de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/browser/EclipseBrowserDialog.java
```diff
@@ -46,4 +47,29 @@ private void centerShellRelativeToParent() {
...
+
+        final Rectangle parentShellBounds = parent.getBounds();
+        final Rectangle shellBounds = shell.getBounds();
+
+        if (shellBounds.height > parentShellBounds.height
+            || shellBounds.width > parentShellBounds.width) {
+            int nWidth = (int) (parentShellBounds.width * 0.8);
+            int nHeight = (int) (parentShellBounds.height * 0.8);
+
+            shell.setSize(nWidth, nHeight);
```
tobous: As a general question: Does the aspect ratio of the "child" shell matter?
Or with other words, does it matter that this adjustment does not preserver the original ration between width and height of the "child" shell?

Jannis1337: The shell Size was never set before, so my fix is only to prevent that the dialog is bigger that the eclipse window. I had some issues on my pc where I wasn't able to close or move the dialog because the dialog opens outside the visible area of the screen.

In general the aspect ratio does't matter for the HTML-GUI because it is responsive. Maybe it makes sense to define some ratios/sizes for different dialogs when the HTML-GUI is nearly finished. But for now it doesn't matter.

de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/browser/EclipseBrowserDialog.java
```diff
@@ -16,6 +16,7 @@
 
     public EclipseBrowserDialog(Shell shell) {
```
srossbach: The CTOR is really bad. If you look at the SWT framework and Eclipse stuff you will notice that the shell is always named parent for a good reason.

This class however expects a concrete shell which will be further mutated during the application lifetime which is really bad because the class is not the owner/creator of the shell.

de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/browser/EclipseBrowserDialog.java
srossbach: See my comment above in the CTOR, create the shell by yourself.

m273d15:
@srossbach Do you disagree with merging this PR? The author is not available anymore, the change introduces a usable fix, but does not improve the existing CTOR. Otherwise I will close this PR and open a corresponding issue that references this PR.

srossbach:
Please do not merge this. Change this so it confirms with the guidelines, i.e you pass in the parent shell (normally the Eclipse Workbench Window) and you create your own shell on top of this shell.

m273d15:
Please, refrence #558 if you want to continue with this issue

PR #194, opened 2018-08-07
[STF][REFACTOR] remove duplicate code, move/rename test and packages

jafey:
All HTML widget had duplicate methods and variables which I moved to the
abstract HTMLSTFRemoteObject.
I renamed the remotebot.widget and remotebot.widget.impl package to
htmlbot.widget and htmlbot.widget.impl
Also I renamed the ComponentViewTest to HtmlBasicWidgetTest and move it
to the StfSelfTest package.

Change-Id: Id77e3110172495fec74724a78d4d82902e638738


Jannis1337:
I see but I don't know why. Whatever, the important thing is that the package name is correct like ...htmlbot.widget.impl and not ...remotebot.widget.impl

jafey:
I see but I don't know why. Whatever, the important thing is that the package name is correct like ...htmlbot.widget.impl and not ...remotebot.widget.impl

PR #198, opened 2018-08-07
[UI][I] Add user notification for session status changes

tobous:
Adds a notification that informs the user when a new session is started.
For the host, this contains all users that the session is started with
(usually nobody). For clients, this contains the host of the joined
session.

Adds a notification that informs the user when the current session ends.
This notification also includes the reason the session ended.

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/context/SarosIntellijContextFactory.java
+++ b/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/context/SarosIntellijContextFactory.java
@@ -11,6 +11,7 @@ import de.fu_berlin.inf.dpp.core.awareness.AwarenessInformationCollector;
 import de.fu_berlin.inf.dpp.core.monitoring.remote.IntelliJRemoteProgressIndicatorFactoryImpl;
 import de.fu_berlin.inf.dpp.core.project.internal.SarosIntellijSessionContextFactory;
 import de.fu_berlin.inf.dpp.core.ui.eventhandler.NegotiationHandler;
+import de.fu_berlin.inf.dpp.intellij.ui.eventhandler.SessionStatusChangeHandler;
 import de.fu_berlin.inf.dpp.core.ui.eventhandler.UserStatusChangeHandler;
 import de.fu_berlin.inf.dpp.core.ui.eventhandler.XMPPAuthorizationHandler;
 import de.fu_berlin.inf.dpp.core.util.IntelliJCollaborationUtilsImpl;
@@ -86,6 +87,7 @@ public class SarosIntellijContextFactory extends AbstractContextFactory {
         Component.create(NegotiationHandler.class),
         Component.create(UserStatusChangeHandler.class),
         Component.create(XMPPAuthorizationHandler.class),
+        Component.create(SessionStatusChangeHandler.class),
 
         Component.create(IChecksumCache.class, NullChecksumCache.class),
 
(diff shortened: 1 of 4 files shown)


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorManager.java
```diff
@@ -449,7 +453,36 @@ private void endSession() {
             openEditorPaths.clear();
         }
 
+        private void notifyUser(
```
srossbach: Misplaced ?  The EditorManager is not even part of the ui package. Please find a better place.

tobous: I agree that it seems misplaced here. To be honest, I just put it here because this deals with the main logic ending the session.

I don't know where else to put it. Maybe into the [`UserStatusChangeHandler`](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/core/ui/eventhandler/UserStatusChangeHandler.java), but this does not seem very fitting either.

I could move it to its own class (that just contains a `ISessionLifecycleListener`), but this still leaves the problem where to put that class.
I am not sure whether the `dpp.intellij.ui` package is the right location. The package mainly contains the implementation of different UI elements. This method just uses parts of the UI to notify the user, but is not part of the UI itself. I would associate it more with the general session logic.

How about `dpp.core.ui.eventhandler` (the core package contained in the IntelliJ project, not the core project)? This would be the best place I could find while looking through the project.

(Or `dpp.core.awareness`, but this does not fit that well either.)

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/core/ui/eventhandler/SessionStatusChangeHandler.java
```diff
@@ -0,0 +1,118 @@
...
+
+        String sessionStartMessage;
+
+        if (session.isHost()) {
+            List<User> participants = session.getRemoteUsers();
+
+            if (participants.isEmpty()) {
+                sessionStartMessage = Messages.SessionStatusChangeHandler_session_started_host_empty_message;
+
+            } else {
```
tobous: I am not  sure if the list of users in the created session is really necessary. As far as I know, sessions can currently only be created empty (without a participant).

tobous:
@srossbach I moved the logic to a separate class `SessionStatusChangeHandler` and changed the location to `dpp.core.ui.eventhandler`. I also included a message when starting a session. (See my note about displaying the list of initial participants.)

srossbach:
To my understand all classes that resides in *.core.* where supposed to be moved to the core. However is this still the case ? If not then please move them to the intelliJ package.

tobous:
I have no idea. This is the first I have heard of it. The package info still says that it's supposed to be moved to the core (at some point). I will ask Franz when he is available again (the week after next week).

Do you have a better suggestion where to put this class in the meantime?

srossbach:
Replace core with intellij and you should be fine. At some point all the logic (Eclipse / IntelliJ) should be moved to the ui/html bundle but that should not be your problem.

tobous:
@srossbach Are there any other problems with this PR? Otherwise I would appreciate you approving it.

I am still not sure whether the option to display the initial members of a created session is necessary/useful. I don't know if it is even possible to create such a session or if sessions can only be created empty.

PR #200, opened 2018-08-08
[DO NOT MERGE] Create index.html and use bootstrap without theme

m273d15:
index

* This change is just a proposal and the switch
  to bootstrap only will be moved into a separate PR


m273d15:
Hi, this is a proposal of an initial index page design. I will create a clean PR that contains this changes.

Issue #167, opened 2018-07-26
Create index page


m273d15:
The index page of the saros project should become a marketing page that contains the most important information and links.
e.g.
Links:
- "Install Saros"
- "Getting Started Documentation"
- "Downloads"

Information:
- Description of Saros in one sentence as a big header
- A more detailed description of Saros in e.g. 4 sentences.

PR #204, opened 2018-08-09
[DOC] Add index page

m273d15:
Implementation of #167


docs/_includes/index_navbar.html
```diff
@@ -0,0 +1,12 @@
+<ul class="list-group">
+    <a class="list-group-item list-group-item-action flex-column align-items-start active bg-blue">
```
tobous: I would prefer another color for the "Resources" header box. From the used blue I would infer that the box is interactive/can be clicked or is a currently selected tab.

m273d15: Do you have any idea for alternatives? I create two other versions:

**CURRENT VERSION**
index

**ALTERNATIVE VERSION 1**
* Resources entry has the same color as the navigation bar of the other sites
res1

**ALTERNATIVE VERSION 2**
* Removed the resources entry. Used light blue for list entries and made jumbotron white
* If hover over a list entry, the background of the entry becomes white
res2

tobous: Both look fine to me. I would prefer alternative 2, but that is just a matter of personal preference.

chrismotian: I like the second one more, too.

m273d15: Implemented "Alternative 2"

PR #207, opened 2018-08-13
[FIX #153][E] Set the added account as active

### LINKED ISSUE #153: Saros/E automatically connects to a pre-existing account when adding new accounts
When offline and adding a new XMPP account to Saros while there are already accounts registered, when finishing the wizard, Saros automatically connects to a pre-existing account (probably the default account).

This behavior seems confusing. 

### Steps to reproduce
1. Start Saros
1.1 Add a new account to Saros (optional, if there is already an existing account)
2. Go offline
3. Add new account to Saros through the wizard
4. Finish the wizard

This connects to the current default XMPP account, which is unexpected behavior.

Expected behavior:
* Some form of acknowledgement, that adding the new account was successful
* Optional: Automatically connect to XMPP server using the newly entered account

chrismotian:
The next step is to create "Some form of acknowledgement, that adding
the new account was successful" as proposed by the issue author "tobous"

### DIFF
--- a/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/wizards/AddXMPPAccountWizard.java
+++ b/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/wizards/AddXMPPAccountWizard.java
@@ -59,8 +59,7 @@ public class AddXMPPAccountWizard extends Wizard {
 
         assert accountStore.getActiveAccount() != null;
 
-        if (preferences.isAutoConnecting())
-            autoConnectXMPPAccount();
+        autoConnectXMPPAccount();
 
         return true;
     }
--- a/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/wizards/AddXMPPAccountWizard.java
+++ b/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/wizards/AddXMPPAccountWizard.java
@@ -78,7 +78,7 @@ public class AddXMPPAccountWizard extends Wizard {
 
     /**
      * Adds the {@link EnterXMPPAccountWizardPage}'s account data to the
-     * {@link XMPPAccountStore}.
+     * {@link XMPPAccountStore} and set the added account as active.
      * 
      */
     private void addXMPPAccount() {
@@ -100,8 +100,9 @@ public class AddXMPPAccountWizard extends Wizard {
         boolean useTLS = enterXMPPAccountWizardPage.isUsingTLS();
         boolean useSASL = enterXMPPAccountWizardPage.isUsingSASL();
 
-        accountStore.createAccount(username, password, domain, server, port,
-            useTLS, useSASL);
+        XMPPAccount addedAccount = accountStore.createAccount(username,
+            password, domain, server, port, useTLS, useSASL);
+        accountStore.setAccountActive(addedAccount);
     }
 
     private void autoConnectXMPPAccount() {


tobous:
The bug is that if you have autoconnect enabled and add an account, you will automatically be connected to your current active account after the account adding wizard has finished. This is a very weird behavior as you just added an account and are then automatically connected to a totally unrelated account (without any user interaction).

With the current changes in this PR, the user is now always automatically connected to the added account.

In my opinion, this is not the ideal behavior as well, as the user should be asked whether they want to use the added account as their active account and whether they want to connect to it. Furthermore, I am not sure whether autoconnect should apply here. I assumed it was meant to only apply when starting the plugin.

This seems to be a difference of opinion between me and @fzieris, who amended the ticket and added the "Expected behavior" section. I would like to hear his opinion on the matter. But he won't be available until next week, so we might have to wait on that.

srossbach:
As I mentioned: https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/wizards/ConfigurationWizard.java#L24

The inheritance makes no sense. It is used to avoid code duplication and so this is the result of such an approach. In the configuration wizard there is an option to enable the auto-connect feature and normally the wizard is only shown when the account store is empty.

I still would prefer to fix the issue in the ConfigurationWizard, but as you said we have to wait for Franz opinion.

fzieris:
On what exactly is my opinion needed?

I wrote the "expected behavior" section in ticket #153 based on my experience when I observed the bug: The weirdest part was getting no feedback as to whether adding a new account was successful. Even if there was no auto-connect, this would still be poor usability!
Since the observable behavior was the auto-connect with another account, a user may come to think that the account addition failed.
 (I've now come to realized that the ticket subject is misleading as it points in a different direction. The problem is mostly about user expectations and feedback.)

If Saros would auto-connect with the newly added account (as this pull request supposedly does), this would provide the user with the necessary feedback that the addition was successful -- even though as a side-effect.

I hope this explains the order in which I put the "expected behavior" points in the ticket #153:
First and foremost, provide feedback; if it's a good idea, auto-connect with the new account.

tobous:
@fzieris Ok, it seems I interpreted the section "Expected behavior" to strictly. Since you are familiar with the Saros/E workflow and with the topic of usability, I would still appreciate your opinion on my suggested new behavior after the account creation. (Stated below.)

With this patch, Saros will only connect to the created account if auto-connect is enabled, meaning it is not a consistent way of notifying the user that the account creation succeeded.
This misunderstanding is my fault. In my head, "with auto-connect enabled" was implied in my statement
>With the current changes in this PR, the user is now always automatically connected to the added account.

But reading it now, I notice that this is not clear at all.

I agree with Stefan that the inheritance from [AddXMPPAccounWizard](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/wizards/AddXMPPAccountWizard.java) does not make much sense. The inheritance should be removed and the whole functionality should be contained in [ConfigurationWizard](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/wizards/ConfigurationWizard.java).


**Suggested Behavior:**

As a solution to the missing user feedback, I would suggest creating a notification for the successful account creation.

Regarding automatically connecting to the new account, I would prefer giving the user a choice on what to do. I would suggest adding a dialog to ask the user whether they want to use the new account as there active account and after that whether they want to connect to it.

fzieris:
@tobous This feature does not need to be overly sophisticated as most users probably only ever create one account and are happy with it. Simplest, yet consistent solution I can think of:

* If auto-connect is enabled: Connect with new account (which serves as feedback).
* If auto-connect is not enabled: Do not change connection status, but show notification (the feedback).

srossbach:
So what is the point of adding a form of feedback ? You just add something to the account store. If the account works or if the credentials are wrong, well we cannot tell it the user. And it is impossible to finish the wizard with duplicate credentials or invalid characters.

fzieris:
@srossbach Feedback is one of Donald Norman's seven design ideas (among Discoverability and others). Without feedback, users cannot really understand how the system works through their actions as they have no visible effect.

But I do see your point, though. Maybe the problem is not so much the lack of feedback, but the wrong kind of feedback. The auto-connect appeared as (confusing) feedback, but was really just a side-effect.

So I see two possible ways now:
* Only connect after a new account was added if auto-connect is enabled. Save the account silently otherwise. OR
* Always connect after a new account was added. This necessarily provides the feedback you said we could not give to the user (does the account "work").

I prefer the second option. Auto-connect would then mean "auto-connect on startup" -- not more, not less. (As @tobous understood it in the first place.)

chrismotian:
I prefer the second option too. (Always connect after a new account was added. This necessarily provides the feedback you said we could not give to the user (does the account "work").)

m273d15:
@chrismotian It seems as if nobody continues with this PR. Can I close this PR?

chrismotian:
> 
> 
> @chrismotian It seems as if nobody continues with this PR. Can I close this PR?

It seems like the issue would be solved well enough with the changes provided by this PR. I will request review from all participants.

srossbach:
My opinion has not changed.

1. Auto-Connect applies to the active account.
2. Adding an account to a non-empty store = do not make that account active, at least ask the user
3. If the user confirms to make that account active / the default one, ask him again if he would like to connect with that account now regardless of the auto-connect settings and tell the user that this account will now be autoconnected every time on startup (if auto-connect is enabled)

chrismotian:
@srossbach Provide arguments for your suggestions. Why should we expect that the user doesn't want to set the added account as active? As @fzieris pointed out if we connect to the newly added account the user get feedback if everything went well (is the account valid), and in his opinion, this is a good feedback. The user can take care if he wants to change the active account afterward because he already should know how to do this. In my opinion, creating notifications and windows for user interactions seem to be an overkill for simple actions like adding a new account because the actions I implemented (change the active account to the newly added account and connect to it) are expectable for a user in this scenario.

srossbach:
The rule is surprisingly simple. If I do not agree on some auto-connect stuff I do NOT want that the added account is connected in any way, nor it is set up as my default account. 

You normally create your first account through the Saros-Configuration wizard. That should be enough. Adding more accounts through this dialog should be considered as a special case.

PR #217, opened 2018-08-28
[BUILD] Fix the HTML-STF tests due to LinkageError

Jannis1337:
Removed unnecessary or duplicate libraries.


m273d15:
Has to be reopened, because of revert #269

PR #219, opened 2018-08-28
[BUILD] Add StartupSarosSTF to start the STF

ghost:
I added a StartupSarosSTF.java and a plugin.xml to the stf project. Now
the STF starts from the stf-project and not from the eclipse-project
like it was befor.


m273d15:
Has to be reopened, because of revert #269

PR #220, opened 2018-08-28
[BUILD] Update the swtbot from 2.0.5 to 2.7.0

ghost:
To coplile the Intellij-Project I need a Java 1.7 compiler. Eclipse Juno
is the first eclipse ide which provides java7 compiler. So for STF
development I need Juno but the old swtbot is not compatible with Juno.
That is the reson why I updated the swtbot.


m273d15:
I think the following steps are required before we can use this PR:
- [ ] adjust the test environment (docker images)
     - [ ] Adjust java version ??
     - [ ] Use eclipse juno instead of 3.7
- [ ] test all stf tests

m273d15:
out-dated and already implemented by #286

PR #221, opened 2018-08-28
[BUILD] Clean up hamcrest libraries

ghost:
There was a issue with hamcrest in some stf tests. I had to switch back
to hamcrest 1.1.0 instead of 1.3.0 which comes with the swtbot update.
Also I removed the libraries from the eclipse project.


m273d15:
I was not able to reproduce the issue with hamcrest during stf testing. All tests were successful after switching to the new version with #286 .

PR #228, opened 2018-09-19
[STF] Add checkbox, radio group, select, multi select and progressbar

jafey:


m273d15:
Same PR as #189 but with the right base.

PR #230, opened 2018-09-19
[STF][RREFACTOR] Extract view enum

jafey:
I extracted the view Enum from IRemoteHTMLView to a new file. Also i
removed that View-Key map and build fields on the enum.

Change-Id: I830e92428b7840ff2b1dfe5ea5dc87ab2a8c4834


jafey:
same as #191

PR #231, opened 2018-09-19
[STF] Add StartSessionWizardTest

jafey:
I found out that it is not necessary to have a extra RemoteBotDialog
because a Dialog which contains an HTML view is also accessible via the
RemoteHTMLView. So I removed the RemoteBotDialog and wrote a test which
uses a RemoteHTMLView to access HTML-Elements in the dialog window.

Change-Id: I61379e3af09927e1f709654c47f8deeb25cb70b4


m273d15:
same as #192

PR #239, opened 2018-09-27
Remove configs that are not used by gradle

m273d15:
Eclipse and Intellij are using the Gradle project model and dependency resolution. Therefore the project files are not needed anymore. Furthermore the Ant build description and libraries are not required anymore.


m273d15:
Will be replaced by a similar PR based on #286

Issue #233, opened 2018-09-19
Saros sometimes creates ChecksumActivities for deleted/moved files


tobous:
The [`ConsistencyWatchdogServer`](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/watchdog/ConsistencyWatchdogServer.java) running on the host periodically creates ChecksumActivities and broadcasts them to all other session participants. The host determines which files to send these activities for by checking which files are open in a local or remote editor (see [lines 214-219](https://github.com/saros-project/saros/blob/88bcb01524978c5c5e01f60b7d0ae658951dc18a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/watchdog/ConsistencyWatchdogServer.java#L214).

This is problematic when such ChecksumActivities are triggered after a file was deleted/moved by the host but before these send activities for the deletion/move were executed by the clients and at least one of the clients has an open editor for the deleted/moved file. In such a case, the host still sends a ChecksumActivity for the deleted file/the old path of the moved file, as there are still remote editors for the file. Once all clients executed the send deletion/move activities and subsequently closed the editors with the old file path, the host will no longer generate ChecksumActivities for these files.

As activities are executed in the correct order, these faulty Checksum activities are executed after the deletion/move activities, meaning the clients no longer have a matching file for the ChecksumActivitiy. This leads to an inconsistency where the clients think they are missing a file that is present on the host's side. This inconsistency can not be resolved through the recovery as the host fails when trying to read the content of the "missing" file requested by the clients (as the file no longer exists).

This bug triggers consistently for Saros/I but does not seem to trigger with the current setup of Saros/E. It is hard to reproduce for Saros/E as it is dependent on the timing of checksum activities. With my setup, I was able to somewhat consistently trigger it on Saros/E by reducing the `CHECKSUM_CALCULATION_INTERVAL` to 200 (from it's original value of 10000).

#### Steps to Reproduce
Needed setup: Reduce the `CHECKSUM_CALCULATION_INTERVAL` in the [`ConsistencyWatchdogServer`](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/watchdog/ConsistencyWatchdogServer.java) to 200.

1. Alice and Bob start a session with a file A
2. Bob opens file A
3. Alice deletes/moves the file A

[...]

PR #246, opened 2018-09-30
[FIX][CORE] #233 Correctly create ChecksumActivities for missing files

tobous:
Adds a check on the host side to ensure that existing checksums for
files are only used if the file still exists locally.

Adds a check on the client side to ensure that checksums for files
that don't exist on both sides don't trigger an inconsistency.

### DIFF
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/watchdog/ConsistencyWatchdogServer.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/watchdog/ConsistencyWatchdogServer.java
@@ -257,8 +257,8 @@ public class ConsistencyWatchdogServer extends AbstractActivityProducer
          * Ensures that the watchdog server doesn't use outdated checksums for
          * files that no longer exist locally.
          */
-        if (!docPath.getFile().exists()
-                && checksum.getHash() != DocumentChecksum.NOT_AVAILABLE) {
+        if (checksum.getHash() != DocumentChecksum.NOT_AVAILABLE && !docPath
+                .getFile().exists()) {
 
             LOG.debug("Updating checksum for " + docPath + " to correctly "
                     + "represent that the file no longer exists locally: "
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/watchdog/ConsistencyWatchdogServer.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/watchdog/ConsistencyWatchdogServer.java
@@ -257,9 +257,11 @@ public class ConsistencyWatchdogServer extends AbstractActivityProducer
          * Ensures that the watchdog server doesn't use outdated checksums for
          * files that no longer exist locally.
          */
-        if (!docPath.getFile().exists()) {
-            LOG.debug("Marking checksum for " + docPath
-                    + " as dirty as the file no longer exists locally: "
+        if (!docPath.getFile().exists()
+                && checksum.getHash() != DocumentChecksum.NOT_AVAILABLE) {
+
+            LOG.debug("Updating checksum for " + docPath + " to correctly "
+                    + "represent that the file no longer exists locally: "
                     + checksum);
 
             checksum.markDirty();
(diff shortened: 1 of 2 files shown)


de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/watchdog/ConsistencyWatchdogServer.java
```diff
@@ -253,7 +253,7 @@ private void updateChecksum(SPath docPath, Set<SPath> localEditors,
             documentChecksums.put(docPath, checksum);
         }
 
-        if (!checksum.isDirty())
+        if (!checksum.isDirty() && docPath.getFile().exists())
```
srossbach: Well for the Eclipse part I goes moving this check behind line 266 would do the trick, however, what will getContent return on IntelliJ for an editor that is not backed-up by file ?

Anyway this could use at least a comment why this file check is done here.

tobous: This check does not work. It is an oversight on my side. The checksum does not seems to be correctly marked as dirty, meaning [`DocumentChecksum.update(String)`](https://github.com/saros-project/saros/blob/7027c7e83c4816fe01a5633abbd95471cdfcf199/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/watchdog/DocumentChecksum.java#L114) wouldn't correctly update the checksum anyway, even if it is executed due to the added check. I will remove it. A better solution would be to correctly mark the checksum as dirty in the class handling the deletion/move.

`EditorManager.getContent(SPath)` correctly returns null. See [line 524](https://github.com/saros-project/saros/blob/7027c7e83c4816fe01a5633abbd95471cdfcf199/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorManager.java#L524). But, as already mentioned, it logs a "superfluous" warning.

tobous: As the checksums are not accessible from outside the `ConsistencyWatchdogServer`, setting the checksum for deleted/moved files as dirty becomes much harder. I would also have to add a bridge class (at least I think this would be classified as a bridge class) for the classes handling the filesystem modifications to abstract from the actual IDE classes.

I would rather use a more hacky solution: moving the check `docPath.getFile().exists()` into a separate if above the current line and set the checksum to dirty if the file does not exist. The if in line 256 can then be turned into an `else if`.

tobous:
The new patch seems to work for IntelliJ, but I still get the same error in Eclipse. The problem seems to be that the call `EditorManager.getContent(SPath)` still returns the file content, even though `IFile.exists()` says the file no longer exists. I could avoid this by adding

``` Java
checksum.update(null);
return;
```
after marking the checksum as dirty. This way, the content of the checksum would be guaranteed to be set to `DocumentChecksum.NOT_AVAILABLE`. This would avoid the call to `EditorManager.getContent(SPath)`.

tobous:
The latest suggested solution works if the host deletes/moves a file. If a client deletes/moves a file (in conjunction with the change resetting the VectorTime of deleted files), this leads to the situation that the client receives ChecksumActivities with a wrong VectorTime for the file. This does not really hurt as the problem should resolve itself when the host also deletes/moves the file and subsequently resets the vector time (the client never notices this problem), but it still logs a lot of unnecessary exceptions that clog up the log.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/watchdog/ConsistencyWatchdogServer.java
```diff
@@ -253,8 +253,23 @@ private void updateChecksum(SPath docPath, Set<SPath> localEditors,
             documentChecksums.put(docPath, checksum);
         }
 
-        if (!checksum.isDirty())
+        /*
+         * Ensures that the watchdog server doesn't use outdated checksums for
+         * files that no longer exist locally.
+         */
+        if (!docPath.getFile().exists()) {
+            LOG.debug("Marking checksum for " + docPath
```
srossbach: I find this log message very confusing because you will read it over and over again until the last affected user has started the recovery. You might think that marking the checksum as dirty has an actual benefit but you only have to do this because otherwise you could risk that the update(null) call might not be applied to the checksum.

tobous: You are right. I improved the logic to check whether the content of the checksum does not match the file state so that this check is only called once. Furthermore, I changed the log message to state that the checksum is updated instead of just stating that it is marked as dirty (as marking it as dirty does not do anything in itself, as you pointed out).

PR #252, opened 2018-10-07
[Internal] [WB] Allow model creation for all shapes but only render supported ones

AbdBarho:
the new html whiteboard contains some shapes that are not supported by the original whiteboard, such as circles
this change allows the new shapes to be accepted, saved, and synchronised with other users without crashing the original whiteboard


de.fu_berlin.inf.dpp.whiteboard/src/de/fu_berlin/inf/dpp/whiteboard/gef/part/ElementRecordPart.java
```diff
@@ -55,9 +57,24 @@ protected void createEditPolicies() {
...
 
+    /**
+     * @return supported children models that can be rendered by the original
+     *         whiteboard
+     */
     @Override
     public List<ElementRecord> getModelChildren() {
-        return getElementRecord().getVisibleChildElements();
+        List<ElementRecord> filtered = new ArrayList<ElementRecord>();
+        for (ElementRecord el : getElementRecord().getVisibleChildElements()) {
```
srossbach: either record or r but not e1

de.fu_berlin.inf.dpp.whiteboard/src/de/fu_berlin/inf/dpp/whiteboard/gef/part/ElementRecordPart.java
```diff
@@ -55,9 +57,24 @@ protected void createEditPolicies() {
             new ElementModelDeletePolicy());
     }
 
+    /**
+     * @return supported children models that can be rendered by the original
```
srossbach: What is the original whiteboard ? I am sure you mean the Eclipse Whiteboard that uses the GEF

AbdBarho:
@srossbach can you please give me more information on how the old whiteboard will still crash? on my machine I tested it and everything worked fine.

srossbach:
Old Whiteboard = All versions contained in our releases. As you have no chance to patch them and you said this patch prevent crashes you can assume that you will crash all former WB versions.

AbdBarho:
the first and last WB version that came out is the original one from 2010.
There was some work done on the WB in 2012 but was abandoned at the end.

https://saros-build.imp.fu-berlin.de/gerrit/#/c/595/

technically speaking, there are no other versions of the GEF Whiteboard

AbdBarho:
Never mind, I guess I understand now what you meant.

if user A updates to the latest Saros Version and uses the HTML WB, while user B is still using the older version without the changes, user B's whiteboard will crash.

that is actually a very valid and critical point that went over my head when implementing.
what are your suggestions for approaching this problem?
mentioning this in the commit message is enough?

technically, this patch prevents the crash from even happening, the crash is actually caused by the HTML WB itself and not this commit.

srossbach:
There is nothing you can do to prevent this issue. I think it is ok to mention this in the commit message of the patch that will actually cause this issue (have not seen it yet as PR).

The next Saros release will be incompatible anyway so this is not really an issue.

PR #250, opened 2018-10-04
[Internal] [WB] Added BrowserSXEBridge and BrowserActionExecuter

AbdBarho:
the BrowserSXEBridge is responsible for the communication between the sxe controller of the whiteboard and the views in the browser, received messages are redirected to the browser where they get handled.
Browser ActionExecuter receives an action object form the browser which is then applied to the sxe controller.

PR #253, opened 2018-10-07
[Internal][WB] Expose certain properties to GSON

AbdBarho:
the records need to be sent to the browser so that it stays in sync with the state in Java.
converting all properties of these records will cause stackoverflows because of circular references.
by exposing certain properties with the @Expose annotation, only these get to be converted to json and sent to the browser.


AbdBarho:
can only compile correctly after #250 is merged

Issue #251, opened 2018-10-04
Saros/I does not save open editors before session negotiation on the host side


tobous:
Saros/I does not save the open editors on the host side belonging to the shared module before starting the project negotiation. This leads to inconsistencies when the editor content for one of the shared files does not match the content of the file on disk as the content of the file on disk is shared.

#### Steps to reproduce
To consistently reproduce this behavior, you will first have to disable IntelliJ automatic saving mechanisms. To do so, go into the Intellij settings. These, choose the tab **Appearance & Behavior** > **System settings** and un-select the following options:
- *Save on frame deactivation*
- *Synchronize files on frame or editor tab activation*
- *Save files automatically if application is idle for X sec.*

As the initial state, Alice and Bob both have a file A with identical content.

1. Alice changes the file A *without* saving it.
2. Alice and Bob start a session sharing the file A.

This will lead to Bob starting the session with the old content for file A, resulting in an inconsistent state.

srossbach:
The Eclipse logic can be found here: https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/wizards/AddProjectToSessionWizard.java#L173

I think most parts have to be put into the core / IPN

tobous:
As far as I can tell, this is already done by the core in AbstractOutgoingProjectNegotiation:
https://github.com/saros-project/saros/blob/21f19b481e225d8d6cbc20d2be9f91f01947761f/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractOutgoingProjectNegotiation.java#L338
`EditorManager.save(IProject)` just did not work correctly in Intellij as it way relying on the internal list of open editors that was not updated correctly. This issue should be fixed with the PR #254.

Maybe the same issue also applies to Saros/E (which is why additional logic was implemented)?

Or am I misunderstanding your comment?

srossbach:
I missed the "on the host side" part 🤦‍♂️ 
I thought you were talking about the client side.
My reply is basically a response to your solution you presented in #265

tobous:
For the issue of moving the logic for saving editors on the client side during project negotiation, I already created the issue #232.

PR #254, opened 2018-10-08
[FIX][I] #251 Fix EditorManager.saveEditors(...)

tobous:
EditorManager.saveEditors(IProject) now uses the EditorPool to determine
which editors are open instead of holding a separate list. This separate
list was superfluous and was leading to issues when trying to save the
open Editors, as the list was not updated correctly, leading to
inconsistencies when starting a session while having an unsaved file
open.

Subsequently removes the hashmap used to store the open editors and
adjusts other methods using this second editor store.


tobous:
This class was initially copied from Saros/E. This logic is a remnant of this as it also exists in Saros/E. I did some digging and it was introduced that way in 2006 with https://github.com/saros-project/saros/commit/f0e78dd2fb6384aa71c0afdd1f9794bc369c9f1d. (The initial commit starting this repos https://github.com/saros-project/saros/commit/155cdba93ae04cf2b3ffaee86b5196bcad072631 did not have this code, but all files of this commit were removed with the next commit and a new version of Saros was introduced with the commit mentioned above.) The EditorPool was already present at that point, but maybe it did different things? I don't know.

I don't know why this divide exists, either. I just though it was an oversight while implementing Saros/I (or rather while adjusting the class for IntelliJ).

From my point of view, the initial implementation of `saveEditors(IProject)` does not make much sense either, as it does not match the [`IEditorStateManager` javadoc](https://github.com/saros-project/saros/blob/9fc29671a436b5cbd9d72ddc657abc425e3900cc/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/editor/IEditorManager.java#L65):
It tries to also save all editors contained in the `UserEditorStateManager`. This is somewhat pointless as the `UserEditorStateManager` only stores remote editors, meaning there does not have to be a local editor for the file. Furthermore, the javadoc of `IEditorStateManager` explicitly states that the method is supposed to save *local* editors.

@srossbach @fzieris Do you know why the EditorManager is holding a separate list of all open editors, even though it also holds the EditorPool? Do you know what exactly `IEditorManager.saveEditors(IProject)` is supposed to do?

srossbach:
The EditorPool is the collection of all editors open by the local. It is used to keep track and install various listeners. On the other hand UserEditorStateManager keeps track on the "remote editors". If you do not know the internal you might think, why I need the UserEditorStateManager ? It is needed because text edits are applied through the editor, even it is not opened on the local side. So it does INDEED only save / flush the local editors (I did not check if the save is broadcasted but I guess it is not)

As for your other question, I really cannot tell you why there is a EditorPool and also a separate list without digging through the EditorManager code.

As for IEditorManager.saveEditors(IProject) after I call this method for a given / or all projects -> all dirty / cached editor contents for files that are part of the sharing have to successfully written to the underlying storage / filesystem -> If I read the file through the filesystem it will have the same content as if I would access the editor content through the given editor provider.

tobous:
Ok, then I interpreted the javadoc to pedantically. The used wording does not really apply to Intellij as the Editors don't hold the content of the file, they are only the interface the user interacts with. The actual content is stored in [Documents](http://www.jetbrains.org/intellij/sdk/docs/basics/architectural_overview/documents.html). There can't be any unsaved "editors" that don't exist locally.

But that is not the point. As already mentioned in #255, I don't think flushing to disk instead of reading from the cache directly is not the right way to go. But I also think that saving now and again is a good idea, just in case the application crashed or there is a problem with the caching logic. I will adjust this logic to save all "open" editors.

srossbach:
The Eclipse stuff does not differ that much from IntelliJ

https://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2Feditors_documents.htm

Basically you have an IEditorpart (the visible stuff) and an IEditorInput where the content is fetched and stored. Between those two interfaces you have a document provider that gives you access to a document instance that is hold in memory containing the editor contents.

I really do not know why everything is prefixed with Editor in our logic as some logic operates on documents but the logic was written long ago.

I do not know if it makes sense to split things, so having an EditorManager and a DocumentManager, I guess this will be confusing as well.

tobous:
@srossbach Do you still have any issues with the PR?

I think splitting the EditorManager in general is a very good idea. It is way to big and does way to much. (For IntelliJ) It executes TextEdit-, Editor, TextSelection-, and Viewport-Activities. Furthermore, it dispatches some EditorActivities, provides some editor utility functions, and it provides functions to generate and dispatch the above mentioned activities (even though these functions are never used inside the class itself).

Splitting of the document stuff would be easy-ish as it (almost) exclusively concerns the part dealing with TextEditActivities. The only problem is that saving also concerns the documents, as this can't be done with an editor. You can easily get the document for an editor, but this would mix concerns again. A solution would be to also listen for SAVED EditorActivities in the document part.

But, as I have already said (too) many times before: I don't think this hurts enough (at least for IntelliJ) for me to deal with it right now. There are still many parts of the basic Saros functionality that are broken or missing in Saros/I.

tobous:
@srossbach Do you still have any issues with this PR? Otherwise, I would appreciate you to approving it?

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorHandler.java
```diff
@@ -196,17 +197,41 @@ public void removeEditor(@NotNull SPath path){
...
-     * @param path
+     * @param path the path for the document to save
      */
-    public void saveFile(SPath path) {
-        Document doc = editorPool.getDocument(path);
-        if (doc != null) {
-            projectAPI.saveDocument(doc);
-        } else {
-            LOG.warn("Document does not exist: " + path);
+    public void saveFile(
```
srossbach: Why not saveDocument ?

tobous: Done.

PR #255, opened 2018-10-08
[INTERNAL][I] Switch to using Document content in IFile.getContents()

tobous:
IntelliJFileImplV2.getContents() now tries to get a Document for the
file before falling back to using the content of the file on disk. This
behavior is preferable as the Document might contain cached changes to
the file that were not written to disk yet.

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/filesystem/IntelliJFileImplV2.java
+++ b/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/filesystem/IntelliJFileImplV2.java
@@ -7,7 +7,10 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.file.FileAlreadyExistsException;
 
+import com.intellij.openapi.editor.Document;
+import com.intellij.openapi.fileEditor.FileDocumentManager;
 import org.apache.commons.io.IOUtils;
+import org.apache.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -23,6 +26,9 @@ import de.fu_berlin.inf.dpp.filesystem.IResource;
 
 public final class IntelliJFileImplV2 extends IntelliJResourceImplV2 implements
     IFile {
+
+    private static final Logger log = Logger
+        .getLogger(IntelliJFileImplV2.class);
 
     private static final int BUFFER_SIZE = 32 * 1024;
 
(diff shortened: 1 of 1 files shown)


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/filesystem/IntelliJFileImplV2.java
```diff
@@ -174,7 +180,20 @@ public InputStream getContents() throws IOException {
             throw new FileNotFoundException(this + " does not exist or is " +
                 "derived");
 
-        return file.getInputStream();
+        Document document = Filesystem.runReadAction(
+            () -> FileDocumentManager.getInstance().getDocument(file));
+
+        if (document == null) {
```
Drakulix: Is there any documentation when this might happen? It seems that there is no exception thrown, so if you did encounter `document` actually being `null`, I would like to see a comment about it.

tobous: I don't think a comment is necessary in this case as the check is based on the [`FileDocumentManager.getDocument` javadoc](https://github.com/JetBrains/intellij-community/blob/36f1ada7ed36bfd72f854677fca58105fc6abd2a/platform/core-api/src/com/intellij/openapi/fileEditor/FileDocumentManager.java#L48):
```
@return the document, or null if the file represents a directory, or is binary without an associated decompiler, or is too large.
```

Drakulix: Fair enough. That is easy to find in case somebody else might be wondering.

tobous:
@srossbach I kind of get your point, but I don't get why we are relying on the file system with this part of the Saros logic. We basically use the editor contents (which are represented by [Documents](http://www.jetbrains.org/intellij/sdk/docs/basics/architectural_overview/documents.html) for Intellij) for everything except calculating checksums and the session negotiation (at least from what I can tell when looking at where `IFile.getContents()` is used). It does not make sense to me to use the files on disk directly for these purposes in the case of IntelliJ. This only leads to inconsistency problems and flushing everything to disk seems like a suboptimal solution to me. Intellij presumably provides the whole caching mechanism to minimize write actions. Why not rely on this system provided by the IDE so that we don't have to deal with these inconsistency problems at all? Why should we force a flush to disk if it can be avoided? The IDE should provide the correct logic ensuring that the cached content is correctly flushed to disk when necessary.

I created this patch to solve the problem you mentioned in #254. In Saros/I, we also write to documents that are not currently open in an editor locally. But this is only an issue if we rely on the content of the file system. If we only use the cached version of the documents, this should not be an issue.

tobous:
Ok, now that I am looking at the class, this seams like a halve measure as it only adjust one method. I would still like to adjust this class to prioritize using the cached version of the files, but this can be done later on (and should include an adjustment of all applicable classes and the javadoc).

PR #262, opened 2018-10-11
[FIX][I] Check if document is shared instead of relying on EditorPool

tobous:
When reacting to a DocumentEvent in StoppableDocumentListener, the logic
now tries to create an SPath for the modified document if it is not
known to the editor pool. This allows us to correctly handle local
edits to files that are not currently open in an editor. Such a
situation can occur when IntelliJ automatically modifies a document, for
example during refactorings or moving/renaming classes.

As we no longer rely on the editor pool to filter which documents are
shared, the created SPath is checked with
SarosSession.isShared(IResource) to ensure that we only create
TextEditActivities for shared files.


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/StoppableDocumentListener.java
```diff
@@ -20,25 +22,55 @@
...
+
         if (path == null) {
-            LOG.debug("Event for document " + event.getDocument()
-                    + " ignored: document is not known to the editor pool");
-            return;
+            VirtualFile virtualFile = FileDocumentManager.getInstance()
+                .getFile(document);
+
+            if (virtualFile == null) {
+                LOG.debug("Ignoring event for document " + document
```
srossbach: Are you tracking all files in the IDE ? If yes this could flood the log file a little bit depending on your sharing setup, e.g you do a partial sharing with only a few files.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/StoppableDocumentListener.java
```diff
@@ -1,16 +1,18 @@
...
+import com.intellij.openapi.vfs.VirtualFile;
 import de.fu_berlin.inf.dpp.activities.SPath;
 import org.apache.log4j.Logger;
 
 /**
- * Tracks modifications of documents opened in editors.
- * <p/>
- * Notify its EditorManager of modifications in an opened document, before
- * the modification has actually happened.
+ * Tracks modifications of Documents and triggers matching TextEditActivities.
```
srossbach: Just FYI

In Eclipse

1. IIRC we act on after change
2. Refactoring changes are mostly broadcasted through file activities if there is not open editor / connected document provider for a given file.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/StoppableDocumentListener.java
tobous: This patch does not increase the number of created log entries. Everything that is logged now would have just been logged beforehand as "... document is not known to the editor pool". But yes, you are right. This is tracking all files in the IDE. This will create a lot of logs, especially when working on non-shared files. I will set it to trace. (I have to modify this patch anyway as I can merge the last two if clauses.)

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/StoppableDocumentListener.java
tobous: Regarding point 1: I just left it the way it was. Is it important to do it after the change? This could be changed easily, but I don't know if this could cause any issues.

As a side note: The new filesystem handling also acts before files are deleted/moved/renamed. Is it important for this part of the logic (also thinking about the future goal of cross-IDE compatibility)?

Regarding point 2: This is not the case in IntelliJ. I added this logic as text edits changing the package defined in the java were not send as there was no editor open for the file at the point they were created. I would guess that IntelliJ closes the old editor, moves the file, adjusts the content through a document for the new file and then re-opens the moved file in an editor.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/StoppableDocumentListener.java
```diff
@@ -50,7 +50,7 @@ public void beforeDocumentChange(DocumentEvent event) {
                 .getFile(document);
 
             if (virtualFile == null) {
-                LOG.debug("Ignoring event for document " + document
+                LOG.trace("Ignoring event for document " + document
```
srossbach: Just for the future, generating the log message even it is not logged requires 3 string concats and a toString method call. I do not think it will haven that often and consumes much CPU time but you should really think about

if (LOG.isTraceEnabled()) … before such log statements

PR #263, opened 2018-10-11
Clean up StoppableDocumentListener

tobous:
Moves the functionality of the classes startListening() and stopListening() into setEnabled(boolean).

Removes an empty implementation of an interface class that has become unnecessary with the introduction of the default modifier for interface methods.

Makes the class package-private as it is not accessed outside the package.


tobous:
@srossbach had to rebase on master. The code did not change. Could you approve the PR again?

PR #264, opened 2018-10-12
Fix linting and run linter checks in Travis

donut87:
I took the liberty of fixing the 70 something linting problems.
To never run again in a situation like this linting is executed on Travis and would *break the build* if rules are violated.


donut87:
From my point of view, yes.

de.fu_berlin.inf.dpp.ui.frontend/html/src/components/App.jsx
```diff
@@ -1,19 +1,19 @@
...
+import React from 'react'
+import StartSessionWizardView from './StartSessionWizardView'
 
 const viewComponents = {
   [views.MAIN]: MainView,
   [views.ADD_CONTACT]: AddContactView,
   [views.START_SESSION_WIZARD]: StartSessionWizardView,
   [views.CONFIGURATION_WIZARD]: ConfigurationWizardView,
-  [views.BASIC_WIDGET_TEST]: BasicWidgetTestView
+  [views.BASIC_WIDGET_TEST]: BasicWidgetTestView,
```
AbdBarho: I went through the linter settings and this is how the linter would want it.

but it does not make sense to allow trailing commas, in older version of web browser (e.g: IE8) trailing commas will cause an exception, I suggest that the settings of the linter gets changed and the rule to comma-dangle should be set to "never"

de.fu_berlin.inf.dpp.ui.frontend/html/src/components/BasicWidgetTestView/index.jsx
```diff
@@ -57,8 +57,8 @@ export default class BasicWidgetTestView extends React.Component {
     } else {
       value = e.target.id
     }
-    $('#button-display-text').html(value)
-    console.log(value)
+    document.querySelector('#button-display-text').innerHtml = value
```
AbdBarho: did the linter say that you have to change from JQuery to Vanilla JavaScript? that is weird

de.fu_berlin.inf.dpp.ui.frontend/html/webpack.config.js
```diff
@@ -13,7 +13,7 @@ const createPage = (name) => {
   return new HtmlWebpackPlugin({
     filename: `${name}.html`,
     title: name,
-    template: 'template.ejs'
+    template: 'template.ejs',
```
AbdBarho: build configurations files dont have to be linted, our compiled GUI will run in browser but the config files will run in node.js environment

de.fu_berlin.inf.dpp.ui.frontend/html/src/components/App.jsx
donut87: Should I change the linting rules in this PR or is this a topic of another PR? Is there a need to involve a wider range of people in this decision?

de.fu_berlin.inf.dpp.ui.frontend/html/webpack.config.js
donut87: These 'configuration files' are code. There should be a standard on how this JS code should look, and I think it should look exactly the same as all the other JS code. 
Linting is not something we are doing for the browsers. We use this to make our code more readable. This is why I would like to keep this included in the linting process.

de.fu_berlin.inf.dpp.ui.frontend/html/src/components/App.jsx
donut87: Nevermind. I read your request.

de.fu_berlin.inf.dpp.ui.frontend/html/src/components/BasicWidgetTestView/index.jsx
AbdBarho: okay, but just for future changes bear in mind that the STF testing requires JQuery.

de.fu_berlin.inf.dpp.ui.frontend/html/webpack.config.js
AbdBarho: I just find it weird to lint both code and config which run in different environments with the same rules, anyway this is not an issue, you dont have to change it

de.fu_berlin.inf.dpp.ui.frontend/html/src/components/BasicWidgetTestView/index.jsx
donut87: The tests are green. Otherwise, Travis would have complained already.

donut87:
Just a friendly reminder, that I cannot merge myself and the approve was nearly two days ago :-)

m273d15:
@donut87  Do you want to become a member of the organization?

PR #265, opened 2018-10-15
[I] Add hack to avoid replicating activities during project negotiation

tobous:
Include the check if there currently is a file replacement in progress
when dispatching text edit activities. This is the same workaround
also used in the SharedResourcesManager to prevent Saros sending
activities for changes made as part of the synchronization during the
project negotiation.

Introduces trace logging when local activities are ignored because a
local file replacement is currently in progress.


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorManager.java
```diff
@@ -769,6 +773,16 @@ void generateTextEdit(int offset, String newText, String replacedText,
             return;
         }
 
+        if (fileReplacementInProgressObservable.isReplacementInProgress()) {
+            if (LOG.isTraceEnabled()) {
```
m273d15: Why do you use `isTraceEnabled` ? This is only useful if you want to avoid to execute an expensive task even if Trace is disabled. In your cases you simply concat strings.

tobous: Maybe it might not be necessary in this case as it should not happen that often, but I included this check in reaction to a comment by @srossbach on #262:

> Just for the future, generating the log message even it is not logged requires 3 string concats and a toString method call. I do not think it will haven that often and consumes much CPU time but you should really think about
> 
> if (LOG.isTraceEnabled()) … before such log statements

tobous:
Will adjust the title/description.

This problem always applies (at least for IntelliJ), even with the current logic handling receiving text activities: https://github.com/saros-project/saros/blob/21f19b481e225d8d6cbc20d2be9f91f01947761f/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorManipulator.java#L188

In general, we need some kind of logic to match a received activity to the effect that it has on the local environment/the local listeners. Otherwise we will always have the problem that we will have to disable the listeners for certain types of changes when applying activities of the same type.

The activities during the project negotiation are just handled differently (as far as I can tell) because they are triggered in the core and there is no representation of the specific listeners (filesystem and editor listener) in the core, meaning we can't explicitly disable the matching listeners (which is what we do in every other case).

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorManager.java
m273d15: Ok, I had the interface of slf4j and log4j2 in mind that supports a method `trace(String formatString, params ...)` that evaluates the format string only if the corresponding log level is set, but this is currently not supported in our environment.

srossbach:
> This problem always applies (at least for IntelliJ), even with the current logic handling receiving text activities:
> 
> [saros/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorManipulator.java](https://github.com/saros-project/saros/blob/21f19b481e225d8d6cbc20d2be9f91f01947761f/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorManipulator.java#L188)
> 
> Line 188 in [21f19b4](/saros-project/saros/commit/21f19b481e225d8d6cbc20d2be9f91f01947761f)
>  manager.disableDocumentListener(); 
> 
Same applies for Eclipse:

https://github.com/saros-project/saros/blob/21f19b481e225d8d6cbc20d2be9f91f01947761f/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/editor/EditorManager.java#L735

Do you have a better solution ? Disabling the listener is simple. The first approach we found in Saros/E as a left over was to check if a TextEditActivity equals the former one and then avoid to send it. Far more complicated and error prone. And as long as you do not have any concurrency issue  or asynchronous execution this is really the way to go in my opinion.

But I think for the FileSystem we have concurrency (at least Eclipse) and asynchronous notifications issues.

The Eclipse broadcasting is actually very tricky (I do not know if this was improved and later versions) but as far as I had working on IDeltas I can tell that broadcasting can still happen at any time and more than worse that the broadcasting can happen by a different thread, e.g 99% of the time the broadcast is done in the thread that modifies the file and 1% of the time it is done in another thread. Happened to me at least with some old SVN plugins.

srossbach:
Only one question, why does the activity queuing not work for Saros/I ?

tobous:
I am assuming that this would mess up the activity queuing for text edit activities during the project negotiation as these activities are never fired. But I am not sure how the queuing is handled.

As to why it did not work beforehand: I don't know. I have not investigated it since I found it during my bachelor's thesis about a year ago. As the basic session did not function, changes during the negotiation breaking everything did not seem like a big deal. I created an [issue 891](https://sourceforge.net/p/dpp/bugs/891/) for this in Sourceforge but did not migrate it, apparently because I could no longer recreate the issue. So maybe it does work and I am just misremembering, but I would not count on it.

But, if the ticket is correct, these activities apparently also weren't queued, but rather just dropped instead (causing an inconsistency). They merely no longer crashed the session in unpredictable ways.

Currently, my work on the implementation dealing with filesystem changes in IntelliJ (re-implementing [SharedResourcesManager](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/project/SharedResourcesManager.java) and [FileSystemChangeListener](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/project/FileSystemChangeListener.java)) is taking its time (the functionality is there, it just needs some cleanup and a bit more testing). But I will check this behavior again once I am done with the rework.

srossbach:
The queuing is really simple. During the project negotiation the host "locks" the session. I put it up in quotes because this is the duty of the implementation which is not that easy. The client then start queuing (I hope we already have project based queuing), send it to the host who releases the session lock.


Every activity that operates on resource (i.e have an IPath) is queued and flushed when the negotiation is done.

PR #269, opened 2018-10-16
Revert new stf project adjustments

m273d15:
The reverted commits have the following problems:
* The new projects are not build during the CI process.
* The modularization is not complete (e.g. stf resources are still in the eclipse module).
* The stf tests are not executable/executed within travis, because the rmi server is not started. See [here](https://travis-ci.com/saros-project/saros/jobs/152031085) (The eclipse instance is started and the script waits for the rmi server, but the server is not started)


m273d15:
I tested this PR locally. The builds, tests and stf tests work.

srossbach:
@m273d15 Do you have an URL where a developer can find the actual regression results for the STF ?

m273d15:
@srossbach You have to look in the [travis build history](https://travis-ci.com/saros-project/saros/builds) and select the last `cron` build. It is [not possible to filter](https://github.com/travis-ci/travis-ci/issues/8531) all cron jobs, but I hope travis will add such a feature in the future.

The parent build job (that triggers stf-tests and stf-self-tests) is always reported as green, but the children builds show the real build result. The parent is always green, because the build result is reported on the last commit of the master branch. Therefore a history of red stf tests would lead to a master branch that contains only commits with build failures.

During my local testing I found approximately 5-7 tests that are unstable and fail sometimes.
My plan concerning the stf tests is to isolate the unstable tests, report the real result to the master branch and write an email to the mailinglist. After that I will rework the isolated tests. If you have a better idea I would propose to move this conversion in an issue.

srossbach:
One last question, the log files (not the ones of the build job) of the regression, i.e the logs that are written by the Eclipse instances and the screenshots of failed tests are located where ?

srossbach:
Ah now I understand this trendy PaaS and SaaS technology. You must use multiple of them to get things done. I really wish we would still have access to our Gerrit and Jenkins installations. The workflow was somewhat easier.

m273d15:
:joy: That is true, but in order to decouple our project from the FU infrastructure there are only a few alternatives.

PR #272, opened 2018-10-20
[CORE] Successive and combined project negotiations

stefaus:
Currently Saros creates a new project negotiation for every individually
added project resource. This leads to multiple concurrent negotiations
per user, which creates more overhead for transportation and user
interaction for every single negotiation. Additionally, it leads to
unexpected behavior because a new negotiation will transfer all files
currently not available on the remote machine, including files from
unfinished negotiation transfers.

This patch ensures that only one project negotiation per user is running
and all new shared resources are queued till all ongoing negotations
finished, before starting a new negotiation with all queued resources /
projects combined.

### DIFF
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ProjectNegotiationCollector.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ProjectNegotiationCollector.java
@@ -1,0 +1,58 @@
+package de.fu_berlin.inf.dpp.negotiation;
+
+import de.fu_berlin.inf.dpp.filesystem.IProject;
+import de.fu_berlin.inf.dpp.filesystem.IResource;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+
+/**
+ * This class collects resources that should be added to a session to add them at once in
+ * preparation of a new single project negotiation. The reason is to prevent multiple concurrent
+ * running project negotiations per user.
+ */
+public class ProjectNegotiationCollector {
+  private Map<IProject, List<IResource>> mapping = new HashMap<IProject, List<IResource>>();
+
+  /**
+   * Add project resource mappings for the next project negotiation.
+   *
+   * @param projectResourcesMapping project resource mappings to add
+   */
+  public synchronized void add(Map<IProject, List<IResource>> projectResourcesMapping) {
+    for (Entry<IProject, List<IResource>> mapEntry : projectResourcesMapping.entrySet()) {
+      final IProject project = mapEntry.getKey();
+      final List<IResource> resources = mapEntry.getValue();
+
+      boolean alreadyFullShared = mapping.containsKey(project) && mapping.get(project) == null;
+
+      /* full shared project */
+      if (resources == null || alreadyFullShared) {
+        mapping.put(project, null);
+        continue;
+      }
+
+      /* update partial shared project */
+      List<IResource> mappedResources = mapping.get(project);
+      if (mappedResources == null) {
+        mapping.put(project, new ArrayList<IResource>(resources));
+      } else {
+        mappedResources.addAll(resources);
+      }
+    }
+  }
+
+  /**
+   * Returns a list of project resource mappings that should be handled by a project negotiation.
+   * Resets the Collector for next additions.
+   *
+   * @return project resource mappings to add
+   */
+  public synchronized Map<IProject, List<IResource>> get() {
+    Map<IProject, List<IResource>> tmp = mapping;
+    mapping = new HashMap<IProject, List<IResource>>();
+    return tmp;
+  }
+}
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ProjectNegotiationObservable.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ProjectNegotiationObservable.java
@@ -106,4 +106,18 @@ final class ProjectNegotiationObservable {
 
     return currentNegotiations;
   }
+
+  /**
+   * Returns <tt>true</tt> if no running negotiations exist.
+   *
+   * @return <tt>true</tt> if no running negotiations exist
+   */
+  public synchronized boolean isEmpty() {
+    for (List<ProjectNegotiation> negotiationList : negotiations.values()) {
+      if (!negotiationList.isEmpty()) {
+        return false;
+      }
+    }
+    return true;
+  }
 }
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
@@ -32,6 +32,7 @@ import de.fu_berlin.inf.dpp.negotiation.NegotiationListener;
 import de.fu_berlin.inf.dpp.negotiation.NegotiationTools.CancelOption;
 import de.fu_berlin.inf.dpp.negotiation.OutgoingSessionNegotiation;
 import de.fu_berlin.inf.dpp.negotiation.ProjectNegotiation;
+import de.fu_berlin.inf.dpp.negotiation.ProjectNegotiationCollector;
 import de.fu_berlin.inf.dpp.negotiation.ProjectNegotiationData;
 import de.fu_berlin.inf.dpp.negotiation.SessionNegotiation;
 import de.fu_berlin.inf.dpp.negotiation.TransferType;
@@ -47,6 +48,7 @@ import de.fu_berlin.inf.dpp.preferences.IPreferenceStore;
 import de.fu_berlin.inf.dpp.preferences.PreferenceStore;
 import de.fu_berlin.inf.dpp.session.internal.SarosSession;
 import de.fu_berlin.inf.dpp.util.StackTrace;
+import de.fu_berlin.inf.dpp.util.ThreadUtils;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
@@ -104,6 +106,10 @@ public class SarosSessionManager implements ISarosSessionManager {
 
   private final ProjectNegotiationObservable currentProjectNegotiations;
 
+  private final ProjectNegotiationCollector nextProjectNegotiation =
+      new ProjectNegotiationCollector();
+  private Thread nextProjectNegotiationWorker;
+
   private XMPPConnectionService connectionService;
 
   private final List<ISessionLifecycleListener> sessionLifecycleListeners =
@@ -127,6 +133,12 @@ public class SarosSessionManager implements ISarosSessionManager {
         @Override
         public void negotiationTerminated(final ProjectNegotiation negotiation) {
           currentProjectNegotiations.remove(negotiation);
+
+          if (currentProjectNegotiations.isEmpty()) {
+            synchronized (nextProjectNegotiation) {
+              nextProjectNegotiation.notifyAll();
+            }
+          }
         }
       };
 
@@ -485,8 +497,56 @@ public class SarosSessionManager implements ISarosSessionManager {
    * @param projectResourcesMapping
    */
   @Override
-  public void addResourcesToSession(Map<IProject, List<IResource>> projectResourcesMapping) {
+  public synchronized void addResourcesToSession(
+      Map<IProject, List<IResource>> projectResourcesMapping) {
+    if (projectResourcesMapping == null) {
+      return;
+    }
 
+    /*
+     * To prevent multiple concurrent negotiations per user. 1. Collect all
+     * new mappings, 2. If active negotiations are running wait till they
+     * finish (collect new mappings in the meantime), 3. Create one
+     * negotiation with all collected resources.
+     */
+
+    nextProjectNegotiation.add(projectResourcesMapping);
+
+    if (nextProjectNegotiationWorker != null && nextProjectNegotiationWorker.isAlive()) {
+      return;
+    } else if (currentProjectNegotiations.isEmpty()) {
+      /* shortcut to direct handling */
+      startNextProjectNegotiation();
+      return;
+    }
+
+    /* else create a worker thread */
+    Runnable worker =
+        new Runnable() {
+          @Override
+          public void run() {
+            synchronized (nextProjectNegotiation) {
+              while (!currentProjectNegotiations.isEmpty()) {
+                try {
+                  nextProjectNegotiation.wait();
+                } catch (InterruptedException e) {
+                  Thread.currentThread().interrupt();
+                  return;
+                }
+              }
+            }
+
+            startNextProjectNegotiation();
+          }
+        };
+    nextProjectNegotiationWorker = ThreadUtils.runSafeAsync(log, worker);
+  }
+
+  /**
+   * This method handles new project negotiations for already invited user (not the first in the
+   * process of inviting to the session).
+   */
+  private synchronized void startNextProjectNegotiation() {
     ISarosSession currentSession = session;
 
     if (currentSession == null) {
@@ -506,9 +566,9 @@ public class SarosSessionManager implements ISarosSessionManager {
     }
 
     List<IProject> projectsToShare = new ArrayList<IProject>();
+    Map<IProject, List<IResource>> mapping = nextProjectNegotiation.get();
 
-    for (Entry<IProject, List<IResource>> mapEntry : projectResourcesMapping.entrySet()) {
-
+    for (Entry<IProject, List<IResource>> mapEntry : mapping.entrySet()) {
       final IProject project = mapEntry.getKey();
       final List<IResource> resourcesList = mapEntry.getValue();
 
@@ -516,8 +576,9 @@ public class SarosSessionManager implements ISarosSessionManager {
       if (!currentSession.isCompletelyShared(project)) {
         String projectID = currentSession.getProjectID(project);
 
-        if (projectID == null)
+        if (projectID == null) {
           projectID = String.valueOf(SESSION_ID_GENERATOR.nextInt(Integer.MAX_VALUE));
+        }
 
         currentSession.addSharedResources(project, projectID, resourcesList);
 
@@ -532,7 +593,6 @@ public class SarosSessionManager implements ISarosSessionManager {
     }
 
     INegotiationHandler handler = negotiationHandler;
-
     if (handler == null) {
       log.warn("could not start a project negotiation because no handler is installed");
       return;
@@ -541,32 +601,30 @@ public class SarosSessionManager implements ISarosSessionManager {
     List<AbstractOutgoingProjectNegotiation> negotiations =
         new ArrayList<AbstractOutgoingProjectNegotiation>();
 
-    synchronized (this) {
-      if (!startStopSessionLock.tryLock()) {
-        log.warn(
-            "could not start a project negotiation because the current session is about to stop");
-        return;
+    if (!startStopSessionLock.tryLock()) {
+      log.warn(
+          "could not start a project negotiation because the current session is about to stop");
+      return;
+    }
+
+    try {
+      for (User user : currentSession.getRemoteUsers()) {
+
+        TransferType type =
+            TransferType.valueOf(
+                currentSession
+                    .getUserProperties(user)
+                    .getString(ProjectNegotiationTypeHook.KEY_TYPE));
+        AbstractOutgoingProjectNegotiation negotiation =
+            negotiationFactory.newOutgoingProjectNegotiation(
+                user.getJID(), type, projectsToShare, this, currentSession);
+
+        negotiation.setNegotiationListener(negotiationListener);
+        currentProjectNegotiations.add(negotiation);
+        negotiations.add(negotiation);
       }
-
-      try {
-        for (User user : currentSession.getRemoteUsers()) {
-
-          TransferType type =
-              TransferType.valueOf(
-                  currentSession
-                      .getUserProperties(user)
-                      .getString(ProjectNegotiationTypeHook.KEY_TYPE));
-          AbstractOutgoingProjectNegotiation negotiation =
-              negotiationFactory.newOutgoingProjectNegotiation(
-                  user.getJID(), type, projectsToShare, this, currentSession);
-
-          negotiation.setNegotiationListener(negotiationListener);
-          currentProjectNegotiations.add(negotiation);
-          negotiations.add(negotiation);
-        }
-      } finally {
-        startStopSessionLock.unlock();
-      }
+    } finally {
+      startStopSessionLock.unlock();
     }
 
     for (AbstractOutgoingProjectNegotiation negotiation : negotiations)
@@ -715,8 +773,7 @@ public class SarosSessionManager implements ISarosSessionManager {
     boolean terminated = false;
 
     while (System.currentTimeMillis() - startTime < NEGOTIATION_TIMEOUT) {
-      if (currentSessionNegotiations.list().isEmpty()
-          && currentProjectNegotiations.list().isEmpty()) {
+      if (currentSessionNegotiations.list().isEmpty() && currentProjectNegotiations.isEmpty()) {
         terminated = true;
         break;
       }


de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -532,44 +550,105 @@ public void addResourcesToSession(
...
+        /**
+         * leave the synchronized block to allow new resource adding and wait
+         * till all running negotiations are done
+         **/
+        while (currentProjectNegotiations.negotiationsInProgress()) {
+            try {
+                synchronized (nextSharedResourcesWorkerNotify) {
+                    nextSharedResourcesWorkerNotify.wait();
+                }
+            } catch (InterruptedException e) {
```
saros-infrastructure: Bug Bug: Either re-interrupt this method or rethrow the "InterruptedException". ([squid:S2142](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3AS2142))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=272&issues=AWaT0XAZyONdBoB89fGy&open=AWaT0XAZyONdBoB89fGy)

[ISSUE_KEY]: AWaT0XAZyONdBoB89fGy

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -135,6 +146,13 @@ public void negotiationTerminated(final SessionNegotiation negotiation) {
         @Override
         public void negotiationTerminated(final ProjectNegotiation negotiation) {
             currentProjectNegotiations.remove(negotiation);
+
+            if (!currentProjectNegotiations.negotiationsInProgress()) {
+                /** notify about the completion of all running negotiations **/
+                synchronized (nextSharedResourcesWorkerNotify) {
+                    nextSharedResourcesWorkerNotify.notify();
```
saros-infrastructure: Bug Bug: "notify" may not wake up the appropriate thread. ([squid:S2446](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3AS2446))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=272&issues=AWaT0XAZyONdBoB89fGw&open=AWaT0XAZyONdBoB89fGw)

[ISSUE_KEY]: AWaT0XAZyONdBoB89fGw

srossbach:
> Additionally, it leads to
unexpected behavior because a new negotiation will transfer all files
currently not available on the remote machine, including files from
unfinished negotiation transfers.

Please fix this issue first. We can talk about enhancements later. But I do not see any point of improving the current system. You rarely share anything additional during the session, you can share multiple projects with the start of a session. Furthermore you cannot just change the logic of some methods to a blocking state. At least I have seen some issues when at least some user do not accept the negotiation (eg. gone AFK). You simply block anything for every user.

srossbach:
1. 95% of the use cases are = full sharing
2. Starting multiple project negotiation with redundant data seems to be a bug in the SessionManager (the fix should be trivial)
3, We have a wizard / dialog where you can select multiple resources that should be shared. The better fix is to disable this right click an share stuff.

Drakulix:
>     3. We have a wizard / dialog where you can select multiple resources that should be shared. The better fix is to disable this right click an share stuff.

Yes that would be a better fix, if the only intend of this PR would be fixing this bug.

But as said this simplifies the Instant Session Start implementation as well as my upcoming PR for non-host project sharing, which is needed for Saros Server.

@fzieris was also in favor of this change, as allowing multiple `ProjectNegotiation`s in parallel is not a very useful feature in the vast majority of cases.

This new limitation will simplify future code, addresses an existing bug - which does not mean we cannot improve the logic of the `SessionManager` anyway - and does not need to change the UI as with your suggestion.

I would argue changing the UI is a separate issue, which can be discussed independently of the PR, if you view the "right-click and share" as problematic. I personally think a change in the user experience is much more problematic and we should not need to do that, if we can still easily support the old UI with this PR.

That said I have not yet had the time to actually review the code, so this is not a statement over the overall code quality just on the issue of "not seeing any point of improving the current system".

stefaus:
> 1. 95% of the use cases are = full sharing

Yes we talk about a maybe rarely used feature, so the user impact should be low. 

> 2. Starting multiple project negotiation with redundant data seems to be a bug in the SessionManager (the fix should be trivial)

From my point of view: The problem for this case lies in the invitation logic, because a negotiation is trying to synchronize all project resources added to the session.
In `SarosSessionManager.addResourcesToSession(Map<IProject, List<IResource>> projectResourcesMapping)` new projects/resources are added to the Session and the belonging Project(s) marked for a Project Negotiation (`projectsToShare`)
https://github.com/saros-project/saros/blob/67f866410eb4318fa660c3077a34fd0b79c96fec/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java#L551-L554

than get handed to a new Project Negotation
https://github.com/saros-project/saros/blob/67f866410eb4318fa660c3077a34fd0b79c96fec/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java#L584-L586

in `AbstractOutgoingProjectNegotiation` the list of projects is used (without naming the resources of this negotiation) to create a file list
https://github.com/saros-project/saros/blob/67f866410eb4318fa660c3077a34fd0b79c96fec/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractOutgoingProjectNegotiation.java#L95-L96

`createProjectNegotiationDataList(...)` uses all the files added to the session in `SessionManager`
https://github.com/saros-project/saros/blob/67f866410eb4318fa660c3077a34fd0b79c96fec/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractOutgoingProjectNegotiation.java#L340-L344

If this would be called just with the added resources list from `SessionManager`, the diff part of the invitation logic would probably result in deleting all other (remote existing) project files.

> 3. We have a wizard / dialog where you can select multiple resources that should be shared. The better fix is to disable this right click an share stuff.

The Wizard allows the same use.

srossbach:
Well for partial sharing removed resources are discarded, see https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/FileListDiff.java#L61

Ok so this is my bad and the fix ain't that trivial. However I would recommend to move the aggregation logic to a class. The SarosSessionManager code is already complex enough. Adding more synchronization objects do not make it easier to understand. Furthermore the ISarosSession interface needs a Javadoc update. If you call this method from the GUI thread you can freeze the whole GUI.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ProjectNegotiationObservable.java
```diff
@@ -116,4 +116,15 @@ public synchronized void remove(ProjectNegotiation negotiation) {
 
         return currentNegotiations;
     }
+
+    /**
+     * @return true if running negotiations exist
+     */
+    public synchronized boolean negotiationsInProgress() {
```
srossbach: I would give this a better name or drop it completely as calling isEmpty on the result returned by list() yields the same result.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ProjectNegotiationObservable.java
```diff
@@ -116,4 +116,15 @@ public synchronized void remove(ProjectNegotiation negotiation) {
 
         return currentNegotiations;
     }
+
+    /**
+     * @return true if running negotiations exist
```
srossbach: Yeah I know the method description will be the same as the return description. However, just add it.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -114,6 +115,16 @@
 
     private final ProjectNegotiationObservable currentProjectNegotiations;
 
+    /** to queue resources for the next project negotiation **/
```
srossbach: That is what the HashMap is supposed to be used for. The description is not really helpful.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -114,6 +115,16 @@
...
+    private final Map<IProject, List<IResource>> nextSharedResources = new HashMap<IProject, List<IResource>>();
+    /** to set the thread for handling the next project negotiation **/
```
srossbach: I had a hard time to understand this. Basically it is a flag / mutex ... better a semaphore

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -114,6 +115,16 @@
...
+    private volatile boolean nextSharedResourcesWorker = false;
+    /**
+     * to notify a thread waiting for completion of all running project
```
srossbach: See my comment before this review, move the code to a dedicated class. The comments aint that helpful, even if you look a the code where all these objects are used.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -135,6 +146,13 @@ public void negotiationTerminated(final SessionNegotiation negotiation) {
         @Override
         public void negotiationTerminated(final ProjectNegotiation negotiation) {
             currentProjectNegotiations.remove(negotiation);
+
+            if (!currentProjectNegotiations.negotiationsInProgress()) {
+                /** notify about the completion of all running negotiations **/
```
srossbach: Do not repeat the code in your comments.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -532,44 +550,104 @@ public void addResourcesToSession(
             return;
         }
 
-        List<IProject> projectsToShare = new ArrayList<IProject>();
+        /** add new shared resources to a combining map for the next negotiation **/
+        synchronized (this) {
+            for (Entry<IProject, List<IResource>> mapEntry : projectResourcesMapping
+                .entrySet()) {
+                final IProject project = mapEntry.getKey();
```
srossbach: if (!containsKey(project) map.put(project, new Set<>)

if (resourceList == null || resourceList.isEmpty) {map.put(project, fullSharedMarkerSet); return;}

Set set = map.get(project)

if set ==  fullSharedMarkerSet
  return;

set.addAll(resourcesList)

Should be easier to understand.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -532,44 +550,104 @@ public void addResourcesToSession(
...
+        /**
+         * leave the synchronized block to allow new resource adding and wait
+         * till all running negotiations are done
+         **/
+        synchronized (nextSharedResourcesWorkerNotify) {
+            while (currentProjectNegotiations.negotiationsInProgress()) {
+                try {
+                    nextSharedResourcesWorkerNotify.wait();
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
```
srossbach: return ?

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ProjectNegotiationObservable.java
stefaus: any name suggestion?
No!, calling isEmpty on the result of list() yields to a different result (you can / will get a list of maybe empty lists) and list() is creating a new ArrayList for every call.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
stefaus: It is a semaphore as a plain variable. Will add something to the doc.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
stefaus: I would prefer to keep it at this place for now and change it with the refactoring of the package.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
stefaus: Please take in account that a new empty Set creates a partial project without resources.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
stefaus: would break the logic, only case escaping this would be a ending session or not?

srossbach: final Object l = new Object();
		int x = 0;
		
		Thread.currentThread().interrupt();
		
		synchronized(l) {
			while(x < 1000) {
				try {
					l.wait();
				} catch (InterruptedException e) {
					System.err.println("interrupted");
					Thread.currentThread().interrupt();
					x++;
				}
			}
		}
translate into busy spinning

If you want to ignore the interruption then catch the exception, remember that you got interrupted and when your logic is done call Thread#interrupt();

However getting interrupted is a signal to stop the thread gracefully, and so it should terminate as early as it can.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ProjectNegotiationObservable.java
tobous: The official oracle doc guide states this from the reversed situation, but the outcome is the same:
> include it for all other methods, even if its content is entirely redundant with the method description

https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@return

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/ProjectNegotiationObservable.java
stefaus: Shame on me, Sorry, I mixed it up with calling `negotiations.values().isEmpty()`.

Ok, than I change it to `isEmpty()` and replace other occurrences.

srossbach:
@tobous I added you to the review list as changes in the core can have in impact on some IntelliJ stuff as well. So it is not necessary to review the code but the impact on the changes it might produce.

But I guess there are not any ?

tobous:
@srossbach I don't know. I haven't come around to dealing with/looking at that stuff in the IntelliJ part yet.

But for now, I don't think this should have any relevance for the IntelliJ part as currently participants are only able to share a single module. There is no way of sharing multiple modules or to add resources to an existing session. Furthermore, sessions can only be created with one participant (+ host) and there is also no way of adding participants. All of this is restricted by the UI, but I don't know if there are any other obstacles (or if this is all done in the core and only the UI is missing).

I tested the commit locally and the project negotiation still seems to work as before. Do you know any any general part of the logic in the IntelliJ part (or specific classes; except the ui stuff) that this could affect?

srossbach:
Stefans current solution has added blocking behavior. If you do not call the affected method from the Application thread you have not to worry about anything.

As for adding users to the session and also adding more resources see: https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/core/ui/util/CollaborationUtils.java

Looking at the code you are already on the safe side.

If I am not mistaken adding a user should work without any issues. However I always suggest to start the "induction" proof with n=2 users and the step from n=2 to n=3 . We have already seen some solutions that work fine in an initial 2 user setup but then fail if another user joins the session.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -532,66 +588,43 @@ public void addResourcesToSession(
...
-            } finally {
-                startStopSessionLock.unlock();
+        try {
+            for (User user : currentSession.getRemoteUsers()) {
+
+                TransferType type = TransferType.valueOf(currentSession
+                    .getUserProperties(user).getString(
+                        ProjectNegotiationTypeHook.KEY_TYPE));
+                AbstractOutgoingProjectNegotiation negotiation = negotiationFactory
+                    .newOutgoingProjectNegotiation(user.getJID(), type,
```
saros-infrastructure: Code Smell Code Smell: Remove this use of "getJID"; it is deprecated. ([squid:CallToDeprecatedMethod](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3ACallToDeprecatedMethod))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=272&issues=AWbfy2AyLpqqSUlcop4G&open=AWbfy2AyLpqqSUlcop4G)

[ISSUE_KEY]: AWbfy2AyLpqqSUlcop4G

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ProjectNegotiationCollector.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private Thread worker = null;
+
+    /**
+     * Add project resource mappings for the next project negotiation.
+     * 
+     * @param projectResourcesMapping
+     *            project resource mappings to add
+     */
+    public synchronized void add(
```
saros-infrastructure: Code Smell Code Smell: Refactor this method to reduce its Cognitive Complexity from 16 to the 15 allowed. ([squid:S3776](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3AS3776))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=272&issues=AWbfy2BCLpqqSUlcop4H&open=AWbfy2BCLpqqSUlcop4H)

[ISSUE_KEY]: AWbfy2BCLpqqSUlcop4H

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -744,7 +777,7 @@ private boolean terminateNegotiations() {
 
         while (System.currentTimeMillis() - startTime < NEGOTIATION_TIMEOUT) {
```
saros-infrastructure: Code Smell Code Smell: Reduce the total number of break and continue statements in this loop to use at most one. ([squid:S135](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3AS135))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=272&issues=AWbfy2AyLpqqSUlcop4E&open=AWbfy2AyLpqqSUlcop4E)

[ISSUE_KEY]: AWbfy2AyLpqqSUlcop4E

stefaus:
Sorry for the delay.

Now I moved logic to a new class and removed the blocking behaviour (with a new thread if needed).

@tobous My goal is to replan the whole invitation process and make it more user friendly, like uniting separate invitation dialogs and delays between them.
In addition refactor this core part and hopefully make IntelliJ and the new GUI Implementaion easier. No fear I will talk to you about it, make a plan and present it in a ticket before work will start.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
saros-infrastructure: Code Smell Code Smell: Remove this use of "getJID"; it is deprecated. ([squid:CallToDeprecatedMethod](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3ACallToDeprecatedMethod))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=272&issues=AWbfy2AyLpqqSUlcop4G&open=AWbfy2AyLpqqSUlcop4G)

[ISSUE_KEY]: AWbfy2AyLpqqSUlcop4G

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
saros-infrastructure: Code Smell Code Smell: Reduce the total number of break and continue statements in this loop to use at most one. ([squid:S135](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3AS135))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=272&issues=AWbfy2AyLpqqSUlcop4E&open=AWbfy2AyLpqqSUlcop4E)

[ISSUE_KEY]: AWbfy2AyLpqqSUlcop4E

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ProjectNegotiationCollector.java
saros-infrastructure: Code Smell Code Smell: Refactor this method to reduce its Cognitive Complexity from 16 to the 15 allowed. ([squid:S3776](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3AS3776))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=272&issues=AWbfy2BCLpqqSUlcop4H&open=AWbfy2BCLpqqSUlcop4H)

[ISSUE_KEY]: AWbfy2BCLpqqSUlcop4H

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ProjectNegotiationCollector.java
```diff
@@ -0,0 +1,123 @@
...
+        Map<IProject, List<IResource>> projectResourcesMapping) {
+        for (Entry<IProject, List<IResource>> mapEntry : projectResourcesMapping
```
saros-infrastructure: Code Smell Code Smell: Reduce the total number of break and continue statements in this loop to use at most one. ([squid:S135](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3AS135))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=272&issues=AWb1y62KW3Y69bz5Xor5&open=AWb1y62KW3Y69bz5Xor5)

[ISSUE_KEY]: AWb1y62KW3Y69bz5Xor5

stefaus:
@srossbach Could you please review again, I hope all points are addressed.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ProjectNegotiationCollector.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /**
+     * Adds all added mappings to the current session, clears the map for next
+     * collection and returns a list of projects that should be handled by a
+     * project negotiation.
+     * 
+     * @return the list of projects to share
+     */
+    public synchronized List<IProject> getProjectsToShare(
```
srossbach: The name of this method does not reflect what it really does. Accessing the session should be done in the SessionManager and not here. The collector here should mainly merge resources that are required when for the next session negotiation.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -561,37 +616,34 @@ public void addResourcesToSession(
...
-            } finally {
-                startStopSessionLock.unlock();
+        try {
+            for (User user : currentSession.getRemoteUsers()) {
+
+                TransferType type = TransferType.valueOf(currentSession
+                    .getUserProperties(user).getString(
+                        ProjectNegotiationTypeHook.KEY_TYPE));
+                AbstractOutgoingProjectNegotiation negotiation = negotiationFactory
+                    .newOutgoingProjectNegotiation(user.getJID(), type,
```
saros-infrastructure: Code Smell Code Smell: Remove this use of "getJID"; it is deprecated. ([squid:CallToDeprecatedMethod](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3ACallToDeprecatedMethod))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=272&issues=AWb814lgTJjFgJLT72lr&open=AWb814lgTJjFgJLT72lr)

[ISSUE_KEY]: AWb814lgTJjFgJLT72lr

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -744,7 +796,7 @@ private boolean terminateNegotiations() {
 
         while (System.currentTimeMillis() - startTime < NEGOTIATION_TIMEOUT) {
```
saros-infrastructure: Code Smell Code Smell: Reduce the total number of break and continue statements in this loop to use at most one. ([squid:S135](https://sonarcloud.io/organizations/saros/rules#rule_key=squid%3AS135))

[See it in SonarCloud](https://sonarcloud.io/project/issues?id=saros_complete&pullRequest=272&issues=AWb814lgTJjFgJLT72lp&open=AWb814lgTJjFgJLT72lp)

[ISSUE_KEY]: AWb814lgTJjFgJLT72lp

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ProjectNegotiationCollector.java
stefaus: I moved the code back in the manager.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -532,66 +588,43 @@ public void addResourcesToSession(
...
         }
 
         List<AbstractOutgoingProjectNegotiation> negotiations = new ArrayList<AbstractOutgoingProjectNegotiation>();
 
-        synchronized (this) {
-            if (!startStopSessionLock.tryLock()) {
-                log.warn("could not start a project negotiation because the current session is about to stop");
-                return;
-            }
+        if (!startStopSessionLock.tryLock()) {
```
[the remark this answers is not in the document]
srossbach: Ok now I remember my logic. I used synchronized as second lock to avoid a possible situation where adding project and receiving a project concurrently would result in one of the operation being discarded. However this is currently not possible due to the architecture (only the host can add projects and only client can receive projects). We still have a patch from Denis that try to reintroduce this behavior for the Saros server. So I do not think moving the sychronized block is the best option here as you increase the size of the code that is being locked.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -135,6 +139,12 @@ public void negotiationTerminated(final SessionNegotiation negotiation) {
         @Override
         public void negotiationTerminated(final ProjectNegotiation negotiation) {
             currentProjectNegotiations.remove(negotiation);
+
+            if (currentProjectNegotiations.isEmpty()) {
+                synchronized (nextProjectNegotiation) {
```
srossbach: Why isn't this behavior moved to the class instead, e.g creating a signal method. The code is getting too complicated. If I count correctly there a now 3 locks involved for the logic. I think we should prefer an asynchronous solution instead of a concurrent one, e.g:

collector.add(resources)
excecutor.submit(runSharing)

the runSharing runnable can then look if there is a running negotiation, if yes abort. if there are no resources, abort

This is somewhat easier to handle, e.g you can execute the code 1 time or 1000 times it will not matter, the result be the same.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ProjectNegotiationCollector.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * This class collects resources that should be added to a session to add them
+ * at once in preparation of a new single project negotiation. The reason is to
+ * prevent multiple concurrent running project negotiations per user.
+ * 
+ * This class allows to hold the reference to a thread that should handle this
+ * new project negotiation.
+ */
+public class ProjectNegotiationCollector {
```
Drakulix: Why is this class holding the worker thread?

It collects mappings just fine, the worker thread stuff seems tagged on. Both parts could be separated easily and I do not think the "Collector" is responsible for the worker thread, but rather the SessionManager, which even creates the worker.

I get that this needs to be synchronized with the other methods, but responsibilities are a bit unclear and it would be nice, if this could be cleaned up a bit, as this stuff belongs into the SessionManager.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -533,20 +589,19 @@ public void addResourcesToSession(
         }
 
         List<IProject> projectsToShare = new ArrayList<IProject>();
-
-        for (Entry<IProject, List<IResource>> mapEntry : projectResourcesMapping
-            .entrySet()) {
-
+        for (Entry<IProject, List<IResource>> mapEntry : nextProjectNegotiation
```
Drakulix: A little comment on the locking logic would not hurt. I have to check the get-function to figure out that this loop is locking. Same applies for the isEmpty call. You could make it a bit more obvious what is going on, so it is easier to check, if any of this could trigger a deadlock.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ProjectNegotiationCollector.java
stefaus: My thought was to group logic to enable calls like `nextProjectNegotiation.isWorkerAlive()`, but rethinking it now... you're right and I moved it.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
stefaus: I moved it to a separate variable, hope that helps.

In general I'm not sure what you think I should comment, for example exisiting calls like `currentProjectNegotiations.add(negotiation);` are blocking as well without further comment.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -485,8 +497,56 @@ public void invite(Collection<JID> jidsToInvite, String description) {
...
+          }
+        };
+    nextProjectNegotiationWorker = ThreadUtils.runSafeAsync(log, worker);
+  }
+
+  /**
+   * This method handles new project negotiations for already invited user (not the first in the
+   * process of inviting to the session).
+   */
+  private synchronized void startNextProjectNegotiation() {
```
srossbach: Please pull the synchronized down to line 603 so it matches the former synchronized block. It should be clear what section must be synchronized. Synchronizing the whole method is not very helpful for understanding.

PR #278, opened 2018-10-23
[FIX][I] Correctly reset connection when unsubscribing editor listener

tobous:
StoppableEditorFileListener.unsubscribe() now correctly drops the held
MessageBusConnection object. If this object is not removed, subsequent
calls  to subscribe(Project) will fail as they think the listener is
already registered.


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/StoppableEditorFileListener.java
```diff
@@ -134,6 +134,8 @@ void subscribe(
      */
     void unsubscribe(){
         messageBusConnection.disconnect();
+
+        messageBusConnection = null;
```
srossbach: :bomb: Amazing how many (senseless) emoticons Github has to offer.

PR #280, opened 2018-10-23
[JUNIT][E] Added test classes to TestSuite

pdler:


m273d15:
Why did you rename the `TestSuite.java` to `EclipseTestSuite.java` ?
If I search for test suites I only find suites named `TestSuite.java` except the parent suite.

m273d15:
@pdler Is this PR still necessary? Otherwise I would close it

m273d15:
Please re-open if required.

PR #282, opened 2018-10-24
[FEATURE][SERVER] Add InviteCommand

Drakulix:
Adds an interactive commands to the server implementation to invite users by JID to the running session.


de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/console/InviteCommand.java
```diff
@@ -0,0 +1,39 @@
...
+        return "invite";
+    }
+
+    @Override public String help() {
+        return "invite <JID>... - Invite users to session";
+    }
+
+    @Override public void execute(String command, PrintStream out) {
+        try {
+            List<JID> jids = Arrays.asList(command.split(" ")).stream().skip(1).map(x -> new JID(x)).collect(Collectors.toList());
```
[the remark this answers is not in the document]
Drakulix: It is just these two commands, nothing more planned right now.
But I will open an issue to keep track of this.

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/console/InviteCommand.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  @Override
+  public void execute(String command, PrintStream out) {
+    try {
+      List<JID> jids =
+          Arrays.asList(command.split(" "))
+              .stream()
+              .skip(1)
+              .map(x -> new JID(x))
```
tobous: Check if JID is valid with `XMPPUtils.validateJID(JID)`?
You could either
- just filter out illegal JIDs before the collect or
- check after the collect and then abort the invitation process and print an error message if invalid JIDs are found.

I am not sure which option would be preferable.

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/ServerLifecycle.java
```diff
@@ -20,6 +22,7 @@
 public class ServerLifecycle extends AbstractContextLifecycle {
 
   private static final Logger log = Logger.getLogger(ServerLifecycle.class);
+  public ServerConsole console = null;
```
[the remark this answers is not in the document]
Drakulix: Forgot to clean that up after initial testing. I have added a setter.

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/console/InviteCommand.java
Drakulix: I decided to filter them out. This console is mostly for testing anyway and the logs state which users got invited.

donut87: But the information on who was invited is pretty dull. The important information is: who was not invited. I know it is intended for 'testing purposes', but could you please add some output information about who was not invited, because of invalid JID?

Drakulix: Added

Drakulix:
@stefaus Can you re-approve?
@tobous anything else?

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/SarosServer.java
```diff
@@ -22,19 +22,16 @@ public SarosServer() {
...
   }
 
-  public void start() {
+  public void start(ServerConsole console) {
 
     // Logging
     URL log4jProperties = SarosServer.class.getResource(LOGGING_CONFIG_FILE);
     PropertyConfigurator.configure(log4jProperties);
 
+    lifecycle.setConsole(console);
```
srossbach: I guess the context has a console but not the lifecycle.

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/ServerLifecycle.java
```diff
@@ -74,4 +78,16 @@ private void connectToXMPPServer(final ContainerContext context) {
...
+  public void setConsole(final ServerConsole console) {
+    this.console = console;
+  }
+
+  public void initConsole(final ContainerContext context) {
+    if (console == null) {
+      return;
+    }
+
+    console.registerCommand(new InviteCommand(context.getComponent(ISarosSessionManager.class)));
```
srossbach: Why dont you add the console to the context/container ? This is why we use DI/IOC to prevent those boilerplate code.

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/console/InviteCommand.java
```diff
@@ -0,0 +1,55 @@
...
+              .filter(
+                  x -> {
+                    if (!XMPPUtils.validateJID(x)) {
+                      log.warn("Invalid JID skipped: " + x);
```
donut87: I am about to start sounding pedantic.
Please do not mix concepts. A filter method should only do that: filter.
So, either filter and print afterward or use a forEach method. Do not introduce the concept of side effects to lambdas.
My personal preference: filter and then print
`filter(x -> !XMPPUtils.validateJID(x)).forEach(x -> log.warn("Invalid JID skipped: " + x));`

donut87: Addition: Yes, that will result in two lists. No, this is not a problem, because we do not have more than 10 JIDs anywhere so there is no performance problem to be seen anywhere.

srossbach: It is not pendantic. The signature is

Stream<T> filter(Predicate<? super T> predicate)

so this implies that there should be no side effects.

tobous: I don't like the proposed solution as it would mean that we would have to iterate over the JIDs twice.

In general, I wouldn't really think of logging as a side effect. Nonetheless, if this were to be changed, I would definitely prefer moving the check and the logging into a forEach call replacing `filter(...).collect(...);` to avoid additional iterations. This would also mean that we could no longer use `Stream.collect(...)` but would rather have to create an empty collection and add all elements separately.

donut87: > we would have to iterate over the JIDs twice.
See my comment above. This is not as bad as introducing side effects.

> In general, I wouldn't really think of logging as a side effect.
Oh, but it is.

Drakulix: > I am about to start sounding pedantic.
All good, you are right - in principle at least - and I am willing to fix this, no need to excuse.

> I don't like the proposed solution as it would mean that we would have to iterate over the JIDs twice.
But I feel exactly the same as @tobous about this point.

Sure, it would be a micro-optimization, nothing really necessary, but iterating twice is something that bothered me enough to seriously consider switch back to for-loops instead.

I get why side-effects should be avoided and that I am mixing imperative and functional programming styles, but this whole stream is small and very easy to understand.
So yes, I feel this is a little pedantic.

I will check the stream api for a way to partition this jids into two lists (sure there is something), which would at least eliminate the need to call `validateJID` twice for every element.

But this feels a lot like bike-shedding over around 10 lines adding a little debug command.

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/ServerLifecycle.java
Drakulix: I will take a look at it. I just quickly hacked something together after your suggestion of a `ContextLifecycle` broke my initial PR.

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/SarosServer.java
srossbach: I first suggest you change the CTOR to InputStream and OutputStream, you can wrap the outputstream inside the ServerConsole.

Next you just do this in the contextfactory: component.add(new ServerConsole(System.in, System.out)


The question is on how you would to register commands. You could also throw each command into the context:
public Mycommand(ServerConsole console) {console.registerCommand(this);}

or you add them through the lifecycle. The choice is up to you.

After looking at the code I just wonder why you have an identifier() method along with a matches() method. And I highly suggest that you throw away the command and only passes the arguments into the command (either already splitted or as string).

srossbach: The console can then be obtained through: server.lifeCycle().getContext().getComponent(ServerConsole.class)

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/console/InviteCommand.java
donut87: > But this feels a lot like bike-shedding over around 10 lines adding a little debug command.

It might feel that way. But to be honest, I would rather not have the logging of not invited JIDs, than this code. It violates a core programming principle and should be avoided.
There are a lot of discussions about using lambdas and I had enough of these discussions in the last 2 years with people used to using Iterators. If you need performance lambdas are the wrong tool. Heck... Java is the wrong tool!
What you gain from this "double loop": Well speaking code. In this project, the turnover rate is exceptionally high. Except for @srossbach everybody just works a short time on the code. So please be kind to the people who follow and write readable, explicit code with as few side effects as possible.

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/SarosServer.java
Drakulix: > After looking at the code I just wonder why you have an identifier() method along with a matches() method.

The idea was to allow classes to handle multiple related commands to avoid code duplication. Looking at the current implementations, that seems over-engineered.

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/console/InviteCommand.java
srossbach: I would also suggest to not optimize code that is mainly just run once. If the application behaves badly performance wise you can just attach a profiler (this is really easy with the JVM). From my experience hotspots normally happen where you do not expect them and the result is normally not from such missing "optimizations" as mentioned here, e.g do not iterate twice but most of the times it is a poor design decision or something else.

I personally seldom use lambas / stream API as they are not helpful and it seems they were tailored to solve enterprise map-reduce algorithms. What is even worse if you meet some guys who are a big fan of functional programming and so ****** up existing easy to read and understandable code by "lambdarizing" everything.

TLDR: If you have the feeling that a lambda expression does solve your problem very cumbersome then do not use it.

PR #283, opened 2018-10-24
[FEATURE][SERVER] Add ShareCommand

Drakulix:
Adds an interactive command to the server to share projects within the servers workspace with the session participants.


de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/console/ShareCommand.java
```diff
@@ -0,0 +1,60 @@
...
+      LOG.error("No Session running, cannot add any resources");
+      return;
+    }
+
+    try {
+      List<IProject> projects =
+          Arrays.asList(command.split(" "))
+              .stream()
+              .skip(1)
+              .map(path -> new ServerProjectImpl(this.workspace, path))
```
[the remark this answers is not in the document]
Drakulix: fixed :)

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/console/ShareCommand.java
```diff
@@ -0,0 +1,54 @@
+package de.fu_berlin.inf.dpp.server.console;
+
+import de.fu_berlin.inf.dpp.filesystem.IWorkspace;
+import de.fu_berlin.inf.dpp.server.filesystem.ServerProjectImpl;
+import de.fu_berlin.inf.dpp.session.ISarosSession;
+import de.fu_berlin.inf.dpp.session.ISarosSessionManager;
+import de.fu_berlin.inf.dpp.session.User;
+import java.io.PrintStream;
+import java.util.*;
```
tobous: It's weird that the formatter does not have a problem with this. According to the google style, wildcard imports are never used. Please remove the wildcard import.

I thought this was already reflected by the current formatting rules in Intellij and Eclipse. How did you format the code?

Drakulix: I have just used the standalone formatter, but that should fix that as well, right?

Drakulix: Fixed it manually.

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/console/ShareCommand.java
```diff
@@ -0,0 +1,54 @@
...
+  public void execute(List<String> args, PrintStream out) {
+    ISarosSession session = sessionManager.getSession();
+
+    if (session == null) {
+      log.error("No Session running, cannot add any resources");
+      return;
+    }
+
+    try {
+      args.stream()
```
srossbach: This is really a good use case to ask the question if we should use lambdas at all.

1. On line 47 you use still the normal iterator for:each, why no lambda ?

2. Is it really your intention to do

for (arg : args) pl.add(new XYZ(arg))
for (p : pl) someFunction(pl)

or

for (arg : args) someFunction(new XYZ(arg))

I am just asking because lets assume you want that an issue with an argument will not affect other arguments so you would normally write:

for (arg : args) try {someFunction(new XYZ(arg))}catch(...){}

but am I right that you have to write 2 catch blocks when using lambdas ?

so arg.stream().map(x -> try{...}catch(...){}).forEach(x -> try{...}catch(...){}))

Drakulix: Instead of making this issue about the usage of lambdas, you could have just pointed out, that I am missing separate error handling for every argument (similar to #282, where I was not validating JIDs), because yes, it is indeed desirable that a malformed argument is either blocking the whole command or just gets skipped instead of aborting the rest.

And indeed error handling with lambdas is quite difficult (partly due to java's exception system), because these iterators are lazy, so all previous arguments will have been processed before the exception is thrown and aborts the whole stream. Also a try-catch-block inside a lambda would not help you at all, because your lambda would still need to return a `ServerProjectImpl` even if you catch the exception.

So yes, lambdas are not a good solution to this problem, but they would have been just fine, if that particular mistake would not have been the case. Just like they did work out in the final (and merged) variant of #282.

So instead of arguing over code, that is merely a matter of personal preference, I would kindly ask you to point out the actual mistake of mine next time.

srossbach: I am sorry but that was really a discussion about lambdas and not about error handling in non production code.

try { doStuff() } catch(Exception weDoNotCare){} just continue is a common pattern for test code.

As this is debug stuff I let the author decide how much error handling is done.

Drakulix:
@tobous @srossbach Anything else? Or is this good to merge?

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/console/ShareCommand.java
```diff
@@ -0,0 +1,60 @@
...
+      for (String path : args) {
+        try {
+          IProject project = new ServerProjectImpl(this.workspace, path);
+          session.addSharedResources(project, project.getName(), null);
+        } catch (Exception e) {
+          log.error(path + " could not be added to the session", e);
+        }
+      }
+      for (User user : session.getRemoteUsers()) {
+        sessionManager.startSharingProjects(user.getJID());
```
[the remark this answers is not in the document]
Drakulix: Fixed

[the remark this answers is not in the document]
srossbach: Garbage API

/**
   * Adds project resources to an existing session.
   *
   * @param projectResourcesMapping
   */
public void addResourcesToSession(Map<IProject, List<IResource>> projectResourcesMapping);

However when you look at https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java#L500

You will notice something.

Drakulix:
> Garbage API

And I was wondering why I needed those two calls. Anyway should be fixed now.

PR #286, opened 2018-10-25
[BUILD] Use JDK8 and eclipse 4.6

m273d15:
* Changed required jdk version to 8
* Changed SWTBot to version 2.7.0

PR #271, opened 2018-10-19
[BUILD] Build saros via Gradle

m273d15:
Switch build tool to gradle
Use the same jdks as before:
* core: jdk6
* ui: jdk6
* eclipse: jdk6
* server: jdk7
* intellij: jdk8

PR #287, opened 2018-10-29
[BUILD] Remove configs and libs obsoleted by gradle

m273d15:
Remove:
* old build.xml and build.properties used by ant
* libraries in the lib directory
* old eclipse configurations (.project, .classpath, .settings)


tobous:
I think you need to rebase this branch on the latest version of #286 as it also contains the commit "[BUILD] Use JDK8 and eclipse 4.6."

tobous:
Please remove the .project files and add them to the .gitignore. Also add the .idea directory to the .gitignore.

srossbach:
As your commit message only states what you are doing, can you please:

Give a proper introduction on how:

1. Format the code (I think you deleted all configured options)
2. Being able to configure the plugin inside the IDE (as you dropped the PDE nature from the project)

m273d15:
> Format the code (I think you deleted all configured options)

IntelliJ: I will move the corresponding configurations to the new location.
Eclipse: The behaviour described below will keep the existing settings.

> Being able to configure the plugin inside the IDE (as you dropped the PDE nature from the project)

I had to change the workflow in eclipse in order to support the PluginNature entry. I don't know why but if I import the project via buildship (gradle support for eclipse) the dependency containers are ignored by PDE in the core and the ui project. I tried several approaches, but nothing worked. Therefore I had to switch to a handling that mainly uses the eclipse plugin of gradle (eclipse support for gradle). This plugin creates the required eclipse configurations and modifies the existing setting files if necessary.

Therefore the whole import process will change to:
* execute the gradle task `prepareEclipse`
* import the project as git project

-------------------------------------
I will add the corresponding changes to the corresponding PRs

srossbach:
What should always work is that after building the required projects is that the launch configurations to actually start the plugin to test stuff should still work. Can you verify that this is still the case ?

m273d15:
I changed the gradle eclipse behaviour in the first PR #271  and adjusted the IDE specific configs in the PR

m273d15:
[Here](https://travis-ci.com/saros-project/saros/jobs/156339084) is an example stf run. I forgot to trigger the stf tests in .travis.yml therefore I will adjust the test script path and the test script (as [here](https://github.com/saros-project/saros/commit/ff17ef8fff8a66b29b6665e728fafdb4343c23dd)). Even if this run is green, there are a few sporadic failures (I think some test of the consistency package are failing sometime).

m273d15:
All PRs are adjusted an are able to execute the STF tests. See [here](https://travis-ci.com/saros-project/saros/jobs/156562489) for the stf test run with the latest PR version. This run contains one failing test I mentioned before.

de.fu_berlin.inf.dpp.intellij/.idea/scopes/scope_settings.xml
```diff
@@ -1,5 +0,0 @@
-<component name="DependencyValidationManager">
```
tobous: Is there a reason to remove this file?

de.fu_berlin.inf.dpp.intellij/.idea/uiDesigner.xml
```diff
@@ -1,125 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
tobous: Is there a reason to remove this file?

de.fu_berlin.inf.dpp.intellij/.idea/scopes/scope_settings.xml
m273d15: This config is only used for creating scopes which are used as alternative view of the project sources. The current config is useless

de.fu_berlin.inf.dpp.intellij/.idea/uiDesigner.xml
m273d15: Nobody uses the UI Designer

de.fu_berlin.inf.dpp.intellij/.idea/.name
```diff
@@ -1 +0,0 @@
-Saros-I
```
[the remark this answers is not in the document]
m273d15: Is removed by gradle/intellij

m273d15:
@tobous @srossbach 
Do you still have change requests for the gradle stuff?

m273d15:
Adjusted the gitignore file and removed the newlines. The run `runIde` issue depends on the intellij plugins.

m273d15:
> For whatever reason, I still get the newline at the end of .idea/compiler.xml that is then automatically removed. Please check if this is also the case when you check out this branch.

Sorry, I removed the newline via vim (vim adds a newline per default at the end of the file), but I forgot to open the file with `:set binary` and `:set noeol` (in order to remove the "invisible" default newline).

> Also, if I call delivery > sarosEclipse, all */.settings/org.eclipse.jdt.core.prefs files change. These changes are picked up by git. These files are also changed by ide > eclipse and are removed by ide > cleanEclipse, which tells me that they probably should not be included in the repo.

If I call `./gradlew sarosEclipse` (I also called the task via IntelliJ) no file changes (on tracked files) are detected by git. Don't call the task `ide > eclipse`. The `eclipse` task is provided by the eclipse plugin. If you use the default handling this job is called for creating initial eclipse configurations. I adjusted this job and we use the task `prepareEclipse`, because the job `eclipse` overwrites the styling configurations etc. per default. As far as I know it is not possible to exclude task descriptions without removing the plugin (which is required).

> In general, why are the intellij tasks in "intellij", but the eclipse tasks are in "ide"?

The tasks are provided by different gradle plugins which use different names for their task groups.

> Both the tasks verification > test and verification > check only run if you have changed something since the last test run. This confused me for a bit and could probably be made more clear.

I think you are confused by the [incremental build](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks) behaviour of gradle.

tobous:
Looks good for the IntelliJ setup.

PR #297, opened 2018-11-04
[INTERNAL][HTML WB] Added constants

AbdBarho:
these constants are divided into 4 files:

- SVG: contains SVG related constants that will be used when setting or accessing properties from and to shapes
- SXE: constants regarding the synchronisation protocol SXE which is used internally by the whiteboard
- fabric config: configuration of the FabricJs library used to render the different shapes in the whitebaord
- logic: mainly used in the menu to attach each button to  a specific action


tobous:
Please stick to the tag rules defined in the [documentation](http://www.saros-project.org/coderules/before-committing) (capitalize the internal tag).

In general, please try to add a description to the commits explaining what the commit does and why it does it.

de.fu_berlin.inf.dpp.whiteboard/frontend/src/constants/fabricconfig.js
```diff
@@ -0,0 +1,39 @@
+// default config for fabricjs objects
+const FABRIC_DEFAULTS = {
+  //applies to all objects
+  GENERAL: {
+    //calculates dimensions from top left cornder
```
srossbach: corner ?

de.fu_berlin.inf.dpp.whiteboard/frontend/src/constants/SVG.js
```diff
@@ -0,0 +1,41 @@
...
+    CY: "cy",
+    WIDTH: "width",
+    HEIGHT: "height",
+    TEXT: "text",
+    COLOR: "color",
+    FILL: "fill",
+    FONT_SIZE: "fontSize"
+  },
+  isNumerical(name) {
+    return name in {
```
srossbach: Why don't you use a Set ?

AbdBarho: my idea was that dictionaries are faster for look up, but what I forget is that the dictionary is actually re-created every time the function is called. it was a micro optimisation nothing more.

m273d15:
@AbdBarho you should add more reviewers to your pull requests.

PR #300, opened 2018-11-04
[INTERNAL][HTML WB] Added a menu containing the available tools

AbdBarho:
the menu will appear to the left of the whiteboard, with ability to be resized to mini mode, making it more compact and allows more drawing space.

the menu is rendered dynamically using the data structure in the `menus.js` file 

the icons that appear on the buttons of the menu are also committed


de.fu_berlin.inf.dpp.whiteboard/frontend/src/components/menus.js
```diff
@@ -0,0 +1,100 @@
+import LOGIC from '../constants/logic.js';
+
+import drawRectImg from '../../assets/images/draw_rect.png';
```
srossbach: You should make a bundle and install that. You will be surprised that nothing works.

I guess that the contents will be bundled a litte bit different so I guess the src folder will not be present in the final bundle artifact.

AbdBarho: I did not fully understand what you meant with making a bundle first and then installing it, could please explain a little bit?

the way that it is currently done is that these imported images will be converted to [Base 64 URI](https://en.wikipedia.org/wiki/Data_URI_scheme#HTMLl) so that the actual image is embedded in the code and does not need to be fetched from anywhere.

[Live example](https://jsbin.com/yasojocuca/edit?js,output)

AbdBarho: yes I delayed uploading the build scripts because I have no idea how to setup and work with travis.
uploading the code first was of higher priority.

srossbach: @m273d15 can you please assist @AbdBarho in setting up the required stuff. In my opinion a running CI environment is more important than pushing code into the repo without verification.

m273d15: @AbdBarho Sorry, I missed this comment. What do you want to integrate into Travis?

de.fu_berlin.inf.dpp.whiteboard/frontend/src/components/menu.jsx
```diff
@@ -0,0 +1,94 @@
...
+   */
+  constructor(props) {
+    super(props);
+    this.toolManager = props.toolManager;
+    this.canvasResizer = props.canvasResizer;
+    this.state = {};
+    this.miniMode = false;
+    this.wrapper = document.getElementById("menuContainer");
+    //start in miniMode because the windows in eclipse are small
+    //this.switchMiniMode();
```
m273d15: Why is this uncommented?

de.fu_berlin.inf.dpp.whiteboard/frontend/src/components/menu.jsx
```diff
@@ -0,0 +1,94 @@
...
+                <img draggable="false" src={tool.imageSrc} className="dropDownMenuItemImage" alt={tool.name}/>
+              </div>)}
+          </div>
+        </div>)}
+    </div>;
+  }
+
+  switchMiniMode() {
+    this.miniMode = !this.miniMode;
+    if (this.miniMode)
```
m273d15: why not `this.setMiniMode(!this.miniMode)`

de.fu_berlin.inf.dpp.whiteboard/frontend/src/components/menu.jsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+  /**
+   * sets mini tool for the menu,
+   * is used  when right mouse button is clicked to force mini mode, but not permenantly
+   * @param {Boolean} state
+   */
+  setMiniMode(state) {
+    if (this.miniMode)
+      return;
+    if (state)
```
m273d15: could be ternay operator

de.fu_berlin.inf.dpp.whiteboard/frontend/src/components/menu.jsx
```diff
@@ -0,0 +1,94 @@
...
+      this.wrapper.className = "miniMode";
+    else
+      this.wrapper.className = "";
+  }
+  /**
+   * sets mini tool for the menu,
+   * is used  when right mouse button is clicked to force mini mode, but not permenantly
+   * @param {Boolean} state
+   */
+  setMiniMode(state) {
```
m273d15: setMiniMode does not change the this.miniMode variable is this correct?

de.fu_berlin.inf.dpp.whiteboard/frontend/src/components/menu.jsx
AbdBarho: It would not be clear to the user that the menu actually has another normal mode, it would seem as if the menu is only restricted to this small size

de.fu_berlin.inf.dpp.whiteboard/frontend/src/components/menu.jsx
AbdBarho: updated

de.fu_berlin.inf.dpp.whiteboard/frontend/src/components/menu.jsx
AbdBarho: Yes this was very bad naming from my side, this function is created specifically for the context menu mode, where the user presses the right mouse button and the menu appears minimised next to the mouse cursor, I created another property on the menu and renamed the function to something more appropriate

Issue #177, opened 2018-07-30
Refactoring Saros Filesystem


pdler:
During the development of Saros for IntelliJ, it came to light, that some concepts of the Saros Core, like the Saros file system, does not match well with the concepts of IntelliJ. The problem is, that the design of Saros Core is project based. 

I’ve set the goal, that Saros Core gets a new filesystem and detach the project-based design. Which requirements has the new Core filesystem to fulfill? 

We have to look the resources like a resource tree (only files and folders like an operation system).
The following graphics shows an example, how Eclipse sees the resources.

```

 +-----+
 |Prj 1 (as project)|
 +-----+
    |
    |       +--------
    +-------+src 1 (as folder)|
    |       +--------
    |           |
    |           +------->file2.txt (as file)
    |
    |
    +---->file1.txt (as file)
....
```

The following graphics shows the same example, but how (refactored) Saros would see it.

```

  +-----+
  |Prj1 (as folder)|
  +-----+        
     |           
     |       +--------
     +-------+src 1 (as folder)|
     |       +--------
     |            |
     |            +------->file2.txt (as file)
     |
     |
     +---->file1.txt (as file)

....
```

An user can share any subtree (ex.: src 1) and src1 has an absolute path in the filesystem. The other user of the session can choose, where the src1 is located. So it can have an another absolute path. But what unites them is the reference point: `IReferencePoint`. Saros has to know (locally), where the reference point is, and can determine the resources with the combination of the reference point and the relative path to the reference point. An IReferencePoint itself is empty, but you can access the reference point from the `core.IFile`/`core.Folder` objects. 

So the filesystem of the IDEs can work well, a mapper, which maps reference point objects to IDE related filesystem objects, will be needed: `ReferencePointManager`. In Eclipse, the `ReferencePointManager` maps an `IReferencePoint` to `eclipse.IProject` objects. In IntelliJ the `ReferencePointManager` maps an `IReferencePoint` to `VirtualFiles`. 

The following branches represents the refactoring steps. Many of them are simple adjusting classes to reference points. The last branch pr/TrimmingFilesystem contains the current final state of the new saros file system. Saros/I and Saros/E implement the new resource interface. The implementation for Saros Server is still missing but I will do it next. 
[...]

PR #311, opened 2018-11-22
[CORE] Adjusting Negotiation to ReferencePoints

pdler:
The negotiation component are now based on reference points. 

The fileListFactory now takes a reference point and the referencePointManager instead of an project. The referencePointManager is for the incomming negotiation important for getting the resources from a reference point. 

Many of the commits are renaming of the classes. 

More details about the idea of the adjusting to reference point:  https://github.com/saros-project/saros/issues/177


Drakulix:
Without reading the actual code, do we really want to rename `ProjectNegotiation` to `ReferencePointNegotiation`?

I get, that we want to get rid of projects eventually and that this is a term introduced by eclipse, but the first implementation of the new file system will not be able to share any reference point right? So there is still a structure, higher in the hierarchy then the reference point, which cannot be split further for a negotiation.

And even if we are *technically* there, the new name is much more un-intuitive. Without knowledge of the filesystem implementation, I cannot really see, what this call is supposed to do.

If we want to get rid of "Project" maybe we want to rename this to something else, like `SharingProjectNegotiation`. Something that makes the use-case more obvious.

In my opinion the class name should be descriptive, where possible, while the use of `ReferencePoint`s really is an implementation detail.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/communication/extensions/ProjectNegotiationOfferingExtension.java
```diff
@@ -14,19 +14,19 @@
...
     public static final Provider PROVIDER = new Provider();
 
-    private List<ProjectNegotiationData> projectNegotiationData;
+    private List<ReferencePointNegotiationData> projectNegotiationData;
     private TransferType transferType;
 
     public ProjectNegotiationOfferingExtension(String sessionID,
         String negotiationID,
-        List<ProjectNegotiationData> projectNegotiationData,
+        List<ReferencePointNegotiationData> projectNegotiationData,
```
Drakulix: If the name is changed, you should change the variable names as well. (This refers to many more locations.)

Also you might want to think about abbreviations of "ReferencePoint" to RefPoint for brevity.

tobous:
Deprecated.

Issue #303, opened 2018-11-05
Introduce unified code formatting for all IDEs


tobous:
Currently, the code formatting rules used in Eclipse do not match the code formatting rules used in IntelliJ. This leads to an unevenly formatted codebase. Furthermore, this can lead to edit-wars/unnecessary formattings when working on code of an IDE specific part of Saros with a different IDE.

As far as I can tell, all project except "intellij" are formatted according to the Eclipse formatting rules.

stefaus:
I think we have two options for this.

1: Use [this Plugin](https://plugins.jetbrains.com/plugin/6546-eclipse-code-formatter) to import the existing Saros Eclipse Format Rules into IntelliJ.
I've tested it, but it has some problems with formating using the existing Saros config.
I think this occures because the Saros format rules File is to old and this would change with using 4.6+ Eclipse (and update the existing rules file).

But as I already expressed at the Stand-Ups I don't like the current formatting rules.

Therefore my preferred option 2: Use the google java format rules. This obsolete discussions about format details, brings a widely used ruleset and it comes with a stand alone formatter.
https://github.com/google/google-java-format

They provide Plugins to format code directly in Eclipse and IntelliJ.
It works from the CLI making it IDE independent and even provide gradle support, so we could check format violations for each check in.

m273d15:
Option 2 looks good! The rule set is immutable and can be checked during the CI process :+1:

tobous:
That sounds like an interesting solution, but we should discuss the contained formatting when considering using predefined *immutable* settings.

For example, I am not really happy with their default indentation width being 2 spaces (instead of 4).

stefaus:
Yes, the indentation was also unusual to me.
But the gain: Block indentation (+2 spaces) is easier to distinguish from continuation lines (+4 spaces).

Changing the format rules would need a project fork, making things difficult.

srossbach:
Why do we need a fork ? We could modify the source to our needs, compile it and put the result in one of our Git Repositories. That would at least ensure that every uses the same version.

srossbach:
I think without a default formatter there is to much room for edit wars (e.g one prefers short line, the next one does not even care etc.). And marking a build as failed because of some violation regarding the layout of the code is in my opinion a very bad idea.

tobous:
[...]

PR #313, opened 2018-11-23
[BUILD] Check java formatting in Travis CI

m273d15:
* Introduced Formatting stage in travis
* Created script for downloading and executing
  the google java formatter


m273d15:
See #303 for further information

travis/script/format/check_java.sh
```diff
@@ -0,0 +1,34 @@
...
+shopt -s globstar
+
+project_root_dir=$1
+formatter_jar_name=formatter.jar
+
+function download_formatter() {
+  local jar_name=$1
+  local url="https://repo.maven.apache.org/maven2/com/google/googlejavaformat/google-java-format/1.6/google-java-format-1.6-all-deps.jar"
+  echo "Downloading formatter from $url"
+  curl -Lo formatter.jar "$url"
```
stefaus: I guess you wanted to use jar_name here

travis/script/format/check_java.sh
```diff
@@ -0,0 +1,34 @@
...
+}
+
+function check_formatting() {
+  local jar_name=$1
+
+  echo 'Checking format'
+  java -jar "$jar_name" --dry-run --set-exit-if-changed **/*.java
```
stefaus: If just check changed files is enough you could maybe use the file list of `git diff master --name-only`.

tobous:
How about also removing the old fotmatting Config files as they are no longer useful and could give users the impression that they are using the correct formatting?

m273d15:
I removed IntelliJ code style settings and the eclipse formatting profile. I did not remove the eclipse clean-up profile, because it configures things like (remove unused imports etc.) and not formatting.

fzieris:
The files in the "patches" folder were always excluded from any auto-format commits. Those are "original" pieces of other projects' code with minor amendments. Auto-formatting these makes it way harder to see the actual "patch".

PR #325, opened 2018-11-27
[FEATURE][I] #261 Add logic to propagate saves

### LINKED ISSUE #261: Propagate saves to other participants
Saros/I currently does not propagate local saves to other participants.

Local saves can be reacted to by using `FileSystemChangedListener.beforeContentsChanged()` and testing `isFromSave()`.

tobous:
Adds the logic to propagate the local saving of a file to other
participants in the session if the saved file is shared.

Resolves #261.


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/project/FileSystemChangeListener.java
```diff
@@ -68,6 +69,8 @@
 
   @Inject private Project project;
 
+  @Inject private LocalEditorHandler localEditorHandler;
```
srossbach: I hope you are aware that you reversed the depencies for all stuff in here. it is really weird that this listener knows what classes it has to notify.

tobous: I don't really understand what you mean. Could you elaborate? Do you mean this class in general or just this patch?
This class is a listener for IntelliJ events and largely just dispatches activities and cleans up the local state (like held annotations, the editor mapping, etc.).

This patch just represents a special case as IntelliJ mixes logic that we separated. We strictly separate the editor logic from the file logic, but, as far as I know, this is the only way to detect and react to saves in Intellij.

Or do you mean that I should call the `EditorManager`, which internally calls the `LocalEditorHandler` (and subsequently move the logic to the `LocalEditorHandler`) instead of doing it the other way around?
For me, there is currently not a clear separation between what `EditorManager`, `LocalEditorHandler` and `LocalEditorManipulator` should handle (which is for example shown by the fact that the EditorManager has to expose the `EditorPool` for the two editor handlers). I was planning on creating a ticket for this in the near future (if you don't disagree and can explain the current separation and why it is correct :wink:).

srossbach: The name of this class is a bit misleading. Normally you would assume the components hook into here to be notified about file changes. However you do it the opposite way as this listener is aware of what components it has to notify,

tobous:

PR #328, opened 2018-11-29
[BUILD] Update gradle from 4.10.1 to 5.0

m273d15:
* Changes in gradle* files/directories are generate by the command:
  `./gradlew wrapper --gradle-version 5.0`
* build-scan plugin version has to be changed to 2.0.2


tobous:
Did you look at the mentioned changes in https://docs.gradle.org/5.0/userguide/upgrading_version_4.html#changes_5.0 to ensure that there are no other manual changes that are necessary for the upgrade of our setup?

m273d15:
The last time we discussed this topic tt was not clear to me whether we have to configure annotation processor. Now I know that we don't have to configure them in gradle, because annotation processor configurations are only required if you use annotations that imply a modification during compile time (modifying the AST(abstract syntax tree)).

Common annotations as `@NotNull` work without a gradle configuration (I tested the annotation with gradle 5). Therefore we don't have to make additional changes. See the [gradle forum](https://discuss.gradle.org/t/regarding-the-annotation-processors-on-compile-classpath-warning-in-gradle-4-6/26144/8) for an example of annotations that required the configuration. Btw. we don't use the (in the forum) described log4j annotation, because they are log4j2 related

tobous:
> Common annotations as `@NotNull` work without a gradle configuration (I tested the annotation with gradle 5).

@m273d15 This is not completely true. I would assume that you tested the IntelliJ annotations. They only work without an explicit gradle configuration as this is done by the gradle-intellij-plugin. The explicit handling of `@NotNull` and `@Nullable` annotations (or rather their replacement with assertions on compile time) can be set with the parameter `instrumentCode`.

But, if we don't use any other annotations that would warrant an annotation processor, your statement (that we don't have to worry about annotation processing) still holds true.

m273d15:
You are right.

PR #330, opened 2018-11-29
[E] Set correct import order in default project settings

tobous:
Sets the google java style import ordering as the default for all Saros
projects. This was all done through the UI, so any changes to the files
were done automatically by Eclipse.


tobous:
Furthermore, when calling the Eclipse option "Source" > "Organize imports" on the project, it still changes  a lot of imports. Some of them have the form
```Java
import src.de.fu_berlin.inf.dpp.ui.widgetGallery.application.Activator;
```
The leading "src" seems a bit weird to me. Also, it replaces a lot of static imports with normal imports.
Here is the complete diff: [orgamize_imports_results.txt](https://github.com/saros-project/saros/files/2631291/orgamize_imports_results.txt)

But these changes don't cause any issues with the formatter.

stefaus:
So this works for me, the thing I noticed is that save actions are disabled for the IntelliJ Project, so imports are only sorted after manual called clean up. So I would prefer activate them for this as well.

`de.fu_berlin.inf.dpp.ui.widgetGallery` looks completely broken, Libraries are missing, JDK 1.6 is selected. This should be the reason why the imports are wrong.
@m273d15 has it been forgotten in the transition? or abandon it?

m273d15:
The widgetGallery was just a demo project that was used to demonstrate the usage of SWT GUI elements. I don't even understand why this is part of the repository since it does not require other saros functionality (as far as I know). If the HTML GUI is available we can remove the project, but I think I will move this project into a separate repository and I will delete the repository if we use the HTML GUI.

srossbach:
The widget gallery isn't just a demo. It is a plugin that offers you to develop widgets in real-time (theoretically) by using the code hot-swap feature of the Eclipse OSGi Debug Classloader, i.e you modify code in your widget an hit refresh in the widget gallery to see the changes immediately. 

But I can give you an good advice. Just delete it from the repository.

m273d15:
Ah, ok. Thank you for that advice. I think we should do it.

srossbach:
> 
> 
> Furthermore, when calling the Eclipse option "Source" > "Organize imports" on the project, it still changes a lot of imports. Some of them have the form
> 
> ```java
> import src.de.fu_berlin.inf.dpp.ui.widgetGallery.application.Activator;
> ```
> 
> The leading "src" seems a bit weird to me. Also, it replaces a lot of static imports with normal imports.
> Here is the complete diff: [organize_imports_result.txt](https://github.com/saros-project/saros/files/2631674/organize_imports_result.txt)
> 
> But these changes don't cause any issues with the formatter.

The leading "src" is not only weird but broken, it should not even compile. Seems Eclipse does not recognize the src directory as source folder.

de.fu_berlin.inf.dpp/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -1,391 +1,390 @@
...
-org.eclipse.jdt.core.formatter.brace_position_for_annotation_type_declaration=end_of_line
-org.eclipse.jdt.core.formatter.insert_space_after_colon_in_conditional=insert
-org.eclipse.jdt.core.compiler.problem.localVariableHiding=ignore
-org.eclipse.jdt.core.formatter.alignment_for_assignment=0
-org.eclipse.jdt.core.formatter.insert_space_before_comma_in_allocation_expression=do not insert
-org.eclipse.jdt.core.compiler.problem.staticAccessReceiver=warning
-org.eclipse.jdt.core.formatter.insert_space_after_opening_paren_in_if=do not insert
-org.eclipse.jdt.core.formatter.insert_space_before_opening_brace_in_block=insert
-org.eclipse.jdt.core.formatter.insert_space_after_semicolon_in_try_resources=insert
+eclipse.preferences.version=1
```
AbdBarho: Most rules here did not change, but only got sorted alphabetically

de.fu_berlin.inf.dpp.whiteboard/.settings/org.eclipse.jdt.ui.prefs
```diff
@@ -55,6 +60,10 @@ eclipse.preferences.version=1
 editor_save_participant_org.eclipse.jdt.ui.postsavelistener.cleanup=true
 formatter_profile=_Saros Formatter
 formatter_settings_version=12
+org.eclipse.jdt.ui.ignorelowercasenames=true
+org.eclipse.jdt.ui.importorder=\#;;
```
AbdBarho: does any one have any idea what does `\#;` stand for?

tobous: I don't really get the syntax, but it should be the setting that sets the import order to match the google style:
```
<static import block>

<import block>
```

But it doesn't really matter as it was automatically generated by eclipse.

tobous:
I also added two commits that set the same base state for all eclipse projects. I did this by importing the clean-up-profile.xml into each project and setting the save action settings to the settings used in the Saros/E project (de.fu_berlin.inf.dpp). Any changes made to the config files were made automatically by Eclipse.

The save action settings should be adjusted later on as they currently do not include the removal of unnecessary imports, etc. But this should be done in a separate PR in which the optimal settings can be discussed further.

tobous:
I am not sure whether I should remove the documentation for importing the cleanup profile during the eclipse setup. It should no longer be necessary as the default settings are included in the project config, but they are only held as an "unmanaged profile". This means they can't be changed locally without creating a new profile. This can be seen as a positive point as it will prevent users from easily overwriting the repo settings. On the other hand, I don't know which other drawbacks unmanaged profiles have.

Does anyone else have an opinion on this?

PR #257, opened 2018-10-10
[CORE] Adjusted SharedProjectMapper to ReferencePoints

pdler:
In Saros Core the IReferencePointManager was introduced. As described on https://github.com/saros-project/saros/issues/177, The IReferencePointManager is needed in Saros Core during adapting step by step. 
The IReferencePointManager maps an IReferencePoint to an core.IProject and returns either an core.IProject given by an IReferencePoint or a set of core.IProject given by a set of IReferencePoint. The ReferencePointManager is an implementation of the IReferencePointManager.

SharedProjectMapper/-test:
The SharedProjectMapper now works with IReferencePoints. The logic itself did'nt changed. The unit tests had to be adjusted. The SharedProjectMapper was renamend to SharedReferencePointMapper in a seperate commit.

IReferencePointManager/ReferencePointManager:
The IReferencePointManager maps an IReferencePoint to an core.IProject.
Because not all classes will not adapted at the same time, the IReferencePointManager
is needed for classes, who still works core.IProjects. The ReferencePointManager
is an implementation of IReferencePointManager.

SarosSession:
After the SharedProjectManager works with reference points, the SarosSession needs
the ÍReferencePointManager for getting IProjects given by IReferencePoints.
The IReferencePointManager will instantiated in SarosSession and kept in
the core session container. The IReferencePointManager will needed only if a session is
running, so that the reason, that the IReferencePointManager is in the core session
container. Currently the SarosSession is responsible for putting the pairs IReferencePoints
and core.IProjects, which should to be shared.

PR #332, opened 2018-12-03
Pr/adjusting shared project mapper v2

pdler:
This PR is an updated version of the PR https://github.com/saros-project/saros/pull/257. The difference is that this PR is based on the new Saros enviroment...

In Saros Core the IReferencePointManager was introduced. As described on https://github.com/saros-project/saros/issues/177, The IReferencePointManager is needed in Saros Core during adapting step by step. 
The IReferencePointManager maps an IReferencePoint to an core.IProject and returns either an core.IProject given by an IReferencePoint or a set of core.IProject given by a set of IReferencePoint. The ReferencePointManager is an implementation of the IReferencePointManager.

SharedProjectMapper/-test:
The SharedProjectMapper now works with IReferencePoints. The logic itself did'nt changed. The unit tests had to be adjusted. The SharedProjectMapper was renamend to SharedReferencePointMapper in a seperate commit.

IReferencePointManager/ReferencePointManager:
The IReferencePointManager maps an IReferencePoint to an core.IProject.
Because not all classes will not adapted at the same time, the IReferencePointManager
is needed for classes, who still works core.IProjects. The ReferencePointManager
is an implementation of IReferencePointManager.

SarosSession:
After the SharedProjectManager works with reference points, the SarosSession needs
the ÍReferencePointManager for getting IProjects given by IReferencePoints.
The IReferencePointManager will instantiated in SarosSession and kept in
the core session container. The IReferencePointManager will needed only if a session is
running, so that the reason, that the IReferencePointManager is in the core session
container. Currently the SarosSession is responsible for putting the pairs IReferencePoints
and core.IProjects, which should to be shared.

EDIT: 
I reworked issue https://github.com/saros-project/saros/issues/177
and you can see the final state in PR https://github.com/saros-project/saros/pull/391

### DIFF
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/IReferencePointManager.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/IReferencePointManager.java
@@ -1,0 +1,33 @@
+package de.fu_berlin.inf.dpp.session;
+
+import de.fu_berlin.inf.dpp.filesystem.IProject;
+import de.fu_berlin.inf.dpp.filesystem.IReferencePoint;
+import java.util.Set;
+
+/** The IReferencePointManager maps an IReferencePoint to IProject */
+public interface IReferencePointManager {
+
+  /**
+   * Insert a pair of reference point and project
+   *
+   * @param referencePoint the key of the pair
+   * @param project the value of the pair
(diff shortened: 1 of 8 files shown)


m273d15:
It seems as if this PR is also included in #391. Which one should be reviewed? If #391 should be reviewed, pleases close this PR. Same for #331

PR #337, opened 2018-12-04
[FIX][I] #285 Introduce logic to handle content recovery

### LINKED ISSUE #285: Introduce better logic to handle content recoveries
The rework of the SharedResorucesManager implemented in #277 removes the logic to recover existing files with the wrong content.

To recover existing files with the wrong content, Saros also dispatches creation FileActivities. The old recovery logic just overwrote the content of the existing file in the filesystem.
This is not the best approach as it bypasses the IntelliJ logic to make changes undo-able, meaning the recovery would irreversibly change the local file content.

A better approach would be to change the content of the Document (the cached version of the file held by IntelliJ that is used to back the local Editors).

tobous:
Adds logic to handle content recovery, meaning the content of an
existing file needs to be adjusted. Previously, this was handled on the
filesystem level, meaning the changes made by the recovery were not
undo-able. This commit introduces a logic to handle content recoveries
on the document level to avoid such issues.

Resolves #285.

Subsequently enables the recovery button as the feature should now be
functional.

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorManipulator.java
+++ b/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorManipulator.java
@@ -120,30 +120,6 @@ public class LocalEditorManipulator {
   }
 
   /**
-   * Replaces the content of the document at the given path. The text is only replaced, if the
-   * editor is writable.
-   *
-   * @param path path of the editor
-   * @param text text to set the document's content to
-   * @return Returns <code>true</code> if replacement was successful, <code>false</code> if the path
-   *     was <code>null></code>, if the path points to a non-existing document or the document was
-   *     not writable.
-   */
-  public boolean replaceText(SPath path, String text) {
-    Document doc = editorPool.getDocument(path);
-    if (doc == null) {
-      return false;
-    }
-    if (!doc.isWritable()) {
-      LOG.error("File to replace text in is not writeable: " + path);
-      return false;
-    }
-
-    editorAPI.setText(doc, text);
-    return true;
-  }
-
-  /**
    * Applies the text operations on the path and marks them in color.
    *
    * @param path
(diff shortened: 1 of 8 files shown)


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorManipulator.java
```diff
@@ -233,4 +218,77 @@ public void adjustViewport(Editor editor, LineRange range, TextSelection selecti
...
+   * @param path the path of the file to recover
+   * @param content the new content of the file
+   * @param encoding the encoding of the content
+   * @param source the user that send the recovery action
+   * @see Document
+   * @see SharedResourcesManager#handleFileRecovery(FileActivity)
+   */
+  public void handleContentRecovery(SPath path, byte[] content, String encoding, User source) {
+    VirtualFile virtualFile = VirtualFileConverter.convertToVirtualFile(path);
+    if (virtualFile == null) {
```
srossbach: warn ?

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorManipulator.java
```diff
@@ -233,4 +218,77 @@ public void adjustViewport(Editor editor, LineRange range, TextSelection selecti
...
+    IFile file = path.getFile();
+    annotationManager.removeAnnotations(file);
+
+    String text;
+    try {
+      text = new String(content, encoding);
+    } catch (UnsupportedEncodingException e) {
+      LOG.warn("Could not decode text using given encoding. Using default instead.", e);
+
+      text = new String(content);
```
srossbach: The chances are good that this trigger another inconsistency.

tobous: Yes, but I don't know what else we should/could do in this situation. As the encoding is specified by the client (who *should* have created the byte array with the given encoding), this case *should* not be possible.

Do you have a suggestion what else to do in this case?

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorManipulator.java
tobous: Done.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorManipulator.java
tobous: Done.

tobous:
Just rebased onto the current master. @srossbach Could you re-approve?

tobous:
This PR has only been rebased since the last approval. Last approval was more than 2 days ago, so I deem this PR ready to merge.

PR #338, opened 2018-12-04
[FIX][I] #117 Add logic to adjust local annotations in document listener

### LINKED ISSUE #117: Correctly modify annotations when a file is modified by save actions when it is closed
Saros/I supports a basic save action that strips out trailing whitespace. By default, this option is enabled for changed lines other than the current caret line. But IntelliJ also can be extended to support custom save actions through a plugin. These save action are applied when the editor for a file is closed.

This is currently a problem for the handling of annotations in Saros/I. Since the editor  already seems to be seen as closed when these save actions are applied, the resulting local edits are no longer tracked by the range highlighters of the annotations for the file. But since these are local edits, these changes are also not tracked by the annotation logic dealing with changes to closed files. This can result in annotations no longer marking the correct section of text.

#### Steps to reproduce:
0. Bob has the whitespace striping save action enabled (Editor > General: "Strip trailing spaces on Save")
(Since the default settings also save the editor when it loses focus, this save action or automatic saving should be disabled for Alice when testing locally on one machine.)
1. Alice starts a session with Bob
2. Alice adds new lines containing trailing whitespace (for example through auto indents)
3. Bob closes his editor without saving it first
(If only one line with trailing whitespace was added, please ensure that it is not the current caret line for Bob or disable "Always keep trailing spaces on caret line")
4. Bob re-opens the closed editor

This will result in the annotations for the added lines (and whitespace) to be moved to a later position in the file, since the annotations did not track the shortening of the file through the striping of trailing whitespace. This also applies to any annotation located behind a section changed by the save action. If an annotations was located directly at the end of the file, this will also lead to an exception as the AnnotationManager will try to add a RangeHighlighter to a position after the end of the file.

tobous:
Adds logic to StoppableDocumentListener.beforeDocumentChange(...) to
adjust the local annotations for the changed document. This is necessary
if the document is not open in a local editor, meaning text annotations
won't be automatically adjusted by Intellij. This can happen when
changes are cause by save actions or through automatic actions taken
during refactorings.

Fixes #117.


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/StoppableDocumentListener.java
```diff
@@ -22,15 +24,23 @@
 
   private static final Logger LOG = Logger.getLogger(StoppableDocumentListener.class);
 
-  StoppableDocumentListener(EditorManager editorManager) {
+  private final ProjectAPI projectAPI;
```
srossbach: See line 17. The Javadoc does not longer match the behavior and I do not know if this is the best place to do such logic. You basically spread the responsibility of the AnnotationManager to different classes. First the FileSystemChangeListener and now this listener. If you have to maintain the code you first have to find these places to understand the logic.

TLDR: Isnt there a better option so this logic can be placed into the AnnotationManager instead ?

tobous: This is a problem with the naming of the `AnnotationManager`. A more fitting name for the current functionality would probably be `AnnotationUtils` or something of the sort. (But I am not really happy with `AnnotationUtils` either. As always, I am open to suggestions.)

The current name stems form the initial plan for the implementation, which was to consolidate all the functionality into the annotation package (either into the `AnnotationManager` or a related class). This would require to also hook into all the listeners that the logic is present in currently, which is definitely possible.

The problem with this solution is that the execution of the annotation stuff is dependent on the interaction with the editor in some cases:
If the modified document is open in an editor, text annotations are automatically managed by Intellij. If I, for example, insert a character at a certain position, all annotations located after the position will be moved to a later position according to inserted text length. This causes the result of annotation "actions" (adding, removing, etc.) to be order dependent. Here is an easy example for inserting a character sequence into an opened document:
1. text change -> annotation change
If I add a character sequence at position `x` with the length `l`, any annotation located after `x` will be moved by `l` characters. If I then add the corresponding contribution annotation, it will mark the section from `x` to `x+l`.
2. annotation change -> text change
If I add the contribution annotation first, it will initially also mark the section from `x` to `x+l`. If I then add the character sequence at position `x` with length `l`, the contribution annotation which should mark the added text is moved back to the section from `x+l` to `x+2*l`.

Since I didn't want to deal with this problem (as I would have to come up with a workaround for case 2 or a way to order the listener executions), I decided to include the calls to the `AnnotationManager` in the original listeners to ensure that we are always dealing with case 1.

(A possible workaround would be to always close the editor when modifying the contained document, but this would introduce a not insignificant amount of addition computation time while not providing that much of a benefit.)

tobous: But since this is a special case where the document is not modified but we are reacting to a local modification, this case could be extracted to a separate listener. So yes, it would be better to move this usage of the `AnnotationManager` to a separate class. Will rework the commit.

tobous: This causes a lot of code duplication, which I solved by introducing an abstract class that implements `setEnabled(boolean)` and that provides `getSPath(Document)` which includes the logic to get an SPath for a Document. This means that the current `StoppableDocumenListerner` does almost nothing. I still have to do some minor improvements (Javadoc, etc.), so I will push it tomorrow. But I am not sure if it is really such a big improvement, so I will still hold on to a backup of the current patch in case we decide against the new implementation.

tobous:
Rebased onto master.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorManager.java
```diff
@@ -399,6 +400,7 @@ private void startSession(ISarosSession newSarosSession) {
           session.addActivityConsumer(consumer, Priority.ACTIVE);
 
           documentListener.setEnabled(true);
+          annotationDocumentListener.setEnabled(true);
```
srossbach: Maybe for later. If I observe this correctly is that what you really need is a disable-"bare" broadcaster, e.g

editorManager().getDocumentListenerBroadCast().addListener(…)

As there are currently only 2 listeners I do not see a reason for doing this now.

PR #341, opened 2018-12-05
[CORE] Makes SessionManager work asynchronous

srossbach:
The SessionManager now works asynchronously, pushing the tasks to a
executor. This should resolve the complicated concurrency issues that
are currently hard to maintain (e.g double locked sections)

Change-Id: I2c135ff64743507ce6cd2060f332bec0f04a882b

[CORE] Makes SessionManager work asynchronous

The SessionManager now works asynchronously, pushing the tasks to a
executor. This should resolve the complicated concurrency issues that
are currently hard to maintain (e.g double locked sections)

Change-Id: If463e1d05ecffefd1c8784c6918d0f232b1902f6


srossbach:
The discussion should be about the concept. Currently there are still import issues (I think Tobias already add a PR for this) and some of the test currently fails because the behavior has slightly changed.

tobous:
@srossbach Yes, the automatic import sorting is fixed with #330.

tobous:
Please disable the git-review functionality in you local git setup and remove the change-id (and the duplicated text) from the commit message.

tobous:
@srossbach What is the status of this PR?
PR #355 states that it introduces conflicts with the changes made in this PR.

srossbach:
I guess it can be closed. I do not play rebase games. Especially the non-project sharing stuff annoys me. The server has multiple other issues regarding operation in a production environment.

srossbach:
We had a broken non-host sharing feature. It didn't work well so we had to disable it. Nobody seems to miss it and the only reason it was reintroduced was because of the server stuff. The whole project sharing stuff is broken (since the session supports more than 2 users and multiple projects) if a user cancel it or an error occurs. But instead of fixing this issue you add a feature on top of a broken feature.

Drakulix:
> We had a broken non-host sharing feature. It didn't work well so we had to disable it. Nobody seems to miss it and the only reason it was reintroduced was because of the server stuff.

Nevertheless this is an improvement in usability for Saros/E as well.

> The whole project sharing stuff is broken (since the session supports more than 2 users and multiple projects) if a user cancel it or an error occurs. But instead of fixing this issue you add a feature on top of a broken feature.

If a user cancels a negotiation, he is kicked from a session. This is indeed broken and something, that should be fixed. But certainly out-of-scope for my patch.

If the host cancels an incoming project from another client, the project never gets added to the session => no error. So how is my feature making the situation any worse?

m273d15:
Is this PR still relevant? The last action was in february...

PR #350, opened 2018-12-10
[FIX][CORE] Moved FollowMode activity creation to core

srossbach:
It seems that the FollowingActivitiesManager was skipped during
migration. Indeed it contains questionable logic that should not be part
of the core. However this patch now moves at least the activity
generation part into the core.

The remaining functionality should be moved to the core once it is needed by other IDE implementations.

Removes the FollowingActivitiesManager class present in the IntelliJ implementation as it was an outdated copy of the Eclipse class. Subsequently removes some other unused classes that were copied from Saros/E.


tobous:
Rebased onto current master.

srossbach:
@tobous It seems I am still the owner of this patch and so I cannot approve it. There was a reason I put you on the review list. I appreciate your additional commits however I do not know if Kevin or Stefan have good knowledge on the impact that this patch introduce (it has none). I added them as reviewers mainly to share knowledge.

tobous:
Just rebased onto the master.

@srossbach This change fit into my plans perfectly as I am currently working on fixing the follow mode for Saros/I by switching to using the FollowModeManager instead of the current "solution".

@m273d15 @stefaus Up for a short review? Since I wrote the second half of this patch, I removed myself as a reviewer (as it doesn't really make sense reviewing my own changes). The review shouldn't really take very long as this PR just extracts a listener to a separate class and throws away a bunch of unnecessary classes in the Intellij part.

tobous:

PR #354, opened 2018-12-13
[C][BUGFIX] Fix missing export for apache.commons.lang3

Drakulix:
Session setup using eclipse is broken (resulting in an established session on both sides and a shared project on one side).

The issue was introduced with PR #346, likely the missing export was forgotten to be committed.
The reason this slipped through is, that the PR did not go through the usual review process (was merged although there were open issues with the code) and honestly, I feel that one should have been reverted immediately.

So this is in master right now and at least breaking my development workflow.
I hope this can be merged quickly.


srossbach:
1. This happens often
2. Should be found by CI
3. Was not my duty as I did not introduced the library
4. The probably that this issue is not noticed during review is high unless you are aware of this OSGi stuff
5. There is no need to revert something most of the times, especially when the fix is a NOP / one liner

PR #355, opened 2018-12-14
[FEATURE] Non-host project sharing

Drakulix:
Continuation of @denisw patchset [gerrit#2264](https://saros-build.imp.fu-berlin.de/gerrit/#/c/2264/9)

This patch allows users other than hosts to add projects to the session.

This works by first sharing the projects with the host, who then shares the project with all other users in the session. Normal project negotiations are used for both sharing steps. The host is free to reject non-host projects, and thus still in complete control over which projects are part of the session.

Because non-hosts don't consider projects they want to share as part of the session until the host accepts them, `AbstractOutgoingProjectNegotiation` (and all derived `ProjectNegotiation`)s cannot rely on `SarosSession` to find out which resources to share.

For this reason, a `ProjectSharingData` class encapsulating this information is introduced and passed to the `OutgoingProjectNegotiation`s constructors.

Note that the ID of a project is always chosen by the project negotiation's initiator, even if that is not the host. While that doesn't increase the probability of ID collisions per se, it allows a malicious non-host to choose a conflicting ID *deliberaitely*. Given that the Saros protocol seems to assume a relatively high amount of trust between the participants, though, this may not be such a big problem.

First tests work even in complex scenarios (3+ users). No modification during transmission of either the client->host or the host->other clients by the host or the first sender was not correctly queued and transmitted to all participants afterwards. No inconsistencies have yet been found.

This was not tested with the new and experimental `InstantOutgoingProjectNegotiation`, although steps were taken to support that as well.

@stefaus I have [moved](https://github.com/saros-project/saros/pull/355/files#diff-ba48fd2931f14e2e2d97b270df6eac2eR139) the queuing code out of the `Archive`- and `InstantIncomingProjectNegotiation` as they were identical and I had to add a `isHost` check there.
This will probably conflict with your PRs #156 (and likely other related ones).

This also conflicts with @srossbach's #341, replacing the `SessionManager` queuing infrastructure.

This patch also deliberately ignores @stefaus new `ProjectNegotiationCollector` introduced with #272 and instead directly runs the follow-up `ProjectNegotiation`s for the other clients in the [`negotiationTerminated` handler](https://github.com/saros-project/saros/pull/355/files#diff-1fd39cf856f7eb86e245d7507b4f2792R138) of the first (client->host) negotiation.

By doing so the multiple project negotiations of this process cannot be interrupted by other project negotiations, which should not be a problem code-wise, but is not desirable from the a user-experience point of view. The user should perceive the whole process as one giant negotiation.

The second commit allows the Eclipse UI to share projects with the host.
@tobous Are similar changes necessary for the IntelliJ UI? Can you give me some hints, what I would need to look for?

~~**Note**: For everyone wanting to test this, you also need #354 as project sharing is currently broken in the master.~~ merged.


tobous:
I don't think dealing with non-host project sharing in Saros/I is sensible right now. Currently, the session is restricted to the initially shared module. This restriction is also present in the UI (the functionality to add modules or start a session with multiple modules does not exist).
A sensible next step would be to allow multiple modules first. Then we can start dealing with enabling non-host sharing. (Since the logic is located in the core, this should only be a UI modification anyway, as you already stated.)

srossbach:
My 1000 feet review. I guess that this will not work with the final implementation of the instant session start. Of course it works. The host buffers all activities until it has fully received the project.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/InstantIncomingProjectNegotiation.java
```diff
@@ -64,30 +62,6 @@ protected void transfer(
       IProgressMonitor monitor, Map<String, IProject> projectMapping, List<FileList> missingFiles)
```
stefaus: moving logic of the transfer method to the abstract class interferes with Instant Session Start as it will slightly differ with #156

Drakulix: In that case rather then moving the queueing logic out of the `transfer` method into the `AbstractIncomingProjectNegotiation` we should move it into a new method (`handleQueueing` or similar). The whole reason I have moved the code is [this `isHost` check](https://github.com/saros-project/saros/pull/355/files#diff-ba48fd2931f14e2e2d97b270df6eac2eR169), which then could still be handled by the `AbstractIncomingProjectNegotiation` (after the `handleQueueing` call and before the `transfer` call).

Right now everything is neatly implemented on the `AbstractIncomingProjectNegotiation` layer and I would like to avoid having to move parts of this patch into the actual implementations.

Do you think that could work out?

Drakulix:
Rebased and added STF test

de.fu_berlin.inf.dpp/test/stf/de/fu_berlin/inf/dpp/stf/test/consistency/EditDuringNonHostAddsProjectTest.java
```diff
@@ -7,20 +7,17 @@
...
+public class EditDuringNonHostAddsProjectTest extends StfTestCase {
 
   @BeforeClass
   public static void selectTesters() throws Exception {
     select(ALICE, BOB, CARL);
   }
 
   @Test
-  @Ignore("Non-Host adding projects is currently deactivated")
   public void testDelayedHostAccept() throws Exception {
```
srossbach: I recommend you change the test setup a little bit.

1. After Bob starts the sharing he continues to write
2. Alice accept the negotiation and after Alice received the project Alice starts writing too
3. Carl accepts and after Carl finished the negotiation he adds also some text.
4. Everyone stops typing.
5. Compare the results.

Drakulix: Done, still works just fine :)

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
```diff
@@ -607,17 +648,39 @@ private synchronized void startNextProjectNegotiation() {
...
+    List<User> recipients = new ArrayList<>();
+    if (session.isHost()) {
+      /*
+       * If we received these project from a non-host user previously,
+       * that user already has the project.
+       */
+      users:
+      for (User user : session.getRemoteUsers()) {
+        for (IProject project : projectSharingData) {
+          if (session.userHasProject(user, project)) {
```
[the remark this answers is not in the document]
Drakulix: Oh god, partial sharing really is a feature that breaks so much. Any ideas how we could solve this one?

Before the Project was just shared with every remote user, so this would probably be fixed, if it would be shared with everyone, but the one that shared the Project with the host?

So in your scenario Alice would share it with Carl a second time, that should fix it, right?

Drakulix:
> A thing I don't like is that when a host cancels a incoming negotiation this also ends the whole session.
> Like always partial sharing is fun...

I have no idea, how this could be handled any better with the current infrastructure.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java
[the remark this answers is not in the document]
stefaus: > So in your scenario Alice would share it with Carl a second time, that should fix it, right?

Yes, we have to start a negotiation like Bob did with Alice (including the new files)

I have a suggestion that would fix both scenarios: add `User originUser` to
`private void executeOutgoingProjectNegotiation(ProjectSharingData projectSharingData)`
replace the block in line 652:
``` java
List<User> recipients = new ArrayList<>();
if (session.isHost()) {
  for (User user : session.getRemoteUsers()) {
    if (user != originUser) recipients.add(user);
  }
} else {
  recipients.add(session.getHost());
}
```

and call like
https://gist.github.com/stefaus/b641b99753a035c7b357bdf5f2acb51c#file-sarossessionmanager-java-L156-L157
and
https://gist.github.com/stefaus/b641b99753a035c7b357bdf5f2acb51c#file-sarossessionmanager-java-L635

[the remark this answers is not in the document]
Drakulix: That is what I had in mind as well. Added/Fixed

Drakulix:
> > I have no idea, how this could be handled any better with the current infrastructure.
> 
> I haven't thought this to the end (this has maybe side effects), but maybe a check at this position would solve it already.
> 
> [saros/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractIncomingProjectNegotiation.java](https://github.com/saros-project/saros/blob/fb9980986bd34465ded1e6a701408857b90f1891/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractIncomingProjectNegotiation.java#L274)
> 
> Line 274 in [fb99809](/saros-project/saros/commit/fb9980986bd34465ded1e6a701408857b90f1891)
>  sessionManager.stopSession(SessionEndReason.LOCAL_USER_LEFT);

I have added a corresponding check. I will tests this later, but I am quite confident, that it should work. The `OutgoingAbstractProjectNegotiation` has a similar check, which I also extended. That code has just relied on the fact, that `Outgoing`-negotiations were always send by the host, while `Incoming`-negotiation, where always received.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractOutgoingProjectNegotiation.java
```diff
@@ -234,7 +248,7 @@ protected void sendFileList(List<ProjectNegotiationData> projectInfos, IProgress
 
   @Override
   protected void executeCancellation() {
-    if (session.getRemoteUsers().isEmpty())
+    if (!session.isHost() || session.getRemoteUsers().isEmpty())
```
stefaus:
```suggestion
    if (session.isHost() && session.getRemoteUsers().isEmpty())
```
you have to check if we are the host, otherwise the client disconnects on a canceled outgoing negotiation

Drakulix: You are right, swapped that up!

stefaus:
Ok, one more thing, when the host cancels a incoming project negotiation, the GUI asks before if the host is sure about it and tells that this action will close the session (which isn't happening now). To skip this dialog with wrong information you could add a check in `AddProjectToSessionWizard.performCancel()`
https://gist.github.com/stefaus/b641b99753a035c7b357bdf5f2acb51c#file-addprojecttosessionwizard-java-L248-L258

Drakulix:
> Ok, one more thing, when the host cancels a incoming project negotiation, the GUI asks before if the host is sure about it and tells that this action will close the session (which isn't happening now). To skip this dialog with wrong information you could add a check in `AddProjectToSessionWizard.performCancel()`
> https://gist.github.com/stefaus/b641b99753a035c7b357bdf5f2acb51c#file-addprojecttosessionwizard-java-L248-L258

Added a fix. Do you think `sessionManager.getSession` needs a null-check here? I am not certain that the session might not have been closed somehow in the meantime, whats your take on this?

stefaus:
> Do you think `sessionManager.getSession` needs a null-check here?

Yes, I think so... and looking at the Impl, I notice that it even has an (incorrectly placed) deprecation mark... but I would say use it anyway, but check it
https://github.com/saros-project/saros/blob/4a43f816677fdc1b92d6638ff29b3135f019d4a2/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java#L348-L358

PR #358, opened 2018-12-18
[STF] Add SarosHTMLView for superBot

jafey:
The Method superbot().views().sarosView() now decides the instance of
the ISarosView. It will return the new SarosHTMLView if the HTML-GUI is
activated and the forceUseHtmlGUI() Method is called before. Otherwise
it will return the old SarosView.
Most of the methods in SarosHTMLView are not yet implemented,
this is because the HTML-GUI actually doesn't provide all features.

### DIFF
--- a/de.fu_berlin.inf.dpp/test/framework/stf/src/de/fu_berlin/inf/dpp/stf/server/rmi/superbot/component/view/impl/Views.java
+++ b/de.fu_berlin.inf.dpp/test/framework/stf/src/de/fu_berlin/inf/dpp/stf/server/rmi/superbot/component/view/impl/Views.java
@@ -1,6 +1,10 @@
 package de.fu_berlin.inf.dpp.stf.server.rmi.superbot.component.view.impl;
 
+import de.fu_berlin.inf.dpp.Saros;
 import de.fu_berlin.inf.dpp.stf.server.StfRemoteObject;
+import de.fu_berlin.inf.dpp.stf.server.rmi.htmlbot.EclipseHTMLWorkbenchBot;
+import de.fu_berlin.inf.dpp.stf.server.rmi.htmlbot.IHTMLBot;
+import de.fu_berlin.inf.dpp.stf.server.rmi.htmlbot.impl.HTMLBotImpl;
 import de.fu_berlin.inf.dpp.stf.server.rmi.remotebot.impl.RemoteWorkbenchBot;
 import de.fu_berlin.inf.dpp.stf.server.rmi.superbot.component.view.IViews;
 import de.fu_berlin.inf.dpp.stf.server.rmi.superbot.component.view.eclipse.IConsoleView;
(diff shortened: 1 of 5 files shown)


de.fu_berlin.inf.dpp/test/framework/stf/src/de/fu_berlin/inf/dpp/stf/server/rmi/superbot/component/view/impl/Views.java
```diff
@@ -24,12 +30,34 @@ public static Views getInstance() {
...
+  private boolean forceUseHtmlGui = false;
+
+  /* Use this method to force the STF to use the HTML-GUI.
+   * This can be removed when the old GUI is fully replaced.
+   * */
+  @Override
+  public void forceUseHtmlGui() throws RemoteException {
+    if (Saros.useHtmlGui() == false) {
+      throw new RuntimeException(
+          "HTML-GUI is not enabled. Please set saros.swtbrowser propertie to true.");
```
fzieris: typo: "property"

de.fu_berlin.inf.dpp/test/framework/stf/src/de/fu_berlin/inf/dpp/stf/server/rmi/superbot/component/view/impl/Views.java
```diff
@@ -24,12 +30,34 @@ public static Views getInstance() {
     return INSTANCE;
   }
 
+  private boolean forceUseHtmlGui = false;
+
+  /* Use this method to force the STF to use the HTML-GUI.
```
fzieris: This doesn't look auto-formatted. Also: Such comments belong in the interface.

de.fu_berlin.inf.dpp/test/framework/stf/src/de/fu_berlin/inf/dpp/stf/server/rmi/superbot/component/view/impl/Views.java
```diff
@@ -24,12 +30,34 @@ public static Views getInstance() {
...
+    forceUseHtmlGui = true;
+  }
+
   @Override
   public ISarosView sarosView() throws RemoteException {
-    SWTWorkbenchBot bot = new SWTWorkbenchBot();
-    RemoteWorkbenchBot.getInstance().openViewById(VIEW_SAROS_ID);
-    bot.viewByTitle(VIEW_SAROS).show();
-    return SarosView.getInstance().setView(bot.viewByTitle(VIEW_SAROS));
+    if (Saros.useHtmlGui() && this.forceUseHtmlGui) {
```
fzieris: It's confusing to have a setter method and the underlying field with the same name. Maybe something like `private boolean shouldTestHtmlGui` and `public void runTestsOnHtmlGui()`?

By the way: Who is supposed to call `forceUseHtmlGui()`? If it's going to be the same party that calls `sarosView()`, why introduce a field and a new method instead of simply a boolean parameter for `public ISarosView sarosView(boolean testHtmlGui)`?

jafey: I also thought about it with Kelvin. The disadvantage is that you have to adjust the `sarosView()` Method in every test. Or do you mean to introduce a second `sarosView(boolean testHtmlGui)` method?

fzieris: I don't have a preference. But since it's Java Code and we control all the code, even the first option would boil down to a single refactoring.

de.fu_berlin.inf.dpp/test/framework/stf/src/de/fu_berlin/inf/dpp/stf/server/rmi/superbot/component/view/impl/Views.java
jafey: fixed in the second commit

de.fu_berlin.inf.dpp/test/framework/stf/src/de/fu_berlin/inf/dpp/stf/server/rmi/superbot/component/view/impl/Views.java
jafey: fixed in the second commit

de.fu_berlin.inf.dpp/test/framework/stf/src/de/fu_berlin/inf/dpp/stf/server/rmi/superbot/component/view/impl/Views.java
jafey: I renamed the variable and the method to your suggestion.

m273d15:
Close this due to inactivity. I am currently creating a list/doc of started/incomplete projects. This should allow us to find such closed (but probably interesting) PRs in the future.

PR #362, opened 2018-12-20
[E/S][FEATURE] Allow Saros/E to join server session

Drakulix:
Allows Saros/E to request an invitation to a server session.

Saros/E implements a `JoinSessionRequestHandler` as well, but that is based of the old Eclipse-Server-Prototype and not usable in it's current state. For that reason the "Request Session Invitation" menu item is only shown, if the contact advertises the server feature. Same goes for the action, it is a no-op, if the JID does not belong to a server.

Along with #355 this allows the server to be run completely headless (without using the console).

### DIFF
--- a/eclipse/src/saros/ui/actions/RequestSessionInviteAction.java
+++ b/eclipse/src/saros/ui/actions/RequestSessionInviteAction.java
@@ -1,0 +1,77 @@
+package saros.ui.actions;
+
+import java.util.List;
+import org.eclipse.jface.action.Action;
+import org.eclipse.jface.viewers.ISelection;
+import org.eclipse.ui.ISelectionListener;
+import org.eclipse.ui.IWorkbenchPart;
+import saros.SarosPluginContext;
+import saros.communication.extensions.JoinSessionRequestExtension;
+import saros.net.ITransmitter;
+import saros.net.xmpp.JID;
+import saros.repackaged.picocontainer.annotations.Inject;
+import saros.session.ISarosSession;
+import saros.session.ISarosSessionManager;
+import saros.ui.Messages;
+import saros.ui.util.selection.SelectionUtils;
+import saros.ui.util.selection.retriever.SelectionRetrieverFactory;
+
+/**
+ * Action for requesting an invitation to a session from a contact.
+ *
+ * <p>This currently relies on the fact, that only Saros/S has a working JoinSessionRequestHandler.
+ * To make this feature generic in the future we need to add another XMPP namespace
+ */
+public class RequestSessionInviteAction extends Action implements Disposable {
+
+  @Inject private ISarosSessionManager sessionManager;
+  @Inject private ITransmitter transmitter;
+
+  private ISelectionListener selectionListener =
+      new ISelectionListener() {
+        @Override
+        public void selectionChanged(IWorkbenchPart part, ISelection selection) {
+          updateActionState();
+        }
+      };
+
+  public static final String ACTION_ID = RequestSessionInviteAction.class.getName();
+
+  public RequestSessionInviteAction() {
+    super(Messages.RequestSessionInviteAction_title);
+    setId(ACTION_ID);
+    SarosPluginContext.initComponent(this);
+    SelectionUtils.getSelectionService().addSelectionListener(selectionListener);
+    updateActionState();
+  }
+
+  @Override
+  public void run() {
+    ISarosSession session = sessionManager.getSession();
+    JID jid = getSelectedJID();
+    if (session != null || jid == null) {
+      return;
+    }
+
+    transmitter.sendPacketExtension(
+        jid, JoinSessionRequestExtension.PROVIDER.create(new JoinSessionRequestExtension()));
+  }
+
+  private JID getSelectedJID() {
+    List<JID> selected = SelectionRetrieverFactory.getSelectionRetriever(JID.class).getSelection();
+
+    if (selected.size() != 1) return null;
+
+    return selected.get(0);
+  }
+
+  private void updateActionState() {
+    ISarosSession session = sessionManager.getSession();
+    setEnabled(session == null && getSelectedJID() != null);
+  }
+
+  @Override
+  public void dispose() {
+    SelectionUtils.getSelectionService().removeSelectionListener(selectionListener);
+  }
+}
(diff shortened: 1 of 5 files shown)


de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/views/SarosView.java
```diff
@@ -441,6 +446,13 @@ public void menuAboutToShow(final IMenuManager manager) {
              * version 14.1.31)
              */
             // manager.add(getAction(SkypeAction.class));
+
+            // TODO: Replace Server check once other implementations support JoinSessionRequests
+            if (contacts.size() == 1
+                && discoveryManager.queryFeatureSupport(
```
srossbach: Beside the NPE that can be thrown this can cause UI freezes, so I am sorry this is not something that is suiteable for production.

Drakulix: Well, it would allow for an alpha release of the saros server (for testing).

I could hide this behind an experimental feature flag, if you like, but as it only activates in the context of using a very experimental server build, I do not really see the point.

That said the whole server topic is really new and I get if at this point this is nothing to be merged. I will then just keep this PR for reference and demonstration purposes.

Drakulix: If this is just about the `discoveryManager` call, what about adding an additional function the `discoveryManager` that only checks the cache in therefor is not blocking?

srossbach: The DiscoveryManager has 3 public methods, should not be to hard the one you need: https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/net/xmpp/discovery/DiscoveryManager.java#L202

Drakulix: Done

de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/views/SarosView.java
```diff
@@ -233,6 +236,8 @@ public void sessionEnded(ISarosSession session, SessionEndReason reason) {
 
   @Inject protected XMPPConnectionService connectionService;
 
+  @Inject protected DiscoveryManager discoveryManager;
```
tobous: Is there a reason to use `protected` instead of `private`?

de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/actions/RequestSessionInviteAction.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private JID getSelectedJID() {
+    List<JID> selected = SelectionRetrieverFactory.getSelectionRetriever(JID.class).getSelection();
+
+    if (selected.size() != 1) return null;
+
+    return selected.get(0);
+  }
+
+  private void updateEnablement() {
```
tobous: Rename to `updateActionState()` or `updateEnabledState()`?

de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/actions/RequestSessionInviteAction.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Override
+  public void run() {
+    ISarosSession session = sessionManager.getSession();
+    JID jid = getSelectedJID();
+    if (session != null || jid == null) {
+      return;
+    }
+
+    // TODO: Currently only Saros/S is known to have a working JoinSessionRequestHandler,
```
tobous: In my opinon, it would be better if this would also be mentioned in the class javadoc.

de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/views/SarosView.java
```diff
@@ -441,6 +446,13 @@ public void menuAboutToShow(final IMenuManager manager) {
              * version 14.1.31)
              */
             // manager.add(getAction(SkypeAction.class));
+
+            // TODO: Replace Server check once other implementations support JoinSessionRequests
+            if (contacts.size() == 1
+                && discoveryManager.isFeatureSupported(
```
tobous: It would be better to query `discoveryManager` separately beforehand and do a null-check on the result before checking the boolean value.

As the method `isFeatureSupported` returns a `Boolean` object (and the primitive type `boolean`), the returned object can be null, at which point the implicit unboxing in the if statement will throw a nullpointer exception.

Drakulix:
@tobous Should all be fixed.

eclipse/src/saros/ui/views/SarosView.java
```diff
@@ -661,6 +674,7 @@ private void createActions() {
     registerAction(new SendFileAction());
     registerAction(new ChangeColorAction());
     registerAction(new RemoveUserAction());
+    registerAction(new RequestSessionInviteAction());
```
m273d15: Could you make this feature optional via introducing a properties in `saros.properties`? Otherwise, we would deploy the new button without deploying the server. This would confuse users.

eclipse/src/saros/ui/actions/RequestSessionInviteAction.java
```diff
@@ -0,0 +1,86 @@
...
+  public void run() {
+    ISarosSession session = sessionManager.getSession();
+    JID jid = getSelectedJID();
+    if (session != null || jid == null) {
+      return;
+    }
+
+    // TODO: Currently only Saros/S is known to have a working JoinSessionRequestHandler,
+    //       remove this once the situation changes / change this to it's own feature.
+    if (!discoveryManager.isFeatureSupported(jid, SarosConstants.NAMESPACE_SERVER)) {
```
tobous: Is there a reason to check twice whether the selected user is a Saros server (here and in line 445 of `SarosView`)? Or am I misunderstanding the checks?

This check is also missing the null-check for the result of ´discoveryManager.isFeatureSupported(...)`.

eclipse/src/saros/ui/views/SarosView.java
Drakulix: We would not deploy this button to any users, because it is only displayed, for servers is in the contact list.

eclipse/src/saros/ui/actions/RequestSessionInviteAction.java
Drakulix: removed

PR #365, opened 2019-01-08
Create a custom picocontainer jar

m273d15:
We are using the picocontainer 2 and intellij uses picocontainer 1. During the build of the Intellij plugin the integration of these two picocontainer libraries leads to conflicts.

The current solution is to exclude the library (and the JUnit library) with:
```gradle
intellij {
  ...
  configureDefaultDependencies = false
  ...
}
afterEvaluate {
  dependencies {
    compile intellij {exclude 'picocontainer*.jar', 'junit*.jar'}
  }
}
```
This workaround interferes with the common intellij IDEA development workflow, because the intellij sources are not available during development. Furthermore [@tobous chatted](https://gitter.im/JetBrains/gradle-intellij-plugin?at=5bfe8f35ed6bcf1ef8733c0d) with the developers of the intellij plugin and they wrote "this option is not documented, was introduced for very-very special case (which is unused now, by the way) and likely will be removed"

This is the implementation of an alternative resolution of this conflict. This PR introduces a build file that creates a new patched picocontainer jar where each package `org.picocontainer` is relocated to `saros.org.picocontainer`.


tobous:
@m273d15 Have you run the STF to check whether this breaks anything at runtime?

m273d15:
Not yet, I only tested a session manually. I will run the tests today.

m273d15:
Sorry I mixed up my changes. No I did not test this PR at all. I wanted to ask you whether this solution is what you expected. I have to check whether the solution works also with eclipse and during runtime

srossbach:
I do not understand the problem. So is this only a compile time issue or a runtime issue ?

tobous:
@srossbach Currently, this is only a compile-time issue (but it can also be a runtime issue; see next paragraph). But, to solve this problem, we are currently using some workarounds that are not officially supported by the developer and are hindering the workflow for developing Saros/I. (IntelliJ sources are no longer automatically linked, meaning they can't be easily accessed through the IDE. This makes working with the IntelliJ API cumbersome. There are, again, local workarounds for this, but they are cumbersome as well).
Furthermore, the developer of the intellij build plugin mentioned that there is a real possibility that the feature we are using for the workaround will be removed in the (near) future, so it would be better to look for an alternative anyway.

This issue now only came up again as I wanted to implement some test for the Saros/I resource implementation, more specifically the usage of the IntelliJ API. For this, I wanted to use [IntelliJ test fixtures](https://www.jetbrains.org/intellij/sdk/docs/basics/testing_plugins.html) to create a headless IntelliJ test instance. With this, however, the issue also becomes a runtime issue as we would need the IntelliJ PicoContainer version to create the test fixture but then would need our PicoContainer version to execute the tests. ([This is the same issue Holger Schmeisky struggled with back in 2014.](https://intellij-support.jetbrains.com/hc/en-us/community/posts/206764205-NoClassDef-found-for-PicoContainer-with-LightFixtureTestCase))

srossbach:
As for the compile time stuff, you only have to adjust the classpath. As for the runtime stuff I do not know the internals but it seems the classloading is not done correctly. I do not know how the plugin classloader works in IntelliJ but I would expect that it works like the OSGi classloader, i.e it first searches the plugin classpath before it delegates to the parent classloader.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/context/ContainerContext.java
```diff
@@ -31,14 +31,14 @@
...
 import org.apache.log4j.Logger;
-import org.picocontainer.ComponentMonitor;
-import org.picocontainer.MutablePicoContainer;
-import org.picocontainer.PicoBuilder;
-import org.picocontainer.PicoContainer;
-import org.picocontainer.injectors.AnnotatedFieldInjection;
-import org.picocontainer.injectors.CompositeInjection;
-import org.picocontainer.injectors.ConstructorInjection;
-import org.picocontainer.injectors.Reinjector;
+import saros.org.picocontainer.ComponentMonitor;
```
srossbach: The convention is to reverse the domain so it is: org.saros.* and not saros.org

tobous: My guess would be that this package name was created by just prepending `saros` to the default picocontainer package name (`org.picocontainer`). But I am not sure what a good package name/path would be.

m273d15:
I renamed the package from `saros.org.picocontainer` to `org.saros.picocontainer`.
This change requires that all developers are refreshing their eclipse environment (with calling `prepareEclipse`; I know this handling is stupid and I already think for an alternative), because the dependencies are changing. I would like to merge this with #385. Otherwise the eclipse developers have to refresh the environment again.

core/META-INF/MANIFEST.MF
```diff
@@ -139,7 +139,7 @@ Bundle-ClassPath: .,
  lib/commons-lang3-3.8.1.jar,
  lib/xstream-1.4.10.jar,
  lib/xpp3-1.1.4c.jar,
- lib/picocontainer-2.11.2.jar,
+ lib/picocontainer-2.11.2-patched_relocated.jar,
```
srossbach: So is this jar already patched or just repacked because the former approach was using classpath reordering so that the original class file was not loaded from the library but the patched one was loaded from our core library.

tobous:
As message updates are not send out as notifications: The issue of the missing classpath entries I reported was resolved by dropping the IntelliJ cache.

core/META-INF/MANIFEST.MF
m273d15: This jar contains the patched classes.

PR #369, opened 2019-01-08
[CORE][FEATURE]add Activity send Bundle File

chrismotian:
Followed the JTour-Bus-Guide for creating activity.
This activity will be triggered with a (not yet existing) button to fire some commits in a bundle file. It uses the JGitFacade introduced in #364.


chrismotian:
#372 will be used to introduce the changes. This PR got messy because I build it on top of the JGitFacade implementation.

PR #371, opened 2019-01-10
Use FollowModeManager to implement Saros/I follow mode functionality

tobous:
This PR contains multiple independent commits which should be review separately. Each commit has a commit message that gives some more information.

This PR does not fix the follow mode completely. For the follow mode to function correctly, there still have to be made some adjustments to the logic changing the local viewport and the consistency of the UserEditorState. For this reason, the saros.properties flag disabling the follow mode is not removed yet.

### PR structure

The commits in this PR can be grouped into the following sections:

#### Fix EditorManager methods
These commits fix some methods of the EditorManager used by the FollowModeManager.
- [DOC][C] Fix minor javadoc issues in IEditorManager
- [FIX][I] Fix EditorManager.adjustViewport(...)
- [FIX][I] Fix EditorManager.jumpToUser(...)
- [FIX][I] Fix EditorManager.generateEditorClosed(...)

#### Replace follow mode logic
These commits remove the custom logic that was added to realize the follow mode
functionality in Saros/I as it was redundant and not working correctly.
Replaces the logic with the usage of the FollowModeManager.

- [FIX][I] Use FollowModeManager instead of custom logic

#### Cleanup
These commits do some minor cleanup.
- [REFACTORING][I] Do minor cleanup

#### Add user notifications
These commits introduce user notifications for the follow mode.
- [UI][I] Add follow mode user notifications

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/followmode/FollowModeNotificationDispatcher.java
+++ b/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/followmode/FollowModeNotificationDispatcher.java
@@ -1,0 +1,69 @@
+package de.fu_berlin.inf.dpp.intellij.followmode;
+
+import de.fu_berlin.inf.dpp.editor.FollowModeManager;
+import de.fu_berlin.inf.dpp.editor.IFollowModeListener;
+import de.fu_berlin.inf.dpp.intellij.ui.Messages;
+import de.fu_berlin.inf.dpp.intellij.ui.util.NotificationPanel;
+import de.fu_berlin.inf.dpp.session.User;
+import java.text.MessageFormat;
+import org.apache.log4j.Logger;
+
+/** Displays user notifications about follow mode status changes. */
+public class FollowModeNotificationDispatcher implements IFollowModeListener {
+
+  private static final Logger log = Logger.getLogger(FollowModeNotificationDispatcher.class);
+
+  private User followedUser;
+
+  public FollowModeNotificationDispatcher(FollowModeManager followModeManager) {
+    followModeManager.addListener(this);
+  }
+
+  @Override
+  public void stoppedFollowing(Reason reason) {
+    String shownReason;
+
+    switch (reason) {
+      case FOLLOWEE_LEFT_SESSION:
+        shownReason = Messages.FollowModeNotificationDispatcher_end_reason_FOLLOWEE_LEFT_SESSION;
+        break;
+      case FOLLOWER_CLOSED_OR_SWITCHED_EDITOR:
+        shownReason =
+            Messages.FollowModeNotificationDispatcher_end_reason_FOLLOWER_CLOSED_OR_SWITCHED_EDITOR;
+        break;
+      case FOLLOWER_CLOSED_EDITOR:
+        shownReason = Messages.FollowModeNotificationDispatcher_end_reason_FOLLOWER_CLOSED_EDITOR;
+        break;
+      case FOLLOWER_STOPPED:
+        shownReason = Messages.FollowModeNotificationDispatcher_end_reason_FOLLOWER_STOPPED;
+        break;
+      case FOLLOWER_SWITCHES_FOLLOWEE:
+        shownReason =
+            Messages.FollowModeNotificationDispatcher_end_reason_FOLLOWER_SWITCHES_FOLLOWEE;
+        break;
+
+      default:
+        shownReason = reason.name();
+        log.warn("Encountered unknown reason for follow mode end. reason: " + reason);
+    }
+
+    NotificationPanel.showInformation(
+        MessageFormat.format(
+            Messages.FollowModeNotificationDispatcher_stopped_following_message,
+            followedUser,
+            shownReason),
+        Messages.FollowModeNotificationDispatcher_stopped_following_title);
+
+    followedUser = null;
+  }
+
+  @Override
+  public void startedFollowing(User target) {
+    NotificationPanel.showInformation(
+        MessageFormat.format(
+            Messages.FollowModeNotificationDispatcher_started_following_message, target),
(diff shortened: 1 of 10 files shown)


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/followmode/FollowModeNotificationDispatcher.java
```diff
@@ -0,0 +1,66 @@
+package de.fu_berlin.inf.dpp.intellij.followmode;
+
+import de.fu_berlin.inf.dpp.editor.FollowModeManager;
+import de.fu_berlin.inf.dpp.editor.IFollowModeListener;
+import de.fu_berlin.inf.dpp.intellij.ui.Messages;
+import de.fu_berlin.inf.dpp.intellij.ui.util.NotificationPanel;
+import de.fu_berlin.inf.dpp.session.User;
+import org.apache.log4j.Logger;
+
+public class FollowModeNotificationDispatcher implements IFollowModeListener {
```
srossbach: Please just add a small JavaDoc.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/followmode/FollowModeNotificationDispatcher.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private static final Logger log = Logger.getLogger(FollowModeNotificationDispatcher.class);
+
+  private volatile User followedUser;
```
srossbach: IIRC all calls are done in the Swing thread so there is no need to use volatile.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/followmode/FollowModeNotificationDispatcher.java
```diff
@@ -0,0 +1,66 @@
...
+        shownReason = Messages.FollowModeNotificationDispatcher_end_reason_FOLLOWER_STOPPED;
+        break;
+      case FOLLOWER_SWITCHES_FOLLOWEE:
+        shownReason =
+            Messages.FollowModeNotificationDispatcher_end_reason_FOLLOWER_SWITCHES_FOLLOWEE;
+        break;
+
+      default:
+        shownReason = reason.name();
+        log.error("Encountered unknown reason for follow mode end. reason: " + reason);
```
srossbach: Warning should be enough at it is only display logic and not something critical.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/ui/actions/FollowModeAction.java
```diff
@@ -37,21 +33,17 @@ public void sessionEnded(ISarosSession oldSarosSession, SessionEndReason reason)
 
           ThreadUtils.runSafeAsync(
```
srossbach: I do not get the reason why we have to run a thread to set those variables. You could either dispatch this to the Swing Thread / EDT or just set the variables. See also line 24, same issue.

tobous: I just modified the existing code. Removed.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/followmode/FollowModeNotificationDispatcher.java
tobous: Done.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/followmode/FollowModeNotificationDispatcher.java
tobous: Done.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/followmode/FollowModeNotificationDispatcher.java
tobous: I am not certain I understand what you mean. Can you elaborate?

tobous:
Resolved issues mentioned in review.
Use `MessageFormat.format(...)` instead of `String.format(...)` for message formatting.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/followmode/FollowModeNotificationDispatcher.java
srossbach: Lets reverse this question. Why do you think this variable has to be volatile ?

tobous: I thought there could be concurrency problems when the currently followed user leaves the session (which is handled by the Saros dispatcher logic) while the local user chooses to follow another user (which is handled on the IntelliJ UI thread). But you are right, as the listeners are notified  on the swing thread (or the IntelliJ equivalent) as well, this is not necessary. Will remove.

tobous:
Removed unnecessary `volatile` modifier in `FollowModeNotificationDispatcher`.

tobous:
Fixed menu item for leaving the follow mode not working caused by a wrong null check where null is the correct value.

tobous:

PR #376, opened 2019-01-10
[INTERNAL] [HTML WB] Add Java Bridge

AbdBarho:
responsible for converting and formatting the data received from Java and redirects it to ViewManager so it can be applied.

the data is received as "Records" in a hierarchical structure but then flatten to an array.

these records can be "new", "set" and "remove" which have a corresponding function to handle them.


de.fu_berlin.inf.dpp.whiteboard/frontend/src/services/javabridge.js
```diff
@@ -0,0 +1,145 @@
...
+    window.message = this.message = (sxeMessage) => {
+      let records = sxeMessage.records || [];
+      if (sxeMessage.messageType === SXE.MESSAGE_TYPES.STATE) {
+        //we ignore root element, all html whiteboard objects will be added to root element
+        records.shift();
+      }
+      records.forEach(record => this.executeRecord(record));
+      this.canvas.renderAll();
+    };
+    //sets the state of the whiteboard at the begining of the session using data from the Java Model
```
m273d15: typo "beginning"

de.fu_berlin.inf.dpp.whiteboard/frontend/src/services/javabridge.js
```diff
@@ -0,0 +1,145 @@
...
+          for (let attr of Object.values(old.$attrs))
+            view.$createAttr(attr.name, attr.chdata, attr.rid);
+          this.viewManager.register(view);
+        } else {
+          view = this.viewManager.getView(parent);
+        }
+        if (!view)
+          return;
+        let attr = view.$createAttr(name, chdata, rid);
+        delete attr.isDirty;
```
m273d15: why do you have to remove this attribute `isDirty`?

de.fu_berlin.inf.dpp.whiteboard/frontend/src/services/javabridge.js
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  /**
+   * applies a "SET" record
+   * @param {Object} recordProps
+   */
+  set(recordProps) {
+    let { chdata, visible, target } = recordProps;
+    // set visible="false" is a workaround to delete elements
+    if (visible === "false") {
```
m273d15: Why do you compare in L.66 with boolean `false` and here with string `"false"`. Did I missed a conversion step or is this JS magic I don't know? I would assume that the string version is slower...

AbdBarho: TLDR; *NEW* records have the value as a Boolean, while *SET* records have the value as a string.

In Java:
A NEW record represents the actual object in Java, so the value is Boolean.
SET records describe a change done on the object, so when a SET record is received while syncing, it contains the serialised value of "false" or "true".

When applying, the value is converted to Boolean, however, SET records are saved exactly the way the are received without the de-serialisation step (the idea behind saving them is to have a history of the states of the object so that in case of a conflict or sync problems, the object can be restored to a previous state).

So the conversion happens only when applying the SET records to a NEW record, resulting in Boolean value in NEW records, but remaining as a string in SET records.

This remains the same when applying the records in the HTML Whiteboard, because the records are taken "as is" from Java.

The whole idea of setting visible to false instead of using REMOVE records was a kind of "hack" to improve performance ( the SXE REMOVE record requires that the element has no children, which means we have to send REMOVE records for all of the children from the bottom up until we reach the parent, resulting in a big overhead)

de.fu_berlin.inf.dpp.whiteboard/frontend/src/services/javabridge.js
AbdBarho: Yes, I understand the confusion, I changed it to `attr.isDirty = false;`

in JavaScript accessing an attribute that does not exist on an object will return `undefined` which behaves almost exactly the same as `false` when casted to boolean.

PR #379, opened 2019-01-13
[REFACTOR] Remove unnecessary TransferType usage

stefaus:
TransferType is set via session negotiation mechanism at session start
for the whole session making it unnecessary to transmit it again at 
project negotiation time.

### DIFF
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/NegotiationFactory.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/NegotiationFactory.java
@@ -130,13 +130,12 @@ public final class NegotiationFactory {
 
   public AbstractOutgoingProjectNegotiation newOutgoingProjectNegotiation(
       final JID remoteAddress,
-      TransferType transferType,
       final ProjectSharingData projectSharingData,
       final ISarosSessionManager sessionManager,
       final ISarosSession session) {
 
     User remoteUser = session.getUser(remoteAddress);
-    transferType = getTransferType(session, remoteUser);
+    TransferType transferType = getTransferType(session, remoteUser);
     if (transferType == null) {
       throw new IllegalArgumentException("transferType must not be null");
     }
@@ -173,14 +172,13 @@ public final class NegotiationFactory {
 
   public AbstractIncomingProjectNegotiation newIncomingProjectNegotiation(
       final JID remoteAddress,
-      TransferType transferType,
       final String negotiationID,
       final List<ProjectNegotiationData> projectNegotiationData,
       final ISarosSessionManager sessionManager,
       final ISarosSession session) {
 
     User remoteUser = session.getUser(remoteAddress);
-    transferType = getTransferType(session, remoteUser);
+    TransferType transferType = getTransferType(session, remoteUser);
     if (transferType == null) {
       throw new IllegalArgumentException("transferType must not be null");
     }
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/NegotiationFactory.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/NegotiationFactory.java
@@ -14,6 +14,8 @@ import de.fu_berlin.inf.dpp.net.xmpp.discovery.DiscoveryManager;
 import de.fu_berlin.inf.dpp.observables.FileReplacementInProgressObservable;
 import de.fu_berlin.inf.dpp.session.ISarosSession;
 import de.fu_berlin.inf.dpp.session.ISarosSessionManager;
+import de.fu_berlin.inf.dpp.session.ProjectNegotiationTypeHook;
+import de.fu_berlin.inf.dpp.session.User;
 import de.fu_berlin.inf.dpp.versioning.VersionManager;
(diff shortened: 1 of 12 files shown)


de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/NegotiationFactory.java
```diff
@@ -129,14 +131,12 @@ public IncomingSessionNegotiation newIncomingSessionNegotiation(
...
       final JID remoteAddress,
-      final TransferType transferType,
       final List<IProject> resources,
       final ISarosSessionManager sessionManager,
       final ISarosSession session) {
 
-    if (transferType == null) {
-      throw new IllegalArgumentException("transferType must not be null");
-    }
+    User remoteUser = session.getUser(remoteAddress);
```
srossbach: Duplicated code. Why not use getTransferType(ISarosSession session, JID remoteAddress) as signature ? Also the user may have already left the session and so this call would return null.

stefaus: Didn't removed the duplicate because I plan to change the signature of `new[In|Out]comingProjectNegotiation` to use `User`, but yes better change it for now.

stefaus:
@srossbach please review the change

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/NegotiationFactory.java
```diff
@@ -211,4 +207,18 @@ public AbstractIncomingProjectNegotiation newIncomingProjectNegotiation(
         throw new UnsupportedOperationException("transferType not implemented");
     }
   }
+
+  private TransferType getTransferType(ISarosSession session, JID remoteAddress) {
+    User user = session.getUser(remoteAddress);
+    if (user == null) {
+      throw new IllegalStateException("User <" + user + "> is not part of the session.");
```
srossbach: I do not know if this is the best decision (throwing exceptions). The user is allowed to leave at any time. Maybe it is better to log this issue and then return null as negotiation object and handle this properly in the Saros session manager.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/NegotiationFactory.java
```diff
@@ -211,4 +207,18 @@ public AbstractIncomingProjectNegotiation newIncomingProjectNegotiation(
...
+    }
+
+    String type = session.getUserProperties(user).getString(ProjectNegotiationTypeHook.KEY_TYPE);
+    if (type.isEmpty()) {
+      throw new IllegalArgumentException("Missing TransferType for User: " + user);
```
srossbach: See above, the exception is caught somewhere in the network layer.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/NegotiationFactory.java
stefaus: I would avoid returning null, as this is more of an unexpected behavior and the caller would need to check for it.

The correct handling in the Saros Session Manager is already difficult. For example this `return`.
https://github.com/saros-project/saros/blob/4a43f816677fdc1b92d6638ff29b3135f019d4a2/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/session/SarosSessionManager.java#L654-L657
In this case, the host user does not see an error message and the project is marked as shared, so `return` when the factory fails does not work properly.
Using an exception shows at least an error in the GUI, even if the project is marked as shared...

Since this exception handling is already broken (with IllegalArgumentException, UnsupportedOperationException) I would prefer to change the handling in `SarosSessionManager` in another PR.

stefaus:
rebased on current master
@srossbach please check my answer.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/NegotiationFactory.java
srossbach: The method to set the handler is expected to be called ONCE during the whole lifetime of the application. It really makes no sense to swap it during runtime.

stefaus: @srossbach Could we please get back to the fact that IllegalArgumentException and UnsupportedOperationException are already thrown at the factory calls. I would like to change the exception handling in another PR, can we agree on that?

srossbach: Because some of you added them. I wrote this class and I do not remember that I had thrown any exceptions. Link the follow up PR and I will approve this patch.

stefaus: @srossbach #403

srossbach:
The problem is that beside of logging the exception there is nothing you can do to handle them. So yes returning null is not that bad.

PR #380, opened 2019-01-14
Fix local editor status change handling

tobous:
#### [INTERNAL][I] Improve LocalEditorStatusChangeHandler.setEnabled(...)

Moves the subscribing and unsubscribing from the held IntelliJ listener
into the setEnabled(boolean) method. This unifies the functionality.
Furthermore, the previous solution that relied on the enabled flag could
cause consistency issues where events that should be ignored were
triggered while the flag was set to false but the listener was only
notified after the flag was set to true again. This is avoided by always
unsubscribing from the listener when disabling the handler.

#### [FIX][I] Fix editor selection change without open editor handling

Removes wrong nullcheck in
LocalEditorStatusChangeHandler.generateEditorActivatedActivity(...)
where null is a valid value.

Fix LocalEditorHandler.activateEditor(...) to also dispatch activities
when there is no editor selected locally as this is also a valid stat
that should be passed to other participants.

#### [FIX][I] Disable editor status handler when initializing editor pool

Disables the LocalEditorStatusChangeHandler while initializing the
editor pool at the start of a session. This is needed as
ProjectAPI.openEditor(...) still switches to the opened editor, even if
the focus flag is set to false. This editor switching would create false
editor activated activities that cause the user editor state of the
other participants to assume that the last editor added to the editor
pool is currently the active editor for the local user, which is not
necessarily the case.

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorStatusChangeHandler.java
+++ b/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorStatusChangeHandler.java
@@ -12,14 +12,12 @@ import de.fu_berlin.inf.dpp.filesystem.IFile;
 import de.fu_berlin.inf.dpp.intellij.editor.annotations.AnnotationManager;
 import de.fu_berlin.inf.dpp.intellij.filesystem.VirtualFileConverter;
 import de.fu_berlin.inf.dpp.intellij.session.SessionUtils;
-import org.apache.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 
 /** Dispatches activities for editor changes. */
 class LocalEditorStatusChangeHandler implements DisableableHandler {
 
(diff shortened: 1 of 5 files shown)


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorStatusChangeHandler.java
```diff
@@ -169,20 +155,28 @@ void subscribe(@NotNull Project project) {
...
-  void unsubscribe() {
+  private void unsubscribe() {
     messageBusConnection.disconnect();
 
     messageBusConnection = null;
   }
 
   /**
-   * Enables or disabled the handler. This is not done by disabling the underlying listener.
+   * Enables or disabled the handler.
```
m273d15: typo disabled->disables

tobous:
Fixed typo and slightly expanded javadoc.

tobous:
Rebased onto the current master without interaction.

PR #381, opened 2019-01-14
Update to Gradle 5.1

m273d15:
* Update Gradle version
* Use new feature that changes the dependency resolution:
The common gradle dependency resolution handling for each dependency is:
Try to fetch the dependency/artifact from the first repository. If the dependency
is not available in this repository then try the next respository. We use two
repositories which contain a single artifact we need. It is not necessary to check
these repositories for the other dependencies.
Therefore these repositories directly match to the specific group of the dependency.


m273d15:
I rebased this change and switched to the new [patched version 5.1.1](https://github.com/gradle/gradle/releases/tag/v5.1.1)

PR #382, opened 2019-01-14
Fix viewport adjustment logic

tobous:
To make the review process easier, the commits should be reviewed (or at least looked at) separately.

#### [FIX][I] Set correct line range in LocalViewPortChangeHandler
Adjusts the logic in LocalViewPortChangeHandler.getLineRange(...) to use
the IntelliJ API to calculate the lines from the given viewport
rectangle instead of calculating it by hand using the line height.

The new logic now uses the logical line positions, meaning it ignores
code folding/soft-wraps. This ensures that both participants see
approximately the same line range, independent of code folding.

#### [FIX][I] Fix basic viewport adjustment logic
Introduces a basic viewport adjustment logic used by the follow mode.
The logic currently always focuses on the center of the remote viewport
range if available or on the center of the remove selection range
otherwise.

There is currently no special handling for the case that the follower
viewport is smaller than the followee viewport.

#### [REFACTOR][I] Do minor cleanup
Adjusts javadoc to better describe the method responsibilities and
restrictions.

Replaces anonymous inner classes with lambda expressions. Subsequently
cleans up the body of the lambda expressions.

Reduces method visibility to minimal necessary level.

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorAPI.java
+++ b/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorAPI.java
@@ -9,10 +9,8 @@ import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.editor.LogicalPosition;
 import com.intellij.openapi.editor.ScrollType;
-import com.intellij.openapi.editor.VisualPosition;
 import com.intellij.openapi.project.Project;
 import de.fu_berlin.inf.dpp.editor.text.LineRange;
-import de.fu_berlin.inf.dpp.intellij.editor.colorstorage.ColorModel;
 import de.fu_berlin.inf.dpp.intellij.filesystem.Filesystem;
 import java.awt.Point;
 import java.awt.Rectangle;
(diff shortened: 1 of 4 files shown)


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorAPI.java
```diff
@@ -33,26 +35,43 @@ public EditorAPI(Project project) {
...
-          @Override
-          public void run() {
+  /**
+   * Returns the logical line range of the local viewport for the given editor.
+   *
+   * @param editor the editor to get the viewport line range for
+   * @return the logical line range of the local viewport for the given editor
+   * @see LogicalPosition
+   */
+  LineRange getLocalViewportRange(Editor editor) {
```
m273d15: You have the same logic as in the previous commit in file LocalViewPortChangeHandler.java. Is there an appropriate location where you could place this function as helper?

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/LocalEditorManipulator.java
```diff
@@ -188,24 +191,85 @@ public void applyTextOperations(SPath path, Operation operations) {
...
-        editor, range.getStartLine(), range.getStartLine() + range.getNumberOfLines());
+    int remoteViewPortCenter = findCenter(remoteStartLine, remoteEndLine);
+
+    editorAPI.scrollToViewPortCenter(editor, remoteViewPortCenter);
+  }
+
+  /**
+   * Returns the center file of the given line range.
+   *
+   * <p>IntelliJ sets the center position of an editor at 1/3 of thw visible line range. This is
```
m273d15: typo `thw`

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorAPI.java
tobous: This is the correct position as a "helper" since this class is only used to encapsulate the IntelliJ API access dealing with editor stuff. Will replace the other usage with a reference to this method.

tobous:
Fixed typo and used `EditorAPI.getLocalViewPortRange(Editor)` in `LocalViewPortChangeHandler`.

tobous:
Rebased onto current master. Fixed minor merge conflict with EditorManager constructor arguments. No logic had to be changed.

PR #384, opened 2019-01-15
Fix UserEditorState initialization

tobous:
Fixes UserEditorState initialization with `[FIX][I] Fix UserEditorState initialization`.

Does some cleanup with the other commits.

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorManager.java
+++ b/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorManager.java
@@ -6,6 +6,7 @@ import com.intellij.openapi.editor.Document;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.editor.event.SelectionEvent;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.util.Pair;
 import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VirtualFile;
 import de.fu_berlin.inf.dpp.activities.EditorActivity;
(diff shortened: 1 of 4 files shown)


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorManager.java
```diff
@@ -278,6 +247,73 @@ public void resourcesAdded(final IProject project) {
               .invokeAndWait(
                   () -> addProjectResources(project), ModalityState.defaultModalityState());
         }
+
+        /**
+         * Sends awareness information to populate the UserEditorState of the participant that
+         * joined the session.
+         *
+         * <p>This is done by first sending the needed state for all locally open editors and then
+         * sending an activity to the the correct editor as the active editor.
```
m273d15: Does this mean the information of the active editor are send twice or is the active editor not part of the editorPool?

tobous: The information that the editor is open is send twice for the active editor. The viewport and selection is not send a second time.
The reasoning behind this is that the UserEditorStateManager always sets the last editor that was reported as open/active as the active editor.
See https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/editor/remote/UserEditorState.java#L94

This could be avoided by either sorting the list of open editors to always handle the active editor last or by filtering for the active editor and then handle the complete state sending (including viewport and selection) separately.

I didn't think about it much when implementing this. Activating the locally active editor again in the end is still a relic of the previous implementation.
If you would prefer another option, I can change it.

m273d15: Ah, that makes sense. Could you add this info (last editor that was reported as open/active is the active editor) in your javadoc?

tobous: Tried to make it more clear what happens. I find it kind of hard to describe it concisely. Does the new javadoc convey all the necessary information?

m273d15: With the new javadoc it is possible to understand the main idea why you send the editor open information twice. I think all necessary information are included.

tobous:
Rebased onto current master. Solved minor merge conflicts in `EditorManager` CTOR arguments and in `addProjectResources(...)` (caused by the introduction of the try-catch block around the loop).

PR #385, opened 2019-01-16
Remove de.fu_berlin.inf.dpp prefix

m273d15:
The content of this PR is generated by a perl script in order to simplify to apply this refactoring to a new state of the master branch.

This PR changes:
* The directories in the repository root:
   * Renamed `de.fu_berlin.inf.dpp` -> `de.fu_berlin.inf.dpp.eclipse`
   * Removed the prefix `de.fu_berlin.inf.dpp.`
* Renamed osgi plugins and gradle projects:
   * Renamed `de.fu_berlin.inf.dpp` -> `de.fu_berlin.inf.dpp.eclipse`
   * Substituted prefix `de.fu_berlin.inf.dpp` with `saros` 
* Renamed the packages:
   * Substituted prefix `de.fu_berlin.inf.dpp` with `saros`

I want to clarify whether somebody has issues with these adjustments. Before this PR is merged I will talk with other developers which are currently working on bigger refactorings (in order to avoid that this PR destroys their work).


m273d15:
@tobous @srossbach What do you think about the refactoring steps? Do you have any issues with the names?

Of course I tested the PR via executing all (stf-)tests.

tobous:
It is really annoying that we don't own the URL "saros.org". According to the [java package naming rules](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html), the technically correct package name would be `org.saros_project`.

I don't like that there is an underscore in the name and the "project" does not contribute anything to the name, but I think it would still be better to follow the conventions.

What are your opinions?

srossbach:
IIRC you also wanted to introduce the common java maven/gradle ? layout, i.e adding main and test folders. Is this already included ? To be honest I am not a big fan of this beside renaming the eclipse package which is necessary.

tobous:
@srossbach It doesn't look like it is included in this PR (here is the [file structure](https://github.com/saros-project/saros/tree/pr/refactor/naming)).

Could you be more specific with what you don't like?
- The [gradle source structure](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html)
- The changes made in this PR (listed in more detail in the PR description)
- The package prefix I suggested (`org.saros_project`)

m273d15:
@srossbach This PR does not change the src-dir structure, because this is more or less a matter of taste and could lead to a longer discussion. Therefore I would like to change the src-dir structure in a follow-up PR (after we clarified which structure is desired).

@tobous Even if this prefix is not really nice, it complies with the java conventions...probably it is the best alternative. This would also become the prefix of the osgi bundles.

m273d15:
@tobous After talking with Lutz and thinking about the best package name, I think we should use the shortest name `saros` instead of using a convention that is focused on the avoidance of naming conflicts between multiple libraries (and we are not a library).

If nobody has issues with this PR then I will create a rebased version at Friday (the 1th of march), run all stf tests and merge the PR (if all tests are successful). If someone has merge conflicts I will provide my script (As attachment of this PR) that I used to perform the refactoring and help if any issue occur.

m273d15:
You will find my refactor script [here](https://github.com/saros-project/saros/files/2918648/refactor_script.zip).

m273d15:
I tested this PR:
* [X] All Unit tests are successful
* [X] All STF tests are successful
* [ ] Intellij Plugin is runnable in IntelliJ IDEA

PR #390, opened 2019-01-21
Fix FollowModeManager issues

tobous:
#### [FIX][CORE] Fix file closing ending follow mode
With the previous logic, either the follower or followee closing a file
would cause the follow mode to end. This commit adjusts the logic so
that the follower and followee can close editors as they please.
If the followee closes the active editor, the new active editor will
automatically be selected by the follower. If there is no active editor
for the followee after closing the previously active editor, the follow
mode is paused until a new editor is opened by the followee.

Subsequently inline stopIfDiverted as it is now only used by one method.

#### [FIX][CORE] Allow follower to close last editor while follow mode paused
Adjusts the logic in FollowModeManager to allow the follower to close
their last editor while the follow mode is paused, meaning the followee
currently has no open editors. If the followee opens an editor, the
editor is also opened for the follower, meaning the follow mode is
continued correctly irrespective of the local editor state of the
followee.

#### [CORE] Remove logic that automatically closes follower editors
Removes the logic from the FollowModeManager that closes editors on the
follower side when the corresponding editor is closed by the followee.
This allows the follower to have full control over which editors are
open during the follow mode.

Subsequently removes the STF tests that were checking for this behavior.

#### [FIX][CORE] Avoid duplicate notifications when follow mode ends
Changes the internal logic of the FollowModeManager that ends the
session based on received activities.

The previous handling used the follow(User) method to end the follow
mode and then notified the listener with the correct reason for the
follow mode end. The method follow(User) internally also notifies the
listener with the reason "FOLLOWER_STOPPED", meaning this incorrect
reason was dispatched first to the listeners. Furthermore, this meant
that the listener was called twice when the follow mode ended through
editor state changes.

The new logic now ends the follow mode in such cases by simply dropping
the held follow mode state without dispatching any notifications. This
means that it is the callers responsibility to notify the listener about
the follow mode end.

#### [FIX][CORE] Avoid unnecessary follow mode state change notifications
Avoids notifying the follow mode listeners on follow(User) calls that
don't change the state. Such calls are follow(...) on a user that is
currently already being followed or on null while there is no current
follow mode.

### DIFF
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/editor/FollowModeManager.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/editor/FollowModeManager.java
@@ -76,20 +76,26 @@ public class FollowModeManager implements Startable {
             return;
           }
 
-          stopIfDiverted(filePath, Reason.FOLLOWER_CLOSED_OR_SWITCHED_EDITOR);
+          Reason reason = Reason.FOLLOWER_CLOSED_OR_SWITCHED_EDITOR;
+
+          EditorState remoteActiveEditor = followeeEditor();
+
+          if (remoteActiveEditor != null && !remoteActiveEditor.getPath().equals(filePath)) {
+
+            follow(null);
+            notifyStopped(reason);
+          }
         }
 
         @Override
         public void editorClosed(User user, SPath filePath) {
           if (!user.equals(localUser) || !isFollowing()) return;
 
-          stopIfDiverted(filePath, Reason.FOLLOWER_CLOSED_EDITOR);
-        }
+          Reason reason = Reason.FOLLOWER_CLOSED_EDITOR;
 
-        private void stopIfDiverted(SPath filePath, Reason reason) {
           EditorState remoteActiveEditor = followeeEditor();
 
-          if (remoteActiveEditor != null && !remoteActiveEditor.getPath().equals(filePath)) {
+          if (remoteActiveEditor != null && remoteActiveEditor.getPath().equals(filePath)) {
 
             follow(null);
             notifyStopped(reason);
--- a/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/editor/FollowModeManager.java
+++ b/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/editor/FollowModeManager.java
@@ -121,11 +121,6 @@ public class FollowModeManager implements Startable {
               }
               break;
             case CLOSED:
-              // TODO This closes every editor the followee closes -- desired?
-              // Problem: In order to determine whether the followee closed
-              // his/her active editor and just an editor in the background,
-              // one needs to keep track of the last few EditorActivities.
-              editorManager.closeEditor(path);
               break;
             case SAVED:
               break;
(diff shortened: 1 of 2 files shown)


tobous:
@fzieris @srossbach As the second commit message states, this introduces a bug in the display of the follow mode state for the followee when the follower ends the follow mode in a certain way (follower has only the followee editor open; leaves the follow mode by closing editor).

As I am not familiar with the Saros/E GUI logic, could you point out to me where the logic displaying the follow mode state is located?

srossbach:
https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/model/session/FollowModeInformationTreeElement.java

https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/model/session/SessionContentProvider.java

https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/awareness/AwarenessInformationCollector.java

srossbach:
https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/model/session/UserElement.java#L100

tobous:
I investigated the issue some more. The display on the followee side is not the real issue but only a result of an issue on the follower side:

When the follower closes the active editor, they correctly leave the follow mode. But the follow mode button does not seem to be updated correctly (at least not always). If the button is clicked, the editor currently focused by the followee is also focused by the follower and the followee displays locally that the follow mode has started. But the follower does not  scroll or switch editors when the followee does, meaning they are not in the follow mode.

This issue can be avoided by explicitly selection "Follow user XYZ" from the drop down menu instead of just clicking the follow mode button on the followee side to start the follow mode.

As already hinted at in the description, this issue does not always occur. Sometimes it does seem to work, but more often then not, the wrong state is created.

I removed the entry from the commit message as I was also able to recreate the issue on the current master, meaning it was not introduced by my changes.

m273d15:
@tobous I will try to run the stf tests (on `master` and `pr/core/fix_followmode_close_editor`) today

tobous:
@srossbach @fzieris TL;DR: Currently, when the followee closes an editor during the follow mode, the corresponding editor is also closed on the follower side. Should this functionality remain or can it be removed?

<details><summary>For the more detailed explanation of the question, click here.</summary>
<p>

With the commit `[FIX][CORE] Fix file closing ending follow mode`, I removed the follow mode logic that automatically closes editors on the follower side when a corresponding editor is closed by the followee.

This behavior seemed weird to me as the follower might want to keep editors open to look at them later. Furthermore, with this PR, the follower now also has the ability to close editors during the follow mode (without influencing the followee or leaving the follow mode), meaning they can clean up their open editors themselves if desired.

But, as there are STF tests reliant on this behavior (the first and third test in [`AllParticipantsFollowUserWithWriteAccessTest`](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/test/stf/de/fu_berlin/inf/dpp/stf/test/permissions/AllParticipantsFollowUserWithWriteAccessTest.java)) and as I can also see the upside of the behavior, I wanted to ask for your opinion.

Should this behavior of the follow mode remain or should it be removed?

</p>
</summary>

m273d15:
@m273d15 I prefer the handling without closing the editors on the follower side. Otherwise the followee could close editors on the side of the follower which are consciously opened by the follower.

tobous:
Moved the removal of the logic to close follower editors to a separate commit. Remove STF tests that were checking the old behavior as part of that commit.

tobous:
@m273d15 Could you run the STF with this PR again?

m273d15:
@tobous All tests are successful

tobous:

PR #393, opened 2019-01-23
[BUILD] Make intellij sandbox dir configurable

m273d15:
* Introduced optional environment variable SAROS_INTELLIJ_SANDBOX
  which specifies the parent directory of the sandbox dirs
  e.g. if SAROS_INTELLIJ_SANDBOX=/tmp and two sandboxes
  are simultaneously the directories /tmp/idea-sandbox{1,2} are
  created


m273d15:
Thank you for the reminder :+1:

PR #395, opened 2019-01-24
[BUILD] Switch to google java format 1.7

### LINKED ISSUE #389: Update google java formatter
We have to adopt the [new 1.7 version](https://github.com/google/google-java-format/releases) of the google java formatter (which is used for format checking in travis).

tobous:
Updates the travis script to use the google java format 1.7.

Subsequently runs the formatter (through the IntelliJ plugin) to adjust
to the new formatting rules.

Resolves #389.

This PR will have to wait until there is an official version of the eclipse plugin.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
         fetch-depth: 1
     - uses: m273d15/java-format-action@v1.1
       with:
-        version: '1.6'
+        version: '1.7'
 
   build:
     runs-on: ubuntu-latest
--- a/buildSrc/src/main/java/saros/gradle/eclipse/configurator/OsgiBundleVersionConfigurator.java
+++ b/buildSrc/src/main/java/saros/gradle/eclipse/configurator/OsgiBundleVersionConfigurator.java
@@ -38,8 +38,7 @@ public class OsgiBundleVersionConfigurator {
       Charset charset = StandardCharsets.UTF_8;
       List<String> oldContentLines = Files.readAllLines(path, charset);
       String newContent =
-          oldContentLines
-              .stream()
+          oldContentLines.stream()
               .map(line -> line.replaceAll(regex, replacement))
               .collect(Collectors.joining("\n"));
       Files.write(path, newContent.getBytes(charset));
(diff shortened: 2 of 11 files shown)


m273d15:
LGTM

tobous:
Rebased onto current master. Fixed some conflicts dealing with imports due to the recent refactoring of the package structure.

tobous:
Rebased onto current master without any interaction. Checked whether new code needs reformatting (nothing found).

tobous:
To better bridge the gap while waiting on a Eclipse release of the google java formatter 1.7, I build the 1.6 plugin for IntelliJ removing the upwards compatibility flag. This allows the plugin to run on IntelliJ versions later than 2018.3. Here is the bundled plugin if anybody is interested:
[google-java-format-1.6.zip](https://github.com/saros-project/saros/files/3433515/google-java-format-1.6.zip)

tobous:
As they now added an update to still work with versions after IntelliJ IDEA 2019.3, I updated the build. The changes I made are now available on [my fork of the google-java-format repo](https://github.com/tobous/google-java-format).

Here is the plugin zip (build on the branch [`release-1.6`](https://github.com/tobous/google-java-format/tree/release-1.6)) for ease of use: [google-java-format.zip](https://github.com/saros-project/saros/files/3832321/google-java-format.zip)

tobous:
Rebased onto current master and re-ran 1.7 formatter.

@m273d15 You mentioned that you had worked on a PR for the google java format eclipse plugin build. Any updated on that?

m273d15:
I think they don't support eclipse anymore. After your reminder, I wanted to create a PR, but they set JDK 11 as current minimum JDK which is not supported by eclipse. I think we should think about another formatter.

tobous:
> I think they don't support eclipse anymore. After your reminder, I wanted to create a PR, but they set JDK 11 as current minimum JDK which is not supported by eclipse. I think we should think about another formatter.

I mean we could, but I am not sure whether the issue is pressing enough to jump ship (for now). The current formatting works and changing the formatter again causes a lot of difficulties for existing PRs and a lot of "unnecessary" code changes.

At some point, Eclipse will also have to update to building with Java 11 as I can't image the Eclipse Foundation just giving up on the project. (But this might still take a while.)

But I am not actually sure whether requiring Java 11 for a plugin is a problem. If the IDE is running on Java 11, this shouldn't be an issue, should it?

For Saros, this would add a new requirement for the user which we might not want to enforce as a lot of enterprise users don't have control over their JDK setup.
But for Saros development, requiring Eclipse to run on Java 11 (even though we are still building again Java 8) would be an acceptable restriction in my opinion.

Is your approach still compatible with the new Java 11 build of the google java formatter?

PR #396, opened 2019-01-24
[FEATURE][S] Support incoming ProjectNegotiations 

Drakulix:
Add support for the server to handle `IncomingProjectNegotiation`s now that #355 has finally landed.

Better check commit separately, first commit adds the new functionality.
Second commit is only cleanup of code using newer Java-Features (lambdas, diamond operator).

There is no code to handle ProjectID conflicts in place, because otherwise sharing projects on a later run would not use files already there, and re-sharing new files, when doing partial shares, would not work as well.

This is an issue for later work on the Server, as it will require some more configuration options.
This is for a first alpha release, which should only be run in trusted environments anyway.

### DIFF
--- a/server/src/saros/server/session/NegotiationHandler.java
+++ b/server/src/saros/server/session/NegotiationHandler.java
@@ -1,11 +1,14 @@
 package saros.server.session;
 
+import java.io.IOException;
 import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import org.apache.log4j.Logger;
 import saros.filesystem.IProject;
+import saros.filesystem.IWorkspace;
 import saros.monitoring.NullProgressMonitor;
 import saros.negotiation.AbstractIncomingProjectNegotiation;
 import saros.negotiation.AbstractOutgoingProjectNegotiation;
@@ -13,8 +16,11 @@ import saros.negotiation.IncomingSessionNegotiation;
 import saros.negotiation.NegotiationTools;
 import saros.negotiation.OutgoingSessionNegotiation;
 import saros.negotiation.ProjectNegotiation;
+import saros.negotiation.ProjectNegotiationData;
 import saros.negotiation.SessionNegotiation;
+import saros.net.xmpp.JID;
 import saros.server.ServerConfig;
+import saros.server.filesystem.ServerProjectImpl;
 import saros.server.progress.ConsoleProgressIndicator;
 import saros.session.INegotiationHandler;
 import saros.session.ISarosSessionManager;
@@ -25,6 +31,7 @@ public class NegotiationHandler implements INegotiationHandler {
   private static final Logger log = Logger.getLogger(NegotiationHandler.class);
 
   private final ISarosSessionManager sessionManager;
+  private final IWorkspace workspace;
   private final ThreadPoolExecutor sessionExecutor =
       new ThreadPoolExecutor(
           0,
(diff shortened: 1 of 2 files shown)


de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/NegotiationHandler.java
```diff
@@ -25,65 +31,64 @@
...
-              case CANCEL:
-                log.info("Session negotiation was cancelled locally");
-                break;
-              case REMOTE_CANCEL:
-                log.info(
-                    "Session negotiation was cancelled by remote: "
-                        + negotiation.getPeer().toString());
-                break;
-            }
+        () -> {
```
srossbach: Is this just a refactor ? If this is just a refactor ... put this stuff into a method ! And not into a lambda

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/session/NegotiationHandler.java
```diff
@@ -92,14 +97,10 @@ public void run() {
...
     sessionExecutor.execute(
-        new Runnable() {
-          @Override
-          public void run() {
-            negotiation.localCancel(
-                "Server does not accept session invites",
-                NegotiationTools.CancelOption.NOTIFY_PEER);
-            negotiation.accept(new NullProgressMonitor());
-          }
+        () -> {
```
srossbach: Refactor ...

de.fu_berlin.inf.dpp.server/src/de/fu_berlin/inf/dpp/server/filesystem/ServerProjectImpl.java
```diff
@@ -75,4 +77,8 @@ public IFolder getFolder(String pathString) {
   private IPath getFullMemberPath(IPath memberPath) {
     return getFullPath().append(memberPath);
   }
+
+  public void create() throws IOException {
```
srossbach: DOC

Drakulix:
@srossbach That is why I split this up into two commits. But anyway its gone now, all good?

Drakulix:
> You mention in the PR message that this logic currently does not handle ProjectID conflicts. Could you explain what you mean by this?
> Could you also add a note/todo to the appropriate section in the code explaining this for future reference?

I have added a TODO. The problem is currently, that you cannot set a name for the folder like in an interactive environment (e.g. Saros/E) for the server. So projects with the same name will get placed in the folder, resulting in conflicts and broken environments.

However this makes it possible to restart a server-session without re-transmitting the whole project. My thesis elaborates on this a lot more, basically this situation was considered okay for the alpha state the server is in. A real solution would also need to solve the problem, how the server cleans-up after itself, which is somewhat related to this problem.

server/src/saros/server/session/NegotiationHandler.java
```diff
@@ -146,15 +136,62 @@ public void run() {
...
+                + errorMessage
+                + " at remote: "
+                + peer.toString());
+        break;
+      case CANCEL:
+        log.info("Project negotiation was cancelled locally");
+        break;
+      case REMOTE_CANCEL:
+        log.info("Project negotiation was cancelled by remote: " + peer.toString());
+        break;
```
tobous: As a suggestion (as this was also noted by Codacy): You could add a default case, either to throw an exception if there is an unhandled case (which would help us notice when an error type got added but this logic was not updated) or to add a catch-all logging statement (could serve the same purpose).

server/src/saros/server/session/NegotiationHandler.java
```diff
@@ -146,15 +136,62 @@ public void run() {
...
+
+      // TODO: The file path is currently dictated by the name, potentially resulting in CONFLICTS
+      if (!project.exists()) {
+        try {
+          project.adaptTo(ServerProjectImpl.class).create();
+        } catch (IOException e) {
+          projectMapping.clear();
+          negotiation.localCancel(
+              "Error creating project folder", NegotiationTools.CancelOption.NOTIFY_PEER);
+          break;
```
tobous: Is a break really the right statement in this case? Do you want to still add the project to the mapping and continue with starting the session negotiation, even though the project creation failed (and you canceled the negotiation)?

tobous: Sorry, misread where the loop ended. So the part of the project mapping does not apply. But I am still not sure whether it is sensible to call the `projectExecutor` to run the negotiation, even though you already canceled it locally.

Drakulix: Yeah, you are right

Issue #223, opened 2018-08-29
Vector time does not get reset correctly when deleting a file


tobous:
Saros does not seem to reset the vector time correctly when a file is deleted.
This leads to issues when a file is deleted and a new file with the same name is subsequently created. When this new file is edited by one of the participants, the vector time between the clients de-synchronizes, resulting in errors on the host side and edits for the created file not being shared.

```
ERROR 14:47:00,826 [DispatchContext] (ConcurrentDocumentServer.java:197) failed to transform checksum activity: ChecksumActivity(path: SPath [project=P/alp (EclipseProjectImpl), projectRelativePath=src/test/Created.java], hash: 743766400, length: 41, jupiterTimestamp: [4,2])
de.fu_berlin.inf.dpp.concurrent.jupiter.TransformationException: precondition #2 violated (Remote vector time (2) is greater than local vector time (0)).
	at de.fu_berlin.inf.dpp.concurrent.jupiter.internal.Jupiter.isCurrent(Jupiter.java:147)
	at de.fu_berlin.inf.dpp.concurrent.jupiter.internal.JupiterDocumentServer.withTimestamp(JupiterDocumentServer.java:131)
	at de.fu_berlin.inf.dpp.concurrent.management.JupiterServer.withTimestamp(JupiterServer.java:114)
	at de.fu_berlin.inf.dpp.concurrent.management.ConcurrentDocumentServer.withTimestamp(ConcurrentDocumentServer.java:195)
	at de.fu_berlin.inf.dpp.concurrent.management.ConcurrentDocumentServer.transformIncoming(ConcurrentDocumentServer.java:128)
	at de.fu_berlin.inf.dpp.session.internal.ActivityHandler.directServerActivities(ActivityHandler.java:422)
	at de.fu_berlin.inf.dpp.session.internal.ActivityHandler.handleIncomingActivities(ActivityHandler.java:144)
	at de.fu_berlin.inf.dpp.session.internal.SarosSession.exec(SarosSession.java:734)
	at de.fu_berlin.inf.dpp.session.internal.ActivitySequencer$3.run(ActivitySequencer.java:379)
	at de.fu_berlin.inf.dpp.util.ThreadUtils$1.run(ThreadUtils.java:38)
```
```
[...]

PR #397, opened 2019-01-24
[FIX][CORE] Add Jupiter-Heartbeats via TimestampOperations

Drakulix:
Alright, this is a difficult one.

The reason for this PR goes back the master thesis of @denisw and I do not think we actually have an issue for this. So let me just quote it (german):

> In der Saros-Implementation des Jupiter-Algorithmus können Operationen ausschließlich dadurch bestätigt werden, dass der Empfänger seinerseits eine neue Operation in Form einer `JupiterActivity` versendet. Darin gespeichert ist die Zahl der bisher empfangenen und angewendetenOperationen der Gegenseite (Remote Operation Count); diese kann so darauf schließen, welche Operationen nicht mehr zwischengespeichert werden müssen. Diese Art der Bestätigung ist ausreichend, wenn alle Sitzungs-teilnehmer regelmäßig Bearbeitungsoperationen durchführen; sobald dasein Teilnehmer aber nicht tut, beispielsweise weil er nur als Gutachter an einer Sitzung teilnimmt, werden die an ihn gesendeten Operationen nie bestätigt und der Sender (sprich, der Host in seiner Rolle als Jupiter-Server) kann sie folglich auch nie löschen, sodass sie sich immer weiter ansammeln.

I have actually verified this behavior using VisualVM (don't have the heap dumps right now, because I forgot to save them, I will add some later, but the screenshot should be self-explanatory):

visual_vm_bad3

As you can see over the course of a good hour, I was able to create a heapset of roughly 20MB of `OperationWrapper`s containing `InsertOperation`s and `DeleteOperation`s with a crafted STF test, where three clients, where bombarding a fourth one with inserts and deletions.

So you might actually argue: "Why do we care about this?"
- This took a long time
- 20Mb is not that bad on a modern system
- The use-case is not really realistic
- This was with spamming clients, no one actually makes edits worth of 20Mb in that time
- Saros should be executed in a trusted environment anyway

This is all absolutely true, but that fourth client, that was **Saros/S**. So this use-case is not unrealistic anymore as the server never does any edits itself (but still is a participating client in the session) and is expected to run for a long time (weeks, months).

So how do we solve this. I came up with three ideas:

- `Ack` every `Activity`
- Send a frequent heartbeat
- Send a heartbeat if no message was send in [amount of time]

Acknowledging every `Activity` would obviously increase the network load by a substantial amount.
Sending a conditional heartbeat is actually difficult to track, since the Jupiter-Algorithm runs for every open editor. We don't want to start the thread per algorithm-instance and I do not think handling a Map of `wasEdited` booleans on a upper layer is a particularly elegant solution.  

I think the second solution strikes a nice middle-ground with relatively small changes to a core-component of Saros and not a huge increase of network load as the amount of time between heartbeats can be quite large (I have right now set it to a minute).

I have re-used the `TimstampOperation` to implement this as it seemed it was supposed to do something similar, but was actually never put to use in the current code.

Here is a profiling run with my patch:
(Some Operations are still expected as I was hitting the server with them constantly, but the amount did not increase over the course of an hour.)

visual_vm_good2


de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/management/JupiterClient.java
```diff
@@ -7,34 +7,68 @@
...
-    this.sarosSession = sarosSession;
-  }
-
+  private static Logger log = Logger.getLogger(JupiterClient.class);
   /**
    * Jupiter instances for each local editor.
    *
    * @host and @client
    */
   protected final HashMap<SPath, Jupiter> clientDocs = new HashMap<SPath, Jupiter>();
```
stefaus: instead of locking, change this to a ConcurrentHashMap, which will handle this for you
additional this should be private

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/management/JupiterClient.java
```diff
@@ -7,34 +7,68 @@
...
 import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.apache.log4j.Logger;
 
 /** A JupiterClient manages Jupiter client docs for a single user with several paths */
-public class JupiterClient {
+public class JupiterClient extends AbstractActivityProducer {
 
   protected ISarosSession sarosSession;
```
stefaus: this should be private final
once you change it to final you have to move the `jupiterHeartbeat` creation to the CTOR because the access problem to `sarosSession.getLocalUser()` in Line 50 will become clear

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/management/JupiterClient.java
```diff
@@ -7,34 +7,68 @@
...
    */
   protected final HashMap<SPath, Jupiter> clientDocs = new HashMap<SPath, Jupiter>();
 
+  private final AtomicBoolean stopHeartbeatThread = new AtomicBoolean(false);
+  private final Thread jupiterHeartbeat =
+      ThreadUtils.runSafeAsync(
+          "dpp-jupiter-heartbeat",
+          log,
+          () -> {
+            while (!stopHeartbeatThread.get()) {
```
stefaus: according to "Java Concurrency in Practice - 7.1.1 Interruption" think about using interruption for cancellation 
`while (!Thread.currentThread().isInterrupted())`

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/management/JupiterClient.java
```diff
@@ -7,34 +7,68 @@
...
+              try {
+                Thread.sleep(60 * 1000);
```
stefaus: maybe it is advisable to use a `ScheduledThreadPoolExecutor` for this.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/management/JupiterClient.java
```diff
@@ -7,34 +7,49 @@
...
   /** @host and @client */
   protected synchronized Jupiter get(SPath path) {
-
-    Jupiter clientDoc = this.clientDocs.get(path);
-    if (clientDoc == null) {
-      clientDoc = new Jupiter(true);
-      this.clientDocs.put(path, clientDoc);
-    }
-    return clientDoc;
+    return clientDocs.computeIfAbsent(path, (key) -> new Jupiter(true));
```
srossbach: This is a refactor. Should not be part of this patch.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/management/JupiterClient.java
```diff
@@ -7,34 +7,49 @@
...
    */
-  protected final HashMap<SPath, Jupiter> clientDocs = new HashMap<SPath, Jupiter>();
+  private final Map<SPath, Jupiter> clientDocs = new ConcurrentHashMap<>();
+
+  private final ScheduledThreadPoolExecutor jupiterHeartbeat =
+      new ScheduledThreadPoolExecutor(1, new NamedThreadFactory("dpp-jupiter-heartbeat"));
+
+  public JupiterClient(ISarosSession sarosSession) {
+    this.sarosSession = sarosSession;
+    jupiterHeartbeat.scheduleWithFixedDelay(
```
srossbach: Beside this lambda madness I do not understand the reason to put the logic into this class. First the server/host also have a running instance of this class. Second, do we have to use TimestampOperations or is a NOP operation sufficient ?

As for the thread access. While you are allowed to call fireActivity from any thread you cannot access Jupiter client stuff from any thread. This has to be done in the UI / Application thread through the UISynchronizer. I do not prefer a solution with monitors or other concurrent data types. It only complicates the code. There are already enough synchronized stuff in here that should not exists.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/management/ConcurrentDocumentClient.java
```diff
@@ -188,4 +189,9 @@ public boolean isCurrent(ChecksumActivity checksumActivity) {
       return false;
     }
   }
+
+  @PreDestroy
```
srossbach: Annotations are metadata. This is never called. We do not use Java EE.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/management/JupiterClient.java
Drakulix: It is not really. The `ConcurrentHashMap` does not solve the locking problem, if I cannot atomically insert and receive the value. Which is why I swapped it to `computeIfAbsent`, which does exactly that.

srossbach: Every method in this class is synchronized. Using a ConcurrentHashMap does not gain you anything.

Drakulix: Okay fair enough. I guess I can rely on synchronizing the whole object instead and maybe try to refactor this later to use a `ConcurrentHashMap` instead.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/jupiter/internal/Jupiter.java
```diff
@@ -142,22 +144,17 @@ public boolean isCurrent(Timestamp timestamp) throws TransformationException {
   public Operation receiveJupiterActivity(JupiterActivity jupiterActivity)
       throws TransformationException {
 
-    if (jupiterActivity.getOperation() instanceof TimestampOperation) {
```
srossbach: This got added with commit https://github.com/saros-project/saros/commit/3276dc90c76f0afe356dbbf5f86f8c9cb28c9d9d

You just removed it. Problem is the following. Almost nobody knows how this Jupiter stuff really works and so it is not the best decision to make modifications to files that where not touched since almost 10 years.

Drakulix: As you can see [here](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/jupiter/internal/Jupiter.java#L147) and more importantly [here](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/jupiter/internal/text/TimestampOperation.java#L14) or alternatively by just doing `grep -r . TimestampOperation` TimestampOperations are **never** used in the whole codebase.

They were most likely intended to serve a similar purpose, but it was never implemented. So alternatively I could just add a new Operation and afterwards make a cleanup PR that removes the unneeded TimestampOperations, but I do not see the point in that.

> it is not the best decision to make modifications to files that where not touched since almost 10 years.

Yeah it is scary, which is why I spent days testing this. But not touching anything would mean the memory leak never gets fixed and that cannot possibly be your opinion on the subject, right?

Drakulix:
Here are the promised heapdumps for everyone to examine:

[heapdumps.zip](https://github.com/saros-project/saros/files/2800234/heapdumps.zip)

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/management/JupiterClient.java
```diff
@@ -71,4 +78,30 @@ public synchronized ChecksumActivity withTimestamp(ChecksumActivity checksumActi
...
     return get(checksumActivity.getPath()).withTimestamp(checksumActivity);
   }
+
+  @Override
+  public void start() {
+    jupiterHeartbeat =
+        new ScheduledThreadPoolExecutor(1, new NamedThreadFactory("dpp-jupiter-heartbeat"));
+    jupiterHeartbeat.scheduleWithFixedDelay(
+        () -> {
+          synchronized (JupiterClient.this) {
```
tobous: This gets a bit hard to read. I would consider moving the synchronized block to a separate method (e.g. `dispatchHeartbeats`) to improve the readability of this method. Then you could also replace the synchronized block by making the method synchronized.

This would also allow you to use a method reference instead of a lambda expression, which would  mean that all of this would fit onto one line.

I had have something like this in mind:

```Java
  @Override
  public void start() {
    jupiterHeartbeat =
        new ScheduledThreadPoolExecutor(1, new NamedThreadFactory("dpp-jupiter-heartbeat"));

    jupiterHeartbeat.scheduleWithFixedDelay(this::dispatchHeartbeats, 1, 1, TimeUnit.MINUTES);
  }

  /** Dispatches an activity for every open document functioning as a heartbeat. */
  private synchronized void dispatchHeartbeats() {
    clientDocs.forEach(
        (path, jupiter) -> {
          JupiterActivity heartbeat =
              jupiter.generateJupiterActivity(
                  new TimestampOperation(), sarosSession.getLocalUser(), path);
          fireActivity(heartbeat);
        });
  }
```

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/concurrent/management/JupiterClient.java
```diff
@@ -71,4 +78,30 @@ public synchronized ChecksumActivity withTimestamp(ChecksumActivity checksumActi
 
     return get(checksumActivity.getPath()).withTimestamp(checksumActivity);
   }
+
+  @Override
+  public void start() {
+    jupiterHeartbeat =
```
tobous: I am not sure how `ScheduledThreadPoolExecutor` is used correctly.
Is it necessary to create a new instance every time start is called?
If the same object could be re-used, would there a reason not to instantiate the executor in the constructor and just to re-schedule the periodic execution here?

Drakulix: From the docs of the `shutdown` method (called in `stop` to "un"-schedule the heartbeat):
> Initiates an orderly shutdown in which previously submitted tasks are executed, but no new tasks will be accepted.

The `getContinueExistingPeriodicTasksAfterShutdownPolicy`-method does suggest, that this also stops repeating tasks:
> Gets the policy on whether to continue executing existing periodic tasks even when this executor has been shutdown. In this case, these tasks will only terminate upon shutdownNow or after setting the policy to false when already shutdown. This value is by default **false**.

`scheduleWithFixedDelay` does return a `ScheduledFuture`, which can be cancelled, but I am not sure, how that works in the context of a fixed rate (the other `schedule`-methods also return such futures). Can this future resolve multiple times or does it never fully resolve? Does it only cancel the current/next tick or does it cancel all of them?

The documentation is also not particular clear on these questions.
E.g. is cancelling the future the same as cancelling the executor? (earlier they only hinted at **shutting down the executor** is this the same?) (also this is the only reference to cancel in the method description):
> If any execution of the task encounters an exception, subsequent executions are suppressed. Otherwise, the task will only terminate via **cancellation** or termination of the **executor**.

Or how nothing is said about cancellation regarding the return value (the future):
> a ScheduledFuture representing pending completion of the task, and whose get() method will throw an exception upon cancellation.

To me using `shutdown` seemed easier then to test the specific nature of this beast.

Drakulix: @stefaus mentioned, that he probably has some knowledge regarding this discussion point. I would love to make this kinda scary patch simpler.

stefaus: I think the solution now is good.
At some time you need to stop the Thread Pool or the JVM will wait forever.
A cancel on the Future or remove wouldn't stop the Thread Pool.

And if you get a Exception in `dispatchHeartbeats` (which will now **silently** cancel all following schedules..., yes no stacktrace) this will be running in the next Session again.
Mentioning this, you should wrap the code of `dispatchHeartbeats` in a try-catch and write exceptions to the log.

core/src/saros/concurrent/jupiter/internal/Jupiter.java
```diff
@@ -142,22 +144,17 @@ public boolean isCurrent(Timestamp timestamp) throws TransformationException {
   public Operation receiveJupiterActivity(JupiterActivity jupiterActivity)
       throws TransformationException {
 
-    if (jupiterActivity.getOperation() instanceof TimestampOperation) {
-
-      // TODO Use timestamps correctly!
-      log.warn("Timestamp operations are not tested at the moment");
-
-      updateVectorTime(jupiterActivity.getTimestamp());
```
tobous: This was the last usage of the method `updateVectorTime`. Sadly, as the mehtod does not have any meaningful javadoc (and the commit message when it was added is not helpful either), I am not sure what exactly it was meant for (why it was added). It seems like it was meant to be used together with the TimestampOperation to remotely update the vector-time of other participants. But why this was added is not clear to me.

I am somewhat torn on this, but I would prefer not to do half measures here. In my opinion, we should either
- completely remove the update stuff (as it is not used anywhere) if you want to reuse the TimestampOperation for heartbeats (I still think the name could be misleading) or
- create a new activity/operation for the heartbeat and leave the TimestampOperation stuff untouched (for now).

I am currently leaning towards the second option (implementing a new operation with a more fitting name) as this gives us the opportunity to evaluate the update logic more thoroughly without blocking this PR. Furthermore, I am not sure whether the functionality to update remote timestamps could still be useful for solving #223.

core/src/saros/concurrent/management/JupiterClient.java
```diff
@@ -71,4 +78,26 @@ public synchronized ChecksumActivity withTimestamp(ChecksumActivity checksumActi
...
   }
+
+  @Override
+  public void start() {
+    jupiterHeartbeat =
+        new ScheduledThreadPoolExecutor(1, new NamedThreadFactory("dpp-jupiter-heartbeat"));
+    jupiterHeartbeat.scheduleWithFixedDelay(this::dispatchHeartbeats, 1, 1, TimeUnit.MINUTES);
+  }
+
+  private synchronized void dispatchHeartbeats() {
```
tobous: Please add a short javadoc explaining what the function does and why heartbeats are necessary.

core/src/saros/concurrent/jupiter/internal/Jupiter.java
srossbach: Things I would like to see:

A dedicated class / component that can be turned on/off and not some thread hidden inside the client class doing "magic heartbeats".

The heartbeat firing should be done in the EDT. It is possible in this setup that you will fire a heartbeat for a non existing file (Eclipse filesystem is concurrent).

Last but not least what is wrong with generating NOPs ? They should also do the trick. Of course this ain't not so easy because you have to fire special TextActivities that will be translated to NOPs.

Drakulix: @tobous Because `TimestampOperation` was not used anywhere meaningful, I just went with it. But sure, I can add another operation, if you want to keep the code for #223. I will add a note on its current state.

@srossbach 

> Last but not least what is wrong with generating NOPs ?

The reason I need a new operation type and not a NOP is [this line](https://github.com/saros-project/saros/pull/397/files#diff-8d4e1fd03035f8adc5945d18e4a6f68aR85). NOPs would get added to the `ackJupiterActivityList`, which is absolutely unnecessary for the heartbeat and might cause more harm then this small special case does.

> A dedicated class / component that can be turned on/off

Sure I can move this into a separate class (and make it togglable). Are you fine with spawning the instance in the `JupiterClient`, because it needs access to the `clientDocs`?

> .. EDT ...
What is that?

> (Eclipse filesystem is concurrent)
In that case the `JupiterClient` will be out-of-sync anyway, which does not seem to be a problem currently, so is there something I need to do about this? This might actually be of use because I am not sure if the stored operations would be cleared if a file is closed. So a heartbeat for a missing file might actually be desirable.

core/src/saros/concurrent/management/JupiterClient.java
```diff
@@ -71,4 +78,26 @@ public synchronized ChecksumActivity withTimestamp(ChecksumActivity checksumActi
...
+    clientDocs.forEach(
+        (path, jupiter) -> {
+          JupiterActivity heartbeat =
```
srossbach: The weird part here is that you fire a JupiterActivity which then gets passed to some Jupiter component again before it is transmitted to every user.

core/src/saros/concurrent/jupiter/internal/Jupiter.java
```diff
@@ -82,8 +82,10 @@ public JupiterActivity generateJupiterActivity(Operation op, User source, SPath
     JupiterActivity jupiterActivity = new JupiterActivity(this.vectorTime, op, source, editor);
 
     // add(op, myMsgs) to outgoing;
-    this.ackJupiterActivityList.add(
-        new OperationWrapper(op, this.vectorTime.getLocalOperationCount()));
+    if (!(op instanceof TimestampOperation)) {
```
srossbach: So you pointed me to this line and what is the problem here ? The problem is that discardAcknowledgedOperations is never called. This was never done in this method anyway.

core/src/saros/concurrent/jupiter/internal/Jupiter.java
srossbach: EDT = Event Dispatch Thread aka the Main / GUI thread of an application.

tobous: To execute code on the IDE specific EDT, we have the core interface [UISynchronizer](https://github.com/saros-project/saros/blob/master/core/src/saros/synchronize/UISynchronizer.java).

(Somewhat off-topic: @srossbach I just noticed that we still have an old implementation for an AWT synchronizer located in the eclipse code that is never used: [AWTSynchronizer](https://github.com/saros-project/saros/blob/master/eclipse/src/saros/synchronize/internal/AWTSynchronizer.java). The commit message suggests that it was added for compatibility's sake. I don't think it is needed anymore as this is now handled separately by the IDE implementations. Am I correct in this assumption?)

core/src/saros/concurrent/jupiter/internal/Jupiter.java
Drakulix: Yeah, you are right `generateJupiterActivity` is not the problem (this is just a rather unnecessary small optimization), but instead, I wanted to point you to `receiveJupiterActivity`. But that is actually a non-issue either, sorry, it has been a while since I was working on this PR.

Drakulix:
Alright, I am not very keen on testing this again, until everyone is happy with the approach, which is why I added a "WIP:" tag to not merge this until I have (again) verified, that the leak is gone.

New approach:
- @srossbach You are right, I am using a `NoOperation` now.
- @srossbach The heartbeat-functionality is now in a separate class.
- @srossbach Heartbeat generation/sending is done in the UI-Thread.
- @tobous The old `TimestampOperation` code is untouched.
- I am using a `ConcurrentHashMap` again to synchronize access to the `Jupiter`-objects with the `JupiterClient`

srossbach:
Mixing java.util.concurrent and synchronized keyword is bad style.

Now 3 classes have a startable interface. Why ? When I said a dedicated class I meant that this class can be put into the session context container to enable heartbeats but not this start / stop chaining.

Drakulix:
> Mixing java.util.concurrent and synchronized keyword is bad style.
 
I know, but I did not want to refactor the `JupiterClient` to remove the use of synchronized there (which I would prefer), because this makes this harder to review and introduces even more threading-related changes in a critical part of the infrastructure.

The HashMap, on the other hand, is accessed from `JupiterClient` the `Heartbeat`-Class and used in the UI-Thread. It makes perfect sense to use the concurrent implementation here imo.

Do you disagree with any of that? If yes, which one would you rather drop / whats your preferred solution for this?

> Now 3 classes have a startable interface. Why ? When I said a dedicated class I meant that this class can be put into the session context container to enable heartbeats but not this start / stop chaining.

Yes, and how would I access the `clientDocs` then? The `JupiterClient` is not part of the container, but of the `ConcurrentDocumentClient` (CDC from here on). So should I instead hack-up a weird interface to access the private variable of a private variable of the CDC, which I then add to the `Heartbeat`-class which has nothing to do with the CDC from a conceptional point of view?

I think that is bad design as well, the heartbeat is part of a JupiterClient from a conceptual and from an implementation standpoint because both classes are very much coupled.

I agree, that the new `Startable`-implementations do not look very nice, but I have laid out, why I do not think the Heatbeat belongs in the container. at least not with the current class-structure, which I do not wanna touch, because I want to touch as less as possible of that Jupiter stuff (another requirement for the PR btw).

We would not have gained much from adding it to the container functionality-wise anyway. 

Your main requirement is done now. It is all in a separate class, which you can quickly comment out / revert in case something breaks.

core/src/saros/concurrent/management/Heartbeat.java
```diff
@@ -0,0 +1,51 @@
...
+import saros.activities.SPath;
+import saros.concurrent.jupiter.internal.Jupiter;
+import saros.concurrent.jupiter.internal.text.NoOperation;
+import saros.repackaged.picocontainer.Startable;
+import saros.session.AbstractActivityProducer;
+import saros.session.ISarosSession;
+import saros.synchronize.UISynchronizer;
+import saros.util.NamedThreadFactory;
+
+public class Heartbeat extends AbstractActivityProducer implements Startable {
```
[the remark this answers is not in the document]
Drakulix: done

core/src/saros/concurrent/management/Heartbeat.java
```diff
@@ -0,0 +1,51 @@
...
+
+  Heartbeat(
+      ISarosSession sarosSession, UISynchronizer uiSynchronizer, Map<SPath, Jupiter> clientDocs) {
+    this.sarosSession = sarosSession;
+    this.uiSynchronizer = uiSynchronizer;
+    this.clientDocs = clientDocs;
+  }
+
+  public void start() {
+    jupiterHeartbeat =
```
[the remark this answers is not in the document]
Drakulix: done

core/src/saros/concurrent/management/Heartbeat.java
```diff
@@ -0,0 +1,51 @@
...
+        new ScheduledThreadPoolExecutor(1, new NamedThreadFactory("dpp-jupiter-heartbeat"));
```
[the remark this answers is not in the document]
Drakulix: I have changed it to follow the style of other uses of NamedThreadFactory in core.

core/src/saros/concurrent/management/HeartbeatDispatcher.java
```diff
@@ -0,0 +1,52 @@
...
+  private final Map<SPath, Jupiter> clientDocs;
+
+  HeartbeatDispatcher(
+      ISarosSession sarosSession, UISynchronizer uiSynchronizer, Map<SPath, Jupiter> clientDocs) {
+    this.sarosSession = sarosSession;
+    this.uiSynchronizer = uiSynchronizer;
+    this.clientDocs = clientDocs;
+  }
+
+  public void start() {
```
tobous: Missing `@Override`.

Drakulix:
@srossbach Ping. Would still be nice, if you could answer on my last comment (no rust though).

srossbach:
> 
> 
> > Mixing java.util.concurrent and synchronized keyword is bad style.
> 
> I know, but I did not want to refactor the `JupiterClient` to remove the use of synchronized there (which I would prefer), because this makes this harder to review and introduces even more threading-related changes in a critical part of the infrastructure.
> 
> The HashMap, on the other hand, is accessed from `JupiterClient` the `Heartbeat`-Class and used in the UI-Thread. It makes perfect sense to use the concurrent implementation here imo.
> 
> Do you disagree with any of that? If yes, which one would you rather drop / whats your preferred solution for this?
> 

It isn't the best style to pass an ADT around to multiple classes. If you look at the code you might refactor it, either removing synchronized keywords are replacing the ConcurrentMap. The last approach will fail badly. Furthermore there were checks in the DocumentClient (I removed them .... for moving the class into the core, my bad ...) that strictly forbid wrong thread access. Normally you can only access the functionality on the main thread, period. It seems to work (the whole component) but it is still questionable to introduce concurrency into this component (which is already the case even without your patch).

> > Now 3 classes have a startable interface. Why ? When I said a dedicated class I meant that this class can be put into the session context container to enable heartbeats but not this start / stop chaining.
> 
> Yes, and how would I access the `clientDocs` then? The `JupiterClient` is not part of the container, but of the `ConcurrentDocumentClient` (CDC from here on). So should I instead hack-up a weird interface to access the private variable of a private variable of the CDC, which I then add to the `Heartbeat`-class which has nothing to do with the CDC from a conceptional point of view?
> 
> I think that is bad design as well, the heartbeat is part of a JupiterClient from a conceptual and from an implementation standpoint because both classes are very much coupled and it is the JupiterClients responsibility to send Jupiter-messages. (This PR fixes a missing detail of the client.)
> 
> I agree, that the new `Startable`-implementations do not look very nice, but I have laid out, why I do not think the Heatbeat belongs in the container. At least not with the current class-structure, which I do not wanna touch, because I want to touch as less as possible of that Jupiter stuff (another requirement for the PR btw).
> 
> We would not have gained much from adding it to the container functionality-wise anyway.
> 
> Your main requirement is done now. It is all in a separate class, which you can quickly comment out / revert in case something breaks.

Instead of chaining the start methods you could also offer package private methods to obtain the current managed documents. Even if this patch is submitted it needs still to be improved as it fires enough activities over the time, i.e imagine the client holds 1000 documents which result in over 1000 activities every minutes (and that activites are not as small as you think). I guess this can result in a spike every minute up to several MB network traffic.

Drakulix:
@srossbach 
I tried to address some of your concerns:
- Removed the Startable-implementations
- Added HeartbeatDispatcher to the context
- Added package-private accessors
- Changed the ADT to an explicit ConcurrentHashMap and added a comment to explain why it is needed.

If no one else has any objections, I start to polish this and add more docs.

core/src/saros/concurrent/management/JupiterClient.java
```diff
@@ -23,18 +23,15 @@ public JupiterClient(ISarosSession sarosSession) {
    * Jupiter instances for each local editor.
    *
    * @host and @client
+   *     <p>Note: This needs to be a *Concurrent*HashMap, because it is periodically iterated by the
+   *     HeartbeatDispatcher class.
    */
-  protected final HashMap<SPath, Jupiter> clientDocs = new HashMap<SPath, Jupiter>();
+  private final ConcurrentHashMap<SPath, Jupiter> clientDocs =
```
tobous:
```suggestion
  private final Map<SPath, Jupiter> clientDocs =
```

core/src/saros/concurrent/management/JupiterClient.java
```diff
@@ -23,18 +23,15 @@ public JupiterClient(ISarosSession sarosSession) {
...
+      new ConcurrentHashMap<SPath, Jupiter>();
```
tobous:
```suggestion
      new ConcurrentHashMap<>();
```

core/src/saros/concurrent/management/JupiterClient.java
```diff
@@ -71,4 +68,9 @@ public synchronized ChecksumActivity withTimestamp(ChecksumActivity checksumActi
 
     return get(checksumActivity.getPath()).withTimestamp(checksumActivity);
   }
+
+  // Package-private function for the HeartbeatDispatcher
+  ConcurrentHashMap<SPath, Jupiter> getClientDocs() {
+    return clientDocs;
```
tobous: I would prefer if you would return an unmodifable view of the map to ensure that the jupiter client still has full control of the map, i.e. no external class can change the content.

If you do adopt this change, this should also be mentioned in the javadoc of the method.
```suggestion
    return return Collections.unmodifiableMap(clientDocs);
```

core/src/saros/concurrent/management/HeartbeatDispatcher.java
```diff
@@ -0,0 +1,67 @@
...
+  private final UISynchronizer uiSynchronizer;
+  private final Map<SPath, Jupiter> clientDocs;
+
+  public HeartbeatDispatcher(
+      ISarosSession sarosSession,
+      UISynchronizer uiSynchronizer,
+      ConcurrentDocumentClient cdClient) {
+    this.sarosSession = sarosSession;
+    this.uiSynchronizer = uiSynchronizer;
+    this.clientDocs = cdClient.getJupiterClient().getClientDocs();
```
tobous: Holding this map of another class as an instance variable seems strange to me. It makes it hard to understand that the map is modified externally. I would prefer to hold the jupiter client instead and just request the map locally in `dispatchHeartbeats()` every time it it needed.

Even though this is probably worse performance-wise as you have to request the object multiple times (periodically), I still think it is worth it as it makes the code easier to understand.

core/src/saros/concurrent/management/JupiterClient.java
Drakulix: I think @srossbach wanted this to be explicit to make sure, everybody gets, that this is *concurrent*.

core/src/saros/concurrent/management/HeartbeatDispatcher.java
Drakulix: k

Drakulix:
Commit 57d800c1 works fine.

Heapdump for reference:
[heapdump-57d800c.zip](https://github.com/saros-project/saros/files/3350666/heapdump-57d800c.zip)

core/src/saros/concurrent/management/HeartbeatDispatcher.java
```diff
@@ -0,0 +1,68 @@
...
+    this.sarosSession = sarosSession;
+    this.uiSynchronizer = uiSynchronizer;
+    this.jupiter = cdClient.getJupiterClient();
+  }
+
+  @Override
+  public void start() {
+    heartbeatScheduledExecutor =
+        new ScheduledThreadPoolExecutor(1, new NamedThreadFactory("JupiterHeartbeat"));
+    // Schedule the heartbeat once per minute, this should be absolutely
```
tobous: This could be formatted nicer, but it does not really matter.

PR #401, opened 2019-01-25
[INTERNAL][HTML WB] Add shortcut manager

AbdBarho:
this class is responsible for handling all the shortcuts, currently only context menu mode is available, which would minify the menu and bring it next to the cursor for better usability, it also adds an overlay that would block all user clicks from drawing while in this mode.


de.fu_berlin.inf.dpp.whiteboard/frontend/app.js
```diff
@@ -23,8 +24,11 @@ class Whiteboard {
 
     ReactDOM.render(
       <Menu toolManager={this.toolManager}
-      canvasResizer={this.canvasResizer}/> ,
-      document.getElementById('menuContainer'));
+        canvasResizer={this.canvasResizer}
+        ref={menu => this.menu = menu} />,
+      document.getElementById('menuContainer'), () => {
+        this.shortcutManager = new ShortcutManager(this.menu);
```
m273d15: Why is it necessary to set the ShortcutManager after the render method is executed?

AbdBarho: It is certainly not necessary, but it is recommended.
There are some cases where react would create and render the component asynchronously and would return a promise (the JavaScript equivalent of `Future` objects), we want to be sure that the shortcut manager receives the actual menu component. 

[For more info](https://reactjs.org/docs/react-dom.html#render)

PR #402, opened 2019-01-25
[DOCS] Add docs for the whiteboard

AbdBarho:


docs/contribute/whiteboard.md
```diff
@@ -0,0 +1,34 @@
...
+Use the command `npm run build`. When building for production, all of the resources of the whiteboard will be bundled into one file `dist/index.html`.
+
+## Technology Stack
+The Saros HTML Whiteboard uses a graphics library called [Fabric.js](http://fabricjs.com/) to display the shapes in the page, it also uses [React](https://reactjs.org/) to render the menu.
+
+As for the bundling, it uses [Webpack](https://webpack.js.org/) to bundle the resources using [Babel](https://babeljs.io/) loader to transpile the code to an older more-browser-friendly version.
+
+## How it works internally
+
+The Saros HTML Whiteboard extends the [GEF (Graphical Editing Framework)](https://www.eclipse.org/gef/) based Whiteboard and uses most of the internal structure such as the [SXE](https://xmpp.org/extensions/xep-0284.html).
```
srossbach: Does it really extends something ? Isn't the HTML Whiteboard a REPLACEMENT for the currently uses Whiteboard which is only available when using Eclipse because it uses the GEF ?

AbdBarho: Agree, "extends" does not fit the HTML Whiteboard, I have changed it.

docs/contribute/whiteboard.md
```diff
@@ -0,0 +1,34 @@
...
+# {{ page.title }}
+
+The Saros Whiteboard was designed to enable Saros users to communicate ideas graphically.
+
+## Developing the Whiteboard
+
+### Setup
+Use the command `npm run setup`, note that `npm install` does not suffice.
+
+### Build for Development
```
m273d15: As far as I know it is necessary (as long as the HTML WB is not default) to uncomment the correponding view in the `plugin.xml`.

m273d15: Are there further steps to do? I was able to access the new WB, but nothing was synced between the participants.
I did the following:
* `npm run setup`, `npm run build`
* uncomment the new view
* start ALICE via the run configuration
* start BOB via the run configuration
* Connection with the xmpp account of alice, same for bob
* Start session with one file
* Open WB on ALICEs side, same for bob
* Draw something on ALICEs side
* Nothing happens on BOBs side (and vice versa)

Did I missed something?

AbdBarho: The problem was actually not in the setup process, there were missing dependencies (#423)

AbdBarho: I have updated the file to include un-commenting the view

PR #405, opened 2019-01-28
Improve recovery button usability

### LINKED ISSUE #340: Disable the recovery button for the host
As the host can never use the recovery button, it would be better to always disable it for them to not cause any confusion.

tobous:
#### [UI][I] Set disabled icon by default
Adjusts the toolbar buttons to also set the disabled icon by default.

#### [UI][I] Disable recovery button for host
The logic now only enables the recovery button for non-host
participants. This also removes unnecessary overhead by never
registering the consistency listeners that don't have any use for the
host.

Resolves #340.

#### [UI][I] Always keep recovery button disabled when consistent
Adjusts the logic of the recovery button to always remain disabled when
in a consistent state.

Subsequently adjusts the cleanup when the session ends to ensure that
the button is always disabled.

#### [UI][I] Improve recovery dialog message
Improves the message shown by the recovery dialog.

Subsequently moves the constant parts of the message into the
messages.properties.

#### [UI][I] Set tooltip for recovery button
Adds logic to set the tooltip of the recovery button according to the
local consistency state.

Adds logic to correctly set and reset the tooltip at the start and end
of a session.

Adjusts the tooltip messages in the message.properties.

#### [INTERNAL][I] Always enable recovery button after recovery
Adjust the logic in the consistency button to always re-enable the
button when button press action is done, even if the recovery is
executed. This moves the responsibility to correctly adjust the button
state to the consistency change listener.

This change allows the user to press the button again if the session is
still in an inconsistent state after the recovery was executed.


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/ui/views/buttons/ConsistencyButton.java
```diff
@@ -68,14 +68,24 @@ public void actionPerformed(ActionEvent e) {
...
         @Override
         public void sessionEnded(ISarosSession oldSarosSession, SessionEndReason reason) {
-          setSarosSession(null);
-          setEnabledFromUIThread(false);
+          if (!oldSarosSession.isHost()) {
+            setSarosSession(null);
+            setEnabledFromUIThread(false);
+
+          } else {
+            setToolTipText("Recover inconsistencies");
```
srossbach: I guess the tooltip should be the same when not in a session regardless if you were the session host or not in a former session.

tobous: The tooltip never gets updated for the client, so this is only needed for the host. But yes, since updating the tooltip would be a good idea, always resetting it when the session ends is the way to go.

tobous:
@srossbach I am not sure whether this change is that useful as it looks weird when the button is disabled for the host. Furthermore, it deviates from the Saros/E behavior.

Instead of actually disabling the button, my idea was to use an image with a gray arrow. The tool-tip can still remain adjusted. (I know this wouldn't do much for color-blind people and gray on gray is hard to see, but hopefully the tool-tip would be enough. Furthermore, clicking the button doesn't actually do anything.) What do you think about this?

But I don't want to create the image myself. With the PR #406, Saros/I also uses the recovery icons used in Saros/E. These icons were introduced by you under the folder "etool16". Can you tell me where you got them?

srossbach:
Do you mean this "dirty" hack in Saros/E ?

https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/actions/ConsistencyAction.java#L146

tobous:
No, not really. With the Saros/E behavior, I meant that the recovery button is not disabled for the host (or at least it doesn't look like it).

srossbach:
Ahh sorry, this one it is

https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/ui/actions/ConsistencyAction.java#L138

The button is actually disabled all the time, regardless if host or not. You can only click it when an inconsistency is found (which is not the case on the host side).

I do not know if IntelliJ offers an option to set the disabled image.

tobous:
Ah, thanks for the hint. Yes, this is possible in Saros/I as we are just using JButtons. I will adjust the logic to also always have the button disabled when the state is consistent.

tobous:
Rebased the PR onto #406 as it would otherwise be in conflict with the refactoring made in that PR.

Re-worked the PR to now always have the recovery button disabled by default and only enabling it the local user is in an inconsistent state.

tobous:
Rebased onto master without any interaction.

tobous:
Minor optional adjustment to javadoc to trigger build correctly.

PR #406, opened 2019-01-28
Replace recovery icon and move to using IconManager

tobous:
Replaces the famfamfam icons used to show the current recovery state with the ones currently used by Saros/E.

Replaces the usage of direct references to the icon paths by using the IconManager instead.

Does some cleanup in the changed classes.


tobous:

tobous:

PR #410, opened 2019-01-31
Move IntelliJ event handlers to a separate package

tobous:
Moves the IntelliJ event handlers to a separate package and groups them into sub-packages by the resource they are dealing with. Subsequently cleans up the handling of event handlers in `EditorManager`.

The bulk of the displayed changes are the new classes added with `[INTERNAL][I] Split LocalEditorStatusChangeHandler into separate classes`, which just contain copied logic from the former `LocalEditorStatusChangeHandler` class.

Also extracts the setup of the context tests to avoid code duplication.

#### [INTERNAL][I] Add method to resolve SPath from Document
Adds a public method to get a matching SPath for a given document if an
editor for the document is known to the editor pool.
The method is added to allow needed read access to the editor pool
without having to make the editor pool itself accessible.

Replaces direct access of the editor pool in the IntelliJ event
handlers with the introduced method. This is done in preparation of
the move of the event handlers to a separate package, as the editor pool
could no longer be accessed directly from outside the editor package.

Replaces the NotNull annotation for the parameter of
EditorPool.getFile(Document) with Nullable as null is a valid input with
which the method behavior is still well-defined.

#### [REFACTOR][I] Move IntelliJ event handlers to a separate package
Moves all event handlers dealing with IntelliJ events to the package
eventhandlers. The package contains sub-package which represent the
differentiation between the different resources the events deal with.

Adds package-info.java files to the added packages describing their
purpose.

Subsequently changes the visibility of used methods from package-private
to public. As these methods should still only be used internally, a note
was added to the javadoc of the changed methods.

Adjusts the visibility of the moved handlers as necessary.

Adjusts the test package-structure to match the new eventhandler package
structure.

#### [INTERNAL][I] Adjust NotNull annotations in event handlers
Adds NotNull annotations where appropriate in the event handlers.

Removes the NotNull annotations from the constructor parameters as these
parameters are all (indirectly) provided by the PicoContainer, meaning
they should not be able to be null during the normal runtime.
Furthermore, these annotations would make testing harder as you would
have to provide instances of the annotated classes in the container in a
test setup, even if they otherwise would not be necessary.

#### [INTERNAL][I] Unify access to en/disabling handlers in EditorManager
Creates methods to enable/disable all handlers for a specific resource.
Replaces direct access to the setEnabled(boolean) method of the handlers
with calls to the new wrapper methods.

This is done to make it easier to keep the state of all handlers for a
resource in sync when adding a new one. Otherwise, we would have to
manually search where the other handlers for the same resources are
enabled/disabled.

Moves the existing methods for this purpose to the same position as the
rest of the methods.

#### [INTERNAL][I] Split LocalEditorStatusChangeHandler into separate classes
Splits LocalEditorStatusChangeHandler into separate classes to better
differentiate between the different responsibilities.

Introduces the abstract parent class
AbstractLocalEditorStatusChangeHandler to avoid duplicating code needed
to set up each new handler.

Groups the handlers in EditorManager by the resource they are handling
to improve the overview of which handlers are used.

Adjusts the context test mock setup to handle multiple calls
instantiating a MessageBusConnection. This is needed as there now are
four classes that create a MessageBusConnection instead of only one.

#### [INTERNAL][I] Simplify context tests by extracting setup
Both context tests had almost exactly the same setup that was needed to
mock different parts of the IntelliJ API. To avoid the code duplication,
this setup was moved to an abstract parent class.

Does some minor cleanup in SarosIntellijContextTest.


srossbach:
This patch pulls us back to the discussion on how we should structure our packages, So do we really want *.eventhandler  or *.editor.eventhandler

If you are happy with the current package naming feel free to merge this patch.

tobous:
I understand the general consideration of whether these classes should be grouped by their general functionality as eventhandlers (which results in `*.eventhanders.editor.*`) or by the resources they are handling (which results in `*.editor.eventhandlers.*`).

My first approach was to just create a sub-package `*.editor.eventhander` and move the handlers there.
But this leaves us with the weird situation, that we also have an eventhandler in the `project` package that also implements `DisableableHandler`, so there isn't really a good place to put the interface.
It seemed like a nicer solution to me to put all IntelliJ eventhandlers into one package to better differentiate them from the Saros logic.

In general, I prefer this way of grouping the event handlers as the handling of IntelliJ events is their main purpose. They are almost exclusively working with IntelliJ resources / the IntelliJ API (the outlier here being the `PreexistingSelectionDispatcher`, which contains the shared check). Anything to do with the Saos logic is handled by calls to other classes.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/EditorManager.java
```diff
@@ -797,31 +773,78 @@ public void jumpToUser(@NotNull final User jumpTo) {
...
+   * @see #setLocalViewPortChangeHandlersEnabled(boolean)
    */
   private void setHandlersEnabled(boolean enable) {
-    localDocumentModificationHandler.setEnabled(enable);
-    localClosedEditorModificationHandler.setEnabled(enable);
-    localEditorStatusChangeHandler.setEnabled(enable);
-    localTextSelectionChangeHandler.setEnabled(enable);
-    localViewPortChangeHandler.setEnabled(enable);
+    setLocalDocumentModificationHandlersEnabled(enable);
+    setLocalEditorStatusChangeHandlerEnabled(enable);
```
m273d15: Why not `setLocalEditorStatusChangeHandlersEnabled`?

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/eventhandler/editor/editorstate/AbstractLocalEditorStatusChangeHandler.java
```diff
@@ -0,0 +1,92 @@
...
+   * @param project the current Intellij project instance
+   */
+  AbstractLocalEditorStatusChangeHandler(Project project) {
+    this.project = project;
+
+    this.enabled = true;
+  }
+
+  /** Creates a MessageBusConnection. */
+  protected void subscribe() {
```
m273d15: why not implementing a template method `subscribe` that calls an abstract method which has to be overwritten

tobous:
Corrected name to `setLocalEditorStatusChangeHandlersEnabled`.

Implemented template method pattern for `AbstractLocalEditorStatusChangeHandler` and subsequently reduced visibility of methods and parameters.

tobous:
@m273d15 Do you still have any issues with this PR?

tobous:

tobous:

PR #411, opened 2019-01-31
[FIX][I] Fix initialization of SessionUtils

### LINKED ISSUE #409: Shared module is not initialized correctly when session is started before an editor was opened
If a session is started in started in Saros/I without ever opening an editor during the current application runtime, the session is not correctly initialized.
No activities are shared as saros sees the shared module as not shared.

I have not investigated the issue enough yet but it seems to me like IntelliJ does some lazy initializing which leads to us relying on some state that has not yet been initialized.

**Update:** Not an issue with IntelliJ but an issue with lazy initialization/evaluation. See fix #411 for more details.

#### Steps to reproduce
1. Alice and Bob both start their IntelliJ instance. Both instances must not have any editors open when they are started.
2. Alice and Bob start a session on a module before ever opening an editor for any file.

tobous:
SessionUtils provides static access to methods to check whether a given
resource is shared. This check provides the current session object (if
existent), which is why it registers an ISarosSessionLifecycleListener
to update the locally held session object when there are any lifecycle
changes.

This listener was registered in a static block, which is problematic as
java uses lazy evaluation for static blocks, meaning they are only
computed once the object is accessed for the first time.
This leads to problems when the object is only first accessed after the
session has already started as the lifecycle listener will only be
registered at this point. In such situation, the SessionUtils object
will not have a session object even though there is currently a running
session, meaning it reports all resources as non-shared. This causes the
local saros instance to not send any activities for shared resources and
to drop most received activities.

This is the cause of the issue #409 as the SessionUtils is called
whenever an editor is opened or during the local session state
initialization if there are open editors. If both was not the case, the
SessionUtils will not have been initialized properly.

Resolves #409.

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/session/SessionUtils.java
+++ b/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/session/SessionUtils.java
@@ -41,6 +41,12 @@ public class SessionUtils {
 
   private static volatile ISarosSession currentSarosSession;
 
+  /**
+   * Initializes the SessionUtils. This is only used as part of the plugin context initialization by
+   * the PicoContainer and should not be called otherwise.
+   *
+   * @param sarosSessionManager the current SarosSessionManager instance
+   */
   public SessionUtils(ISarosSessionManager sarosSessionManager) {
     sarosSessionManager.addSessionLifecycleListener(lifecycleListener);
   }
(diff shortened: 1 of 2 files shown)


srossbach:
I thought a simple Class.ForName() ... in the lifecycle would be sufficient to explicitly load the class. However 

https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/context/AbstractContextLifecycle.java#L98

does not let you do that.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/session/SessionUtils.java
```diff
@@ -24,31 +22,27 @@
...
 
-            currentSarosSession = null;
-          }
-        });
-  }
+  private static volatile ISarosSession currentSarosSession;
 
-  private SessionUtils() {
-    // NOP
+  public SessionUtils(ISarosSessionManager sarosSessionManager) {
```
srossbach: Maybe you should add a comment that this must not be called.

tobous: Done.

PR #414, opened 2019-02-01
Introduce ToolBar containing Saros buttons

m273d15:
Introduce a ToolBar that contains the Saros buttons:
* Split-Button:
  * that implies the current connection state
  * lists the accounts

t

TODO in up-coming PR:
* Allow to change the active user and connect as the user via selecting an entry from the accounts list.


de.fu_berlin.inf.dpp.ui.frontend/html/src/components/MainView/ActiveAccount.jsx
```diff
@@ -0,0 +1,29 @@
+import { Text } from 'react-localize'
+import React from 'react'
+
+class ActiveAccount extends React.Component {
+  render () {
+    const { activeAccount } = this.props
+    var accountText
+    var className
+
+    if (typeof activeAccount === 'object' && activeAccount.username && activeAccount.domain) {
```
AbdBarho: can the active Account be anything other than an object?

if you simply want to check that it is not null or not undefined you try this:
`if (activeAccount && activeAccount.username && activeAccount.domain)`

de.fu_berlin.inf.dpp.ui.frontend/html/src/components/MainView/ActiveAccount.jsx
```diff
@@ -0,0 +1,29 @@
+import { Text } from 'react-localize'
+import React from 'react'
+
+class ActiveAccount extends React.Component {
+  render () {
+    const { activeAccount } = this.props
+    var accountText
```
AbdBarho: I would recommend using `let` instead of `var` whenever possible since it is [less error prone](https://stackoverflow.com/questions/762011/whats-the-difference-between-using-let-and-var-to-declare-a-variable-in-jav)

de.fu_berlin.inf.dpp.ui.frontend/html/src/components/ToolBar/ConnectionSplitButton.jsx
```diff
@@ -0,0 +1,51 @@
...
+class ConnectionSplitButton extends React.Component {
+  render () {
+    const connectionStateMap = {
+      'INITIALIZING': { onClick: this.props.onConnect, messageId: 'action.connect' },
+      'NOT_CONNECTED': { onClick: this.props.onConnect, messageId: 'action.connect' },
+      'CONNECTED': { onClick: this.props.onDisconnect, messageId: 'action.disconnect' },
+      'ERROR': { onClick: noop, messageId: 'action.connectionError' },
+      'CONNECTING': { onClick: noop, messageId: 'action.connecting' },
+      'DISCONNECTING': { onClick: noop, messageId: 'action.disconnecting' }
+    }
```
AbdBarho: I am not very familiar with how react updates it's contents internally, going from my naive assumption that react would basically call the render() function, it seems that `connectionStateMap` would be re-created every time even though nothing in its contents changes, maybe move it to the constructor?

The thing that I am worried about is that the internal browser in Eclipse on some machines is really old (IE9 on my machine at least), and since babel would transpile the code, it would even less efficient.

m273d15:
@AbdBarho Thanks for reviewing this PR. 👍

de.fu_berlin.inf.dpp.ui.frontend/html/src/components/MainView/ActiveAccount.jsx
m273d15: Done

Issue #394, opened 2019-01-23
Saros/I plugin causes assertion error when closing and then re-opening project


tobous:
When closing the current project and re-opening it (or another project), the Saros plugin causes an assertion error trying to use the old project object. (The full stacktrace is given below.) This most likely happens because the project object is held in the saros plugin context.

The project object is put into the plugin context through the [IntellijProjectLifecycle](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/IntellijProjectLifecycle.java), which in turn calls the [SarosIntellijContextFactory](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/context/SarosIntellijContextFactory.java#L105). The lifecycle is instantiated by the [SarosComponent](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/SarosComponent.java#L50), which hooks into IntelliJ through the [plugin.xml](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.intellij/META-INF/plugin.xml#L31)

The question now is why this context is not cleaned up between the project changes. Saros/I should be a project level component, meaning a new instance of the plugin should be created once the new project is opened, then also containing the correct project object.

#### Stacktrace
```Java
java.lang.AssertionError: Already disposed: Project (Disposed) HostTestProject
	at com.intellij.openapi.components.impl.ComponentManagerImpl.lambda$throwAlreadyDisposed$1(ComponentManagerImpl.java:248)
	at com.intellij.openapi.application.ReadAction.lambda$run$1(ReadAction.java:53)
	at com.intellij.openapi.application.impl.ApplicationImpl.runReadAction(ApplicationImpl.java:955)
	at com.intellij.openapi.application.ReadAction.compute(ReadAction.java:57)
	at com.intellij.openapi.application.ReadAction.run(ReadAction.java:53)
	at com.intellij.openapi.components.impl.ComponentManagerImpl.throwAlreadyDisposed(ComponentManagerImpl.java:246)
	at com.intellij.openapi.components.impl.ComponentManagerImpl.getPicoContainer(ComponentManagerImpl.java:239)
	at com.intellij.openapi.components.impl.ComponentManagerImpl.getPicoContainer(ComponentManagerImpl.java:45)
[...]

PR #417, opened 2019-02-03
[RELEASE FIX] Add workaround for project switching

tobous:
This PR is a workaround for the issue #394.

This is only meant as a temporary solution for the upcoming release. Afterwards, this issue will be solved by making Saros/I an application level plugin an explicitly handling the choice which project to share.

#### [RELEASE-FIX][I] Add workaround to replace old project object

THIS IS A RELEASE FIX! DO NOT MERGE ONTO MASTER!

Makes the IntellijProjectLifecycle singleton.

Introduces a wrapper for the held project object as it is changeable
during the plugin lifecycle.

Adds a workaround to replace the old project object when a new project
is opened and the currently held project object is disposed. This is
needed as the internal Saros logic otherwise tries to work on the old
project object which is no longer valid, leading to an assertion error.

This is only a very rudimentary workaround as it still leaves users in a
headless state if they open two projects and then close the one opened
first. This state is not handled gracefully by Saros (it will cause
assertion errors shown to the user in the IDE when trying to start a
session). The headless state is resolved once the user opens another
project which will then be used for Saros.

#### [RELEASE-FIX][I] Stop session when current project is closed

THIS IS A RELEASE FIX! DO NOT MERGE ONTO MASTER!

Stops the current session when the shared project is closed. This is
done as the session state is dropped at that point, meaning we can no
longer work with the old session.

Furthermore, dropping the old session should guarantee that we don't
have any references to resources of the closed project. This is
necessary as trying to access these resources would lead to an assertion
error.

#### [RELEASE-FIX][I] Adhere to Intellij plugin lifecycle setup

THIS IS PART OF A RELEASE FIX! DO NOT MERGE ONTO MASTER!

Moves the plugin initialization in SarosComponent from the CTOR into
initComponent() in order to adhere to the Intellij plugin lifecycle
setup guidelines. This is important as getComponent() calls are not
allowed for Intellij object from within the CTOR.

Subsequently improves the javadoc of SarosComponent.

#### [RELEASE-FIX][I] Display warning message when entering headless state

THIS IS A RELEASE FIX! DO NOT MERGE ONTO MASTER!

Displays a warning message to the user in all remaining open projects
when the shareable project is closed. Closing the shareable project
leaves Saros in a headless state where it does not have access to a
valid project object to load shareable modules from. This headless state
gets resolved when the user opens a new project. The newly opened
project will be used as the shareable project.

### DIFF
--- a/intellij/src/saros/intellij/IntellijProjectLifecycle.java
+++ b/intellij/src/saros/intellij/IntellijProjectLifecycle.java
@@ -4,10 +4,13 @@ import com.intellij.openapi.project.Project;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import org.apache.log4j.Logger;
+import org.jetbrains.annotations.NotNull;
 import saros.HTMLUIContextFactory;
 import saros.context.AbstractContextLifecycle;
 import saros.context.IContextFactory;
 import saros.intellij.context.SarosIntellijContextFactory;
+import saros.intellij.project.ProjectWrapper;
 import saros.intellij.ui.swt_browser.SwtLibLoader;
 
 /**
@@ -22,32 +25,48 @@ import saros.intellij.ui.swt_browser.SwtLibLoader;
  * FIXME: Project lifecycle causes serious issues, switch back to application lifecycle and correct the file system implementation as needed !
  */
 public class IntellijProjectLifecycle extends AbstractContextLifecycle {
+  private static final Logger log = Logger.getLogger(IntellijProjectLifecycle.class);
 
   private static IntellijProjectLifecycle instance;
+  private static ProjectWrapper projectWrapper;
 
   /**
-   * Creates a new IntellijProjectLifecycle singleton instance from a project.
+   * Returns the current instance of IntellijProjectLifecycle.
    *
-   * @param project
-   * @return
+   * <p>IntellijProjectLifecycle is singleton, meaning this call will always return the same object.
+   * If a new project object is passed, it will replace the currently held project object in the
+   * plugin context.
+   *
+   * @param project the project object to put into the plugin context
+   * @return the singleton instance of the IntellijProjectLifecycle
    */
-  public static synchronized IntellijProjectLifecycle getInstance(Project project) {
-    instance = new IntellijProjectLifecycle(project);
+  public static synchronized IntellijProjectLifecycle getInstance(@NotNull Project project) {
+    if (instance != null) {
+      if (projectWrapper.getProject().isDisposed()) {
+        log.debug(
+            "Replacing disposed project currently held in plugin context with project "
+                + project.getName());
+
+        projectWrapper.setProject(project);
+      }
+
+    } else {
+      log.debug("Initializing plugin context with project " + project.getName());
+      instance = new IntellijProjectLifecycle(project);
+    }
 
     return instance;
   }
 
-  private Project project;
-
   private IntellijProjectLifecycle(Project project) {
-    this.project = project;
+    projectWrapper = new ProjectWrapper(project);
   }
 
   @Override
   protected Collection<IContextFactory> additionalContextFactories() {
     List<IContextFactory> nonCoreFactories = new ArrayList<>();
 
-    nonCoreFactories.add(new SarosIntellijContextFactory(project));
+    nonCoreFactories.add(new SarosIntellijContextFactory(projectWrapper));
 
     if (SarosComponent.isSwtBrowserEnabled()) {
       SwtLibLoader.loadSwtLib();
--- a/intellij/src/saros/intellij/IntellijProjectLifecycle.java
+++ b/intellij/src/saros/intellij/IntellijProjectLifecycle.java
@@ -6,6 +6,7 @@ import java.util.Collection;
 import java.util.List;
 import org.apache.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import saros.HTMLUIContextFactory;
 import saros.context.AbstractContextLifecycle;
 import saros.context.IContextFactory;
@@ -60,6 +61,16 @@ public class IntellijProjectLifecycle extends AbstractContextLifecycle {
 
   private IntellijProjectLifecycle(Project project) {
     projectWrapper = new ProjectWrapper(project);
+  }
+
+  /**
+   * Returns the current project held by the lifecycle.
+   *
+   * @return the current project held by the lifecycle
+   */
+  @Nullable
+  Project getProject() {
+    return projectWrapper.getProject();
   }
 
   @Override
--- a/intellij/src/saros/intellij/SarosComponent.java
+++ b/intellij/src/saros/intellij/SarosComponent.java
@@ -18,7 +18,7 @@ import saros.session.SessionEndReason;
 import saros.util.ThreadUtils;
 
 /**
- * Component that is initalized when a project is loaded. It initializes the logging, shortcuts and
+ * Component that is initialized when a project is loaded. It initializes the logging, shortcuts and
  * the {@link IntellijProjectLifecycle} singleton.
  */
 public class SarosComponent implements com.intellij.openapi.components.ProjectComponent {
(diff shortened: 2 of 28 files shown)


tobous:
Hmm, that's weird.

For whatever reason, the branch build failed because of an error in the JS part of the whiteboard. The normal tests all ran without a problem locally.

And the PR build fails because of a [`DataTransferManagerTest`](https://scans.gradle.com/s/xmv6ih5qivehi/tests/rr6z7ipbhecyo-tdj3ugxguvmva?openStackTraces=WzBd), which is weird as I didn't modify the core.

tobous:
The build-failure seems to be unrelated to this PR as it can be re-created on the current master. I created issue #418 for this.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/SarosComponent.java
```diff
@@ -94,6 +105,16 @@ public void projectOpened() {
 
   @Override
   public void projectClosed() {
-    // TODO: Update project
+    ISarosSessionManager sarosSessionManager =
```
srossbach: I guess this needs a lot more of explanation. If you look at the code you are just wondering why the lifecycle is not started in projectOpened and is not stopped/disposed on projectClosed

tobous: This is just a WIP. It seems to work, but I wanted your opinion on what other pitfalls there are with this handling.

I was also wondering why we don't do anything in `projectOpened`. I will look into what the CTOR and what `projectOpened` should be used for.

tobous:
Rebased onto the current master without any interactions.

tobous:
Adjusted `SarosComponent` to adhere to [IntelliJ plugin lifecycle setup guidelines](http://www.jetbrains.org/intellij/sdk/docs/basics/plugin_structure/plugin_components.html#plugin-components-lifecycle).

Improves javadoc of `SarosComponent` and `ProjectWrapper`.

tobous:
@srossbach Is the new documentation sufficient?

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/SarosComponent.java
```diff
@@ -47,7 +71,54 @@ public SarosComponent(final Project project) {
...
+  }
+
+  /**
+   * {@inheritDoc}
+   *
+   * <p>Does nothing for the Saros plugin as the initialization is already done in {@link
+   * #initComponent()}.
+   */
+  @Override
+  public void projectOpened() {
```
srossbach: Isnt this called multiple times. As far as I can remember you should not expect that projectClosed() is called before projectOpened() is called.

To keep the issue simple: Isn't there a option to shutdown Saros before a new project is either opened or the user switches the project ?

tobous: @srossbach Yes, this gets called multiple times and yes, this can get called before the last project was opened.

The problem is that each setup method is called for ever project that is opened (CTOR, `initComponent()`, and `projectOpened()`), even if there is already an open project.

I was not sure how to handle the case that a user opens a second project, so my previous inclination was to exclude the case through the release notes (which is not a very good handling of the case). But not mentioning the problem at all was a mistake. My bad.

But thinking about it now, my preferred handling of this case would be that Saros only acts on the project that was opened first. Any additional projects will be ignored as long as there is currently an open project. (Later on, we can deal with this by always using the project that is focused by the user when trying to start a session, but I don't think this should be implemented for the first alpha. This should be done when moving to Intellij application scope.)
This would mean that we would have to check whether there is currently already a project object in the context and only add the opened project if there isn't one (no possible with the current handling, but easily adjustable; would also mean null-checks, but violations here would have thrown disposed access assertions instead anyway).

This still leaves us with a couple of problems that I don't know how to handle gracefully:
- What do we do when the user closes the shareable project? This gets worse when there are multiple open projects. Even if we store all project object, which one do we choose?
- The [`NegotiationHandler`](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/core/ui/eventhandler/NegotiationHandler.java#L111) still uses a deprecated method that chooses the project object by looking at the current focus. I am not sure if it would be possible to create a weird corner case with this that leads to problems.

Furthermore, we would have to adjust [`SarosFileShareGroup`](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/ui/menu/SarosFileShareGroup.java#L84) to check whether the focused project is shareable (or update the context).

In general, there is the potential for a lot of corner cases when allowing multiple open project and I don't think we can handle them well with the current setup. So I am not sure whether allowing multiple open projects is really a good idea before moving the plugin to application scope. (I know this somewhat contradicts my statement above, but I am pondering as I am writing :wink:).

We can still release a second version not long after , but I would like to get something out there soon-ish so that we at least appear on the radar and to show that something is actually happening.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/IntellijProjectLifecycle.java
```diff
@@ -24,30 +26,40 @@
...
+   * <p>IntellijProjectLifecycle is singleton, meaning this call will always return the same object.
+   * If a new project object is passed, it will replace the currently held project object in the
+   * plugin context.
+   *
+   * @param project the project object to put into the plugin context
+   * @return the singleton instance of the IntellijProjectLifecycle
    */
-  public static synchronized IntellijProjectLifecycle getInstance(Project project) {
-    instance = new IntellijProjectLifecycle(project);
+  public static synchronized IntellijProjectLifecycle getInstance(@NotNull Project project) {
```
m273d15: Add information in the commit message that you make this class a real singleton

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/SarosComponent.java
```diff
@@ -47,7 +56,9 @@ public SarosComponent(final Project project) {
       LogLog.error("could not load saros property file 'saros.properties'", e);
```
m273d15: Check whether LogLog is required

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/SarosComponent.java
m273d15: I also think that the restriction of "one open project at a time" is acceptable for the first alpha version. This (change of plugin licecycle to application lifecycle) is also a point that should we document as "ongoing work".

tobous:
Fixed merge conflict with master.

tobous:
I am currently working on a proper solution for this problem that involves moving the project object into the session context and moving the plugin to the application level. This would make this workaround unnecessary.

As this workaround introduces unnecessary wrappers and does not contribute to the state after the real solution was implemented, I don't think it should be merged onto the master.
I would suggest that, in case we want to release a first alpha before the real solution is implemented, we create a release branch that this PR gets merged onto as a final step. I would still have to adjust it before that to implement a usable behavior (probably holding the first opened project and only loading the new one into the context if there currently isn't one; the current handling of always using the last opened project is not usable as it creates a lot of trouble if a new project is opened during a session).

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/IntellijProjectLifecycle.java
tobous: Done.

de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/SarosComponent.java
tobous: As it is currently not planned that this PR remains on the master, I would prefer not dealing with it here. It does not seem to cause problems.

tobous:
Rebased onto current master.

Added some logging and ensured that logging is initialized in time.

Adjusted IntellijProjectLifecycle to always use the project that was opened first and to only replace the held project object when it is disposed (meaning the project was closed). This now still can't handle closing the current project with another project open, as the project that is remaining open will not be recognized by IntelliJ and using the menu at this point will lead to issue.

Adjusted SarosComponent to only end the session when the shared project is closed.

tobous:
Rebased the PR on the current master. Resolved a lot of path conflicts caused by the project structure renaming.

tobous:
Switched the target of the PR to the newly created release branch for Saros/I. Rebased the PR on the current release branch. Added a warning message informing the user when they are entering a headless state.

PR #419, opened 2019-02-04
[INTERNAL][HTML WB] fix #418

### LINKED ISSUE #418: Travis builds fail while building HTML whiteboard
Currently, all travis builds seem to fail because of the HTML whiteboard.

As the build of the master after merging the last PR worked and both the current master and the state before the fix fail, this seems to be caused by an external issue.

[Build on the current master (with NOP commit)](https://travis-ci.com/saros-project/saros/jobs/174967481)
[Build before the null check addition](https://travis-ci.com/saros-project/saros/jobs/174968529)

AbdBarho:
set the version of the breaking terser library to an older version.
once the problem is fixed this can be removed.

### DIFF
--- a/de.fu_berlin.inf.dpp.whiteboard/frontend/package.json
+++ b/de.fu_berlin.inf.dpp.whiteboard/frontend/package.json
@@ -36,6 +36,7 @@
     "json3": "^3.3.2",
     "style-loader": "^0.21.0",
     "webpack": "^4.28.1",
+    "terser": "3.14.1",
     "webpack-cli": "^3.2.1"
   }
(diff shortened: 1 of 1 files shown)


AbdBarho:
@m273d15 what a coincidence, we both added the same patch at the same time :p
anyway here is the fix for the whiteboard, the build is successful.
when the problem is fixed in the terser plug-in I will create  a new pull request for the updated dependencies.

m273d15:
I violated our rules (waiting for two approvals / or wait two days) since this fix is important, very small and therefore manageable

PR #423, opened 2019-02-04
[INTERNAL][HTML WB] Add missing  GSON dependency 

AbdBarho:
the file for this library was missing, I updated the version and added the missing file.

this missing library did not cause any problems in compile time, but it caused an issue in run time.

@m273d15 this fixes the freeze problem, my assumption is that Java is searching everywhere for the class but cant find it.


srossbach:
When there are no compile issues then this library is loaded dynamically. I did not find the code that does that. Where is it located ?

m273d15:
The gson library is also still in a `lib` dir.
Therefore you have to remove the `lib` dir and add `releaseDep 'com.google.code.gson:gson:2.8.5'`
into the gradle `dependencies` block:
```gradle
...
dependencies {
  compile project(':de.fu_berlin.inf.dpp.core')
  compile project(':de.fu_berlin.inf.dpp')
  releaseDep (log4jVersion) {
      exclude group: 'com.sun.jmx', module: 'jmxri'
      exclude group: 'com.sun.jdmk', module: 'jmxtools'
      exclude group: 'javax.jms', module: 'jms'
    }
  releaseDep 'com.google.code.gson:gson:2.8.5'
  testCompile configurations.testConfig
}
...
```

PR #424, opened 2019-02-04
Fix recovery button state

### LINKED ISSUE #339: Recovery button does not reset when session ends during inconsistency
The recovery button is not reset properly when the session ends while the client is in an inconsistent state instead, the button remains active in the inconsistent state. If the button is clicked in this state, the dialog whether the user want to start a recovery is still shown. If such a recovery is started, nothing happens and the button is subsequently disabled. Furthermore, the message that the user is in an inconsistent state is displayed again when the session ends.

#### Steps to Reproduce:
1. Alice (host) and Bob start a session with a shared file A.
2. Create an inconsistency (see below on an easy setup on how to do this).
3. Wait until Bob detects the inconsistency.
4. End the session (either by Bob or Alice leaving the session).

For an easy setup to generate an inconsistent content state for a file, simply add the following statement before the activity generation in [StoppableDocumentListener](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/editor/StoppableDocumentListener.java#L69). This will cause Saros to transmit a "B" every time an "A" is typed in a shared editor.
```Java
if (newText.equals("A")) newText = "B";
```

tobous:
#### [FIX][I] #339 Ignore consistency change without a current session
Adjusts the ConsistencyButton to ignore consistency changes when there
currently is no active session.

With the old logic, if a participant would leave the session while they
are in an inconsistent state, the logic would react to the consistency
change caused by the session end by re-enabling the button and
displaying the inconsistency warning message again.

Fixes #339.


tobous:
Dropped commit that changed the logic to always have the recovery button enabled.

Changed the target of the PR to master as it no longer contains changes influenced by the refactoring in #406.

PR #426, opened 2019-02-05
[INTERNAL][CORE] User instead of JID in ProjectNegotiation

stefaus:
The Negotiation class stores the JID to identify the remote participent. 
Multiple calls in its ProjectNegotiation subclasses need the User object
which has to be retrived via session.getUser(jid). A call which result
can be null when a user left the session and information which is partly
already available in the call stack.

This patch removes the need of some jid -> user -> jid -> user 
conversions (removes / simplify some checks) and partly the dependecy
on JIDs. As SessionNegotiation continues to use JID for now this moves
storing of the peer to the sub classes. The focus of this change lies in
the ProjectNegotiation classes as a base for further refactorings.


de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractIncomingProjectNegotiation.java
```diff
@@ -477,15 +477,15 @@ protected void awaitActivityQueueingActivation(IProgressMonitor monitor)
       throws SarosCancellationException {
 
     monitor.beginTask(
-        "Waiting for " + getPeer().getName() + " to continue the project negotiation...",
+        "Waiting for " + getRemoteUser() + " to continue the project negotiation...",
```
srossbach: You should not expect that the toString() method of objects deliver something that is meant to be read by a normal user. So the correct replacement is getRemoteUser().getJID().getName()

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/Negotiation.java
```diff
@@ -70,10 +67,8 @@
    * @param transmitter transmitter used for sending negotiation messages
    * @param receiver receiver used for receiving negotiation messages
    */
-  protected Negotiation(
```
srossbach: I do not get the changes made here. You have an user object in the subclasses which has access to a JID and so there is no reason to make getPeer abstract.

stefaus: Yeah sorry this was based on my misconception that some places would call Negotiation.getPeer(). I mixed that up with something different and now removed the abstract method.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractOutgoingProjectNegotiation.java
```diff
@@ -117,10 +116,10 @@ public Status run(IProgressMonitor monitor) {
 
       transfer(monitor, fileLists);
 
-      User user = session.getUser(getPeer());
-      if (user == null) throw new LocalCancellationException(null, CancelOption.DO_NOT_NOTIFY_PEER);
+      if (!getRemoteUser().isInSession())
```
srossbach: The check is pointless. The user could leave the session 1 ms after you did the check.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/Negotiation.java
srossbach: You misunderstood me. You should not remove this method. You should change the CTOR super calls in the affected subclasses e.g AbstractProjectNegotiation(User user ....){super(user.getJid()) ...

Your commit message states that you did a "refactor" (this patch here is not a refactor as we use the refactor tag) on the project negotiation classes. This class here is the base class. So either put the code back or amend your commit message.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractIncomingProjectNegotiation.java
stefaus: With retrieve I was for now just looking at `jid.getName()`, but yes XmppConnection in User is not a good idea. So probably set the display name in the CTOR and unify the whole sometimes the GUI is displaying the jid and sometimes a nick name.

srossbach:
I had added a feature where you could chose you own session name regardless of the nickname of the JID. It got removed because the names did not match the one in the roster.

On the other hand nicknames are never consistent because on user A user C can be displayed as "Foo" and on user B user C is displayed as "Bar"

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/Negotiation.java
stefaus: Ok, sorry for the wrong tag.
I changed the title / commit message.

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/AbstractOutgoingProjectNegotiation.java
stefaus: I am also not satisfied with the cancellation part here, but should I really change this in this PR?

srossbach: Nope

de.fu_berlin.inf.dpp.core/src/de/fu_berlin/inf/dpp/negotiation/ProjectNegotiation.java
```diff
@@ -155,4 +161,17 @@ protected void monitorFileTransfer(FileTransfer transfer, IProgressMonitor monit
   protected void notifyTerminated(NegotiationListener listener) {
     listener.negotiationTerminated(this);
   }
+
+  /**
+   * Returns the JID of the peer with which the negotiation takes place.
+   *
+   * @return peer JID
+   */
+  public JID getPeer() {
```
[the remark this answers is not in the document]
stefaus: Now ProjectNegotiation can store User final, moving `setPeer` to it's only caller and no additional JID is stored.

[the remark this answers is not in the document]
srossbach: This is still duplicated code.

As the getPeer method was not marked as final you could use override it in the affected class, i.e make the JID private final in the Negotiation class and let getPeer return this final JID. This would at least 3 classes work as they should. You could then hack together you own stuff in the affected class.

As for the JID stuff. If you can ensure that the correct fully qualified JID is passed in from the UI (the one with Saros) you could drop the JID query. I have my doubt that Saros/E will present you the correct JID all the time and I have no clue what Saros/I is passing in.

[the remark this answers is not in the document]
srossbach: ?

private final JID jid;
public Negotiation(final JID jid ,,,){this.jid = jid} public getPeer() {return jid}

private final User user;
public ProjectNegotiation(final User user ...) {super(user.getJID()); this.user = user} public getRemoteUser() {return user}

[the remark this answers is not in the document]
srossbach: Do you even need getPeer() in the ProjectNegotiation ?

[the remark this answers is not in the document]
stefaus: > private final JID jid;

You can't make this final without changing `OutgoingSessionNegotiation` and I don't understand where you see the advantage of keeping jid in Negotiation.

> Do you even need getPeer() in the ProjectNegotiation ?

not in ProjectNegotiation but didn't want to change dependent classes now. Can mark it deprecated if it helps.

stefaus:
will take up this work at a later date

PR #432, opened 2019-02-08
reimplemented renderer classes

danielpaul1986:
Part 4: All renderer classes reimplemented in kotlin


m273d15:
Please re-open if required

PR #434, opened 2019-02-08
reimplemented all model classes

danielpaul1986:
Part 6: All model classes reimplemented: Contact, ProjectTree, State, ValidationResult


m273d15:
Please re-open if required

PR #435, opened 2019-02-08
reimplemented ide_embedding classes

danielpaul1986:
Part 7: All classes in ide_embedding reimplemented in kotlin


m273d15:
Please re-open if required

PR #437, opened 2019-02-08
first part of browserfunctions reimplemented: 

danielpaul1986:
Part 9: Browserfunction reimplemented from AddContact to DisconnectAccount in kotlin


m273d15:
Please re-open if required

PR #439, opened 2019-02-09
[INTERNAL][HTML WB] Update dependencies (fix #418)

### LINKED ISSUE #418: Travis builds fail while building HTML whiteboard

AbdBarho:
update the whiteboard dependencies to latest versions.

most importantly, use Babel 7 instead of Babel 6.

this also fixes (#418)


srossbach:
The point of a Ci is to disallow submitting patches that cause a build to fail. Pushing patches to recover from such an issue should not happen.

AbdBarho:
Long story short, it was decided in the Saros team that a certain file with the name `package-lock.json` (which contains all versions of all sub-dependencies (and sub-sub-dependencies and so on...) in any JavaScript/node based project) not be committed to the repository, mainly because of its huge size (some times reaching 10 000 lines) and because the git commit pollution that would result from it.

One of the internal sub-dependencies of the whiteboard (named `terser`) got a breaking update which made it not compatible with other dependencies of the whiteboard (more specifically, not compatible with `webpack`), and since the `package-lock.json` was not available, the latest version of `terser` was installed and it broke the build process.

As a work around, `terser` was added as a dependency with a fixed version (see #418 #419).

Now that breaking change has been reset from the developers of `terser` and we can use the latest version.

I know that this will not solve the problem permanently, and the problem will remain open in every web-based project as long as the `package-lock.json` file is not committed.

this needs to be discussed in the Saros team, but as it stands now, the file is still excluded from the repository.

de.fu_berlin.inf.dpp.whiteboard/frontend/.eslintrc.json
```diff
@@ -20,7 +20,7 @@
   ],
   "settings": {
     "react": {
-      "version": "detect"
+      "version": "latest"
```
m273d15: What does "latest" mean? Does it mean that always the eslint rules of the latest react release are used (even if we don't use the latest version)?

de.fu_berlin.inf.dpp.whiteboard/frontend/package.json
```diff
@@ -2,41 +2,43 @@
   "name": "whiteboard",
   "version": "0.0.1",
   "description": "Whiteboard for Saros built with HTML",
+  "author": "Abdullah Barhoum",
```
tobous: We generally don't use author tags in the Saros sourcecode.

de.fu_berlin.inf.dpp.whiteboard/frontend/.eslintrc.json
AbdBarho: this change will fit more likely to another patch that I will be adding soon.
I will remove this now
But yes "latest" would mean latest react version even if were are not using the latest one.

de.fu_berlin.inf.dpp.whiteboard/frontend/package.json
AbdBarho: I had it there even before saros, I will remove it in the next commit.

PR #441, opened 2019-02-12
Rework documentation layout 

m273d15:
Changed the layout of the documentation to a common and simple format.
See [my forked documentation](https://m273d15.github.io/saros/documentation/) for a running version.


docs/_includes/footer.html
```diff
@@ -1,6 +1,10 @@
...
-  </div>
+<footer class="px-5 page-footer">
+  <span class="text-muted" id="copyright">Copyright &copy; {{ site.time | date: '%Y' }}, <a class="no-underline" href="{{ '/documentation/history.html' | relative_url }}">Saros Team</a></span>
+  <span class="float-right no-underline">
+    <a href="{{ site.contact.gitter.channel-link }}"><i class="fab fa-gitter"></i></a>
+    <a href="{{ site.contact.twitter.user-link }}"><i class="fab fa-twitter"></i></a>
+    <a href="{{ site.contact.github.project-link }}"><i class="fab fa-github"></i></a>
+    <a class="pr-3" href="{{ site.contact.google-groups.user-group-link }}"><i class="fas fa-envelope"></i></a>
+    <a href="http://www.mi.fu-berlin.de/fb/impressum/" id="impressum" target="_blank">Imprint</a>
+  </span>
```
AbdBarho: Would it make more sense to have these as part of the navbar? just like the "view on github" button. they are too small and hidden way down in the corner at the end of the page, I actually had to search for them to find them.

docs/_includes/forkme_banner.html
```diff
@@ -1 +1 @@
-<a class="forkme_banner" href="{{ site.github.repository_url }}">View on GitHub <i class="fab fa-github"></i></a>
+<a class="forkme_banner" href="{{ site.contact.github.project-link }}">View on GitHub <i class="fab fa-github"></i></a>
```
AbdBarho: is it possible to make this not disappear when scrolling down the page? always visible means that potential developers will more likely click on it.

docs/_includes/toc.html
```diff
@@ -1,4 +1,87 @@
...
+        {% for i in (1..indentAmount) %}
+            {% assign space = space | prepend: '    ' %}
+        {% endfor %}
+
+        {% unless include.item_class == blank %}
+            {% capture listItemClass %}{:.{{ include.item_class | replace: '%level%', headerLevel }}}{% endcapture %}
+        {% endunless %}
+
+        {% capture my_toc %}{{ my_toc }}
+{{ space }}{{ listModifier }} {{ listItemClass }} [{% if include.sanitize %}{{ header | strip_html }}{% else %}{{ header }}{% endif %}]({% if include.baseurl %}{{ include.baseurl }}{% endif %}#{{ html_id }}){% if include.anchor_class %}{:.{{ include.anchor_class }}}{% endif %}{% endcapture %}
```
AbdBarho: I have very little experience with jekyll, and I kind of understand what you are doing here, but I dont understand why.
It seems like you are taking an html file and trying to parse the contents of table of contents using each <h></h> in that file, is that correct?

it would also be great if these big lines of code were on multiple lines so it would be easier to follow along. I am assuming that multi lines wont affect the resulting document at the end. some lines (like this one) also have what I assume to be incorrect indentation

m273d15: Yes this is correct. The header ids (which are used) are generated by jekyll/kramdown.

This file is more or less an external dependency from https://github.com/allejo/jekyll-toc . Unfortunately Github Pages provides a jekyll hosting with a restricted set of plugins. Therefore  I had to use this solution instead of importing a plugin.

docs/_includes/forkme_banner.html
m273d15: Ah, this was not intended that the banner disappears. I will fix this.

docs/_includes/footer.html
m273d15: Sure I already thought about it. I think if I move these icons into the navbar, the fork-me banner could be removed, since we already link to the repository via the icons.

m273d15: Is this better?
navbar_with_icons

AbdBarho: much better

PR #442, opened 2019-02-12
Update plugin.xml

tobous:
#### [I] Update plugin id
Sets the plugin id to "saros" instead of
"de.fu_berlin.inf.dpp.intellij".

Subsequently updates the reference to the plugin id in SarosComponent.

Also replaces all other places that had a hard-coded reference to the
actual plugin id with references to the variable defined in
SarosComponent.

#### [I] Update plugin.xml
Updates the information in the plugin.xml.
- Sets the correct minimal Intellij version
- Updates the plugin description
- Updates the change notes
- Adds a vendor email address
- Removes the deprecated modifier "version" from the idea-plugin field
- Replaces the links to Intellij documentation with new version

#### [REFACTOR][I] Restructure plugin.xml to match template

### DIFF
--- a/de.fu_berlin.inf.dpp.intellij/META-INF/plugin.xml
+++ b/de.fu_berlin.inf.dpp.intellij/META-INF/plugin.xml
@@ -1,23 +1,23 @@
-<idea-plugin version="2">
+<idea-plugin>
     <id>saros</id>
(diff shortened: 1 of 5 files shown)


tobous:

PR #444, opened 2019-02-14
[CORE][FEATURE] Add Manager And Activities To Share Commits

chrismotian:
Recreated #400

### DIFF
--- a/core/src/saros/git/GitManager.java
+++ b/core/src/saros/git/GitManager.java
@@ -1,0 +1,121 @@
+package saros.git;
+
+import java.io.File;
+import java.io.IOException;
+import org.apache.log4j.Logger;
+import org.picocontainer.Startable;
+import saros.activities.GitCollectActivity;
+import saros.activities.GitRequestActivity;
+import saros.activities.GitSendBundleActivity;
+import saros.annotations.Component;
+import saros.session.AbstractActivityConsumer;
+import saros.session.AbstractActivityProducer;
+import saros.session.IActivityConsumer;
+import saros.session.IActivityConsumer.Priority;
+import saros.session.internal.SarosSession;
+
+/**
+ * This Class is responsible for the negotiation between user that want to sent,receive and fetch
+ * commits.
+ *
+ * <p>In order to start the negotiation the user needs to call the procedure {@code
+ * sendCommitRequest()}.
+ *
+ * <p>Before starting negotiation the user need to set up his working directory tree by calling the
+ * procedure {@code changeWorkDirTree(File workDirTree)}. After calling it the first time the
+ * Manager hold a {@link JGitFacade} object.
+ */
+@Component(module = "core")
+public class GitManager extends AbstractActivityProducer implements Startable {
+
+  private static final Logger log = Logger.getLogger(GitManager.class.getName());
+
+  private final SarosSession session;
+
+  JGitFacade jGitFacade;
+
+  public GitManager(SarosSession session) {
+    this.session = session;
+  }
+
+  @Override
+  public void start() {
+    session.addActivityProducer(this);
+    session.addActivityConsumer(consumer, Priority.ACTIVE);
+  }
+
+  @Override
+  public void stop() {
+    session.removeActivityProducer(this);
+    session.removeActivityConsumer(consumer);
+  }
+
+  public void sendCommitRequest() {
+    fireActivity(new GitRequestActivity(session.getLocalUser()));
+  }
+
+  public void changeWorkDirTree(File workDirTree) throws IOException {
+    log.info("You changed your working directory tree");
+    if (jGitFacade == null) {
+      this.jGitFacade = new JGitFacade(workDirTree);
+    } else {
+      this.jGitFacade.setWorkDirTree(workDirTree);
+    }
+  }
+
+  private final IActivityConsumer consumer =
+      new AbstractActivityConsumer() {
+        @Override
+        public void receive(GitRequestActivity activity) {
+          if (jGitFacade != null) {
+            log.info("You recieved a request to fetch commits");
+            try {
+
+              String basis = jGitFacade.getSHA1HashByRevisionString("HEAD");
+              fireActivity(new GitCollectActivity(session.getLocalUser(), basis));
+
+            } catch (IOException e) {
+              log.debug(session.getLocalUser().toString() + " can't access the Git Directory");
+            }
+          } else {
+            log.info("Your workDirTree is not selected.");
+          }
+        }
+
+        @Override
+        public void receive(GitCollectActivity activity) {
+          if (jGitFacade != null) {
+            log.info("Your request to send your commits was accepted.");
+            String basis = activity.getBasis();
+            try {
+              byte[] bundle = jGitFacade.createBundle("HEAD", basis);
+              fireActivity(new GitSendBundleActivity(session.getLocalUser(), bundle));
+
+            } catch (IOException e) {
+              log.info("Please check that the recievers HEAD is ancestor of your HEAD.");
+              log.debug("failed at create bundle");
+            } catch (IllegalArgumentException e) {
+              log.debug(e);
+            }
+          } else {
+            log.info("Your workDirTree is not selected.");
+          }
+        }
+
+        @Override
+        public void receive(GitSendBundleActivity activity) {
+          if (jGitFacade != null) {
+            log.info("You recieving the commits between your HEAD and the senders HEAD");
+            byte[] bundle = activity.getBundle();
+            try {
+              jGitFacade.fetchFromBundle(bundle);
+              log.info("Fetched the commits");
+            } catch (Exception e) {
+              log.debug(e);
+            }
+          } else {
+            log.info("Your workDirTree is not selected.");
+          }
+        }
+      };
+}
--- a/core/src/saros/git/GitManager.java
+++ b/core/src/saros/git/GitManager.java
@@ -7,7 +7,6 @@ import org.picocontainer.Startable;
 import saros.activities.GitCollectActivity;
 import saros.activities.GitRequestActivity;
 import saros.activities.GitSendBundleActivity;
-import saros.annotations.Component;
 import saros.session.AbstractActivityConsumer;
 import saros.session.AbstractActivityProducer;
 import saros.session.IActivityConsumer;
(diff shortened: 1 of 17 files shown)


core/src/saros/git/GitManager.java
```diff
@@ -0,0 +1,121 @@
...
+ * commits.
+ *
+ * <p>In order to start the negotiation the user needs to call the procedure {@code
+ * sendCommitRequest()}.
+ *
+ * <p>Before starting negotiation the user need to set up his working directory tree by calling the
+ * procedure {@code changeWorkDirTree(File workDirTree)}. After calling it the first time the
+ * Manager hold a {@link JGitFacade} object.
+ */
+@Component(module = "core")
```
srossbach: The annotation is useless. We do not produce component diagrams since years.

core/src/saros/git/GitManager.java
```diff
@@ -0,0 +1,121 @@
...
+import saros.session.AbstractActivityProducer;
+import saros.session.IActivityConsumer;
+import saros.session.IActivityConsumer.Priority;
+import saros.session.internal.SarosSession;
+
+/**
+ * This Class is responsible for the negotiation between user that want to sent,receive and fetch
+ * commits.
+ *
+ * <p>In order to start the negotiation the user needs to call the procedure {@code
```
srossbach: Functions are called methods in OOP and not procedure.

core/src/saros/git/GitManager.java
```diff
@@ -0,0 +1,121 @@
...
+import saros.activities.GitSendBundleActivity;
+import saros.annotations.Component;
+import saros.session.AbstractActivityConsumer;
+import saros.session.AbstractActivityProducer;
+import saros.session.IActivityConsumer;
+import saros.session.IActivityConsumer.Priority;
+import saros.session.internal.SarosSession;
+
+/**
+ * This Class is responsible for the negotiation between user that want to sent,receive and fetch
```
srossbach: a user
send, receive, and fetch

core/src/saros/git/GitManager.java
```diff
@@ -0,0 +1,121 @@
...
+      this.jGitFacade.setWorkDirTree(workDirTree);
+    }
+  }
+
+  private final IActivityConsumer consumer =
+      new AbstractActivityConsumer() {
+        @Override
+        public void receive(GitRequestActivity activity) {
+          if (jGitFacade != null) {
+            log.info("You recieved a request to fetch commits");
```
srossbach: debug and honestly, imagine you read such a sentence in a log file ...

core/src/saros/git/GitManager.java
```diff
@@ -0,0 +1,121 @@
...
+            try {
+
+              String basis = jGitFacade.getSHA1HashByRevisionString("HEAD");
```
srossbach: do not use magic numbers ... aassing "HEAD" to a string variable.

core/src/saros/git/GitManager.java
```diff
@@ -0,0 +1,121 @@
...
+              fireActivity(new GitCollectActivity(session.getLocalUser(), basis));
+
+            } catch (IOException e) {
+              log.debug(session.getLocalUser().toString() + " can't access the Git Directory");
```
srossbach: toString() is always invoked when concatenating stings while using objects

srossbach: Are you sure to swallow the stack trace of the exception and log this a debug ?

core/src/saros/git/GitManager.java
```diff
@@ -0,0 +1,121 @@
...
+            }
+          } else {
+            log.info("Your workDirTree is not selected.");
+          }
+        }
+
+        @Override
+        public void receive(GitCollectActivity activity) {
+          if (jGitFacade != null) {
```
srossbach: if (jGitFacade == null) { .... ; return;}

core/src/saros/session/internal/SarosSession.java
```diff
@@ -1038,6 +1042,18 @@ public void changeColor(int colorID) {
     changeColorManager.changeColorID(colorID);
   }
 
+  public void gitSendCommitRequest() {
```
srossbach: The methods do not belong here.
1. They are GIT specific and not generic VCS.
2. What do you think will happen if every new component adds method to this class ?

core/test/junit/saros/git/GitManagerTest.java
```diff
@@ -0,0 +1,239 @@
...
+      @Override
+      public void created(IActivity activity) {
+        if (activity instanceof GitRequestActivity) {
+          target.exec(rewriteGitRequestActivity((GitRequestActivity) activity));
+        } else if (activity instanceof GitCollectActivity) {
+          target.exec(rewriteGitCollectActivity((GitCollectActivity) activity));
+        } else if (activity instanceof GitSendBundleActivity) {
+          target.exec(rewriteGitSendBundleActivity((GitSendBundleActivity) activity));
+        } else {
+          return;
```
srossbach: else branch is not needed

core/test/junit/saros/git/GitManagerTest.java
```diff
@@ -0,0 +1,239 @@
...
+
+  @Test
+  public void testCreation() {
+    GitManager gitManager = new GitManager(alicesSession);
+    gitManager.start();
+    gitManager.stop();
+    EasyMock.verify(alicesSession);
+  }
+
+  @Test
```
srossbach: This test does not test anything.

core/src/saros/git/GitManager.java
chrismotian: SHA1 is the name of the hash-"type" or algorithm.
Could you tell me in other words what is the requested change in "aassing "HEAD" to a string variable." ?

core/src/saros/session/internal/SarosSession.java
chrismotian: > 1. They are GIT specific and not generic VCS.

The support I implement is indeed GIT specific and not generic VCS. 

> 2.What do you think will happen if every new component adds method to this class ?

Is there another/better place to add the needed methods? I will need them when I want to give the user access to the commands "send commit" and "change working directory tree".

core/test/junit/saros/git/GitManagerTest.java
chrismotian: To write a test for a manager I get inspired by [StopManagerTest.java](https://github.com/saros-project/saros/blob/master/core/test/junit/saros/synchronize/StopManagerTest.java). I used the test I have seen there to assert in my test `Assert.assertEquals(alicesBob, activity.getSource());`

core/src/saros/git/GitManager.java
chrismotian: done

core/src/saros/git/GitManager.java
chrismotian: done

core/src/saros/git/GitManager.java
chrismotian: done

core/src/saros/git/GitManager.java
chrismotian: done

core/src/saros/git/GitManager.java
chrismotian: done

core/src/saros/git/GitManager.java
chrismotian: done

core/test/junit/saros/git/GitManagerTest.java
chrismotian: done

core/src/saros/git/GitManager.java
```diff
@@ -0,0 +1,108 @@
...
+    } else {
+      this.jGitFacade.setWorkDirTree(workDirTree);
+    }
+  }
+
+  private final IActivityConsumer consumer =
+      new AbstractActivityConsumer() {
+        @Override
+        public void receive(GitRequestActivity activity) {
+          if (jGitFacade == null) return;
```
stefaus: what about adding the ActivityConsumer after creating the facade and removing this checks?

core/src/saros/git/GitManager.java
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * <p>Before starting negotiation the user need to set up his working directory tree by calling the
+ * method {@code changeWorkDirTree(File workDirTree)}. After calling it the first time the Manager
+ * hold a {@link JGitFacade} object.
+ */
+public class GitManager extends AbstractActivityProducer implements Startable {
+
+  private static final Logger log = Logger.getLogger(GitManager.class.getName());
+
+  private final SarosSession session;
```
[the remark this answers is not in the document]
srossbach: This is a leftover after I told him to not add the methods to the interface. I expected it would be clear that this also means that the methods will not appear in the SarosSession class ... but this is the current result.

core/src/saros/git/GitManager.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private final IActivityConsumer consumer =
+      new AbstractActivityConsumer() {
+        @Override
+        public void receive(GitRequestActivity activity) {
+          if (jGitFacade == null) return;
+
+          try {
+            String basis = jGitFacade.getSHA1HashByRevisionString("HEAD");
```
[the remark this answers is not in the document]
chrismotian: I found two problems with your proposed solution:

1. If I understand it right the return value of `jGitFacade.getHeadObjectId()` is an ObjectId and this is a JGit class. In order to hide all the JGit classes by the facade, I don't want to use it in the manager. 
2. My solution is more general. Creating a method just for the HEAD instead of using the generic version seems to be redundant.

core/src/saros/git/JGitFacade.java
```diff
@@ -0,0 +1,192 @@
...
+   * Set the {@code git} object which offer the functionality of the library. See {@link Git}
+   *
+   * @throws IllegalArgumentException {@code workDirTree} is null or can't be resolved
+   * @throws IOException failed while read from workDirTree
+   */
+  private void setGit() throws IOException {
+    try {
+      if (workDirTree == null)
+        throw new IllegalArgumentException("workDirTree is null and can't be resolved");
+      git = Git.open(workDirTree);
```
stefaus: `git.close()` is never called.

core/src/saros/git/JGitFacade.java
```diff
@@ -0,0 +1,192 @@
...
+      ObjectId basisID = git.getRepository().resolve(basis);
+
+      if (basisID == null) throw new IOException("failed while resolved basis");
+
+      try {
+        RevWalk walk = new RevWalk(repo);
+
+        RevCommit basisCommit = walk.parseCommit(basisID);
+
+        walk.close();
```
stefaus: `close` is maybe never called, please use  try-with-resources for `RevWalk walk`

PR #449, opened 2019-02-21
Move plugin META-INF into resource directory and add plugin icon

tobous:
#### [I] Move plugin.xml to resources folder

Move the META-INF folder defining the plugin structure into the
resources directory to match the normal structure used by IntelliJ
plugins. Furthermore, this allows the IntelliJ code inspection to
recognize the resources used in the plugin.xml file correctly.

#### [I] Add plugin icon

Adds a plugin icon which will be displayed on the update site.


tobous:
The used logo was taken from the [saros-project/saros-graphics repo](https://github.com/saros-project/saros-graphics/blob/master/logo.svg). The IntelliJ code inspection still has some problems with the svg file that I haven't looked at in detail. Will do so in the next couple of days.

m273d15:
Did you tested that  the build process still produces an installable jar? I would expect that you have to adjust the project specific build.gradle. At least the following entry should be adjusted/removed:
```groovy
jar {
...
  into ('META-INF')
  {
    from 'META-INF/plugin.xml'
  }
}
```

tobous:
Good point. No, I did not check it. Will do.

I still have a general question about the build process. We have multiple build jobs (`build.jar` and `delivery.sarosIntelliJ`) as well as multiple build directories (`build/distribution/intellij/` and `de.fu_berlin.inf.dpp.intellij/build/distribution/`). What is the relevance of the different jobs and directories?

tobous:
From what I can tell, it does not make any difference whether I remove the section from the build script or not. It builds with or without it.

The only adjustment that is needed is that now that the plugin.xml is located correctly, the gradle-intellij-plugin mechanism to set the build compatibility is used (as it is enabled by default). See flag `updateSinceUntilBuild` on the [plugin info page](https://github.com/JetBrains/gradle-intellij-plugin#setup-dsl).

We have to decide whether we want to set compatibility in the build file or the plugin.xml and adjust the gradle setup accordingly (by either setting the corect values or disabling the functionality). The values of the plugin.xml that can be modified by the gradle build are described [here](https://github.com/JetBrains/gradle-intellij-plugin#patching-dsl).

tobous:
I decided to disable the feature to set the version compatibility as part of the gradle build for now. If we at some point decide that we want to use it, we can explicitly enable it and set the correct versions in a separate PR.

de.fu_berlin.inf.dpp.intellij/resources/META-INF/pluginIcon.svg
```diff
@@ -0,0 +1,117 @@
...
+       y1="186.6479"
+       x2="347.14285"
+       y2="186.6479"
+       id="linearGradient3808"
+       xlink:href="#linearGradient3802"
+       gradientUnits="userSpaceOnUse" />
+  </defs>
+  <metadata
+     id="metadata7">
+    <rdf:RDF>
```
tobous: IntelliJ has a problem with these tags that are probably related to it not loading the first three schema defined in lines 5-7. But even after loading these schemata (line 6 could not be loaded), this section could not be parsed correctly.

This does not seem to influence the graphic at all, so the decision now is whether to keep it as it is, improve it or to remove the metadata section and its schemata.

But since the metadata section does not seem to contain any information of value (even the title or description is left blank) I would suggest we either remove it or put in some information (at least the "title" and "about" section). I am not sure how useful this metadata is.

PR #451, opened 2019-02-22
Prepare to remove project from plugin context (1 of X)

tobous:
This PR starts to move project dependent components to the session context.

This is done in preparation of an upcoming change that would no longer make the project object accessible in the plugin context.

Moving the project into the session context is needed when moving Saros from a project plugin to an application plugin as we can no longer determine which project to run on during the plugin initialization. The project-module combination will be chosen by the user as part of the session start, meaning the project object will only be available in the session context.

### DIFF
--- a/intellij/src/saros/core/project/internal/SarosIntellijSessionContextFactory.java
+++ b/intellij/src/saros/core/project/internal/SarosIntellijSessionContextFactory.java
@@ -2,6 +2,7 @@ package saros.core.project.internal;
 
 import saros.intellij.editor.LocalEditorHandler;
 import saros.intellij.editor.LocalEditorManipulator;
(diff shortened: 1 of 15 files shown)


tobous:
Rebased onto current version of base PR.

tobous:
Rebased onto the base PR without any interaction.

tobous:
Rebased onto current master. Fixed a lot of conflicts dealing with imports due to the recent refactoring of the package structure.

tobous:

tobous:

tobous:

tobous:
Rebased onto current master with minimal interaction to fix merge-conflicts dealing with import statements.

PR #452, opened 2019-02-22
Prepare to remove project from plugin context (2 of X)

tobous:
Continuation of #451.

Splits off document handling from ProjectAPI and moves ProjectAPI to session context.

PR #453, opened 2019-02-24
Prepare to remove project from plugin context (3 of X)

tobous:
Continuation of #451 and #452.

Adds a utility to obtains a valid project object for UI purposes. Replaces the plugin context project access of the UI components with calls to the newly added `UIProjectUtils`.

The commits of this PR should be reviewed separately.

### Commits
<details>
<summary>
Click here to show all commit messages.
</summary>

#### [NOP][I] Remove unused dialog methods

Removes unused methods to display different types of dialogs.

#### [INTERNAL][I] Use toolbar project object for UI components

Adjusts the UI logic related to the toolbar and its contained components
to use the project object passed to the toolbar during its
initialization. This allows the UI components to always use the correct
project object as the reference point when creating new UI components
like dialogs.

Subsequently adjusts the dialog utility classes to accept a project
object as an additional parameter instead of always using the project
object contained in the plugin context.

As DialogUtils.showConfirmationDialog(...) is still used by components
that currently can't access a project object (besides the one held in
the plugin context), the method is kept as oldShowConfirmationDialog.
The method is marked as deprecated and will be removed once all required
components have access to a project object.

#### [INTERNAL][I] Add UIProjectUtils

Adds a utility class that provides access to current project objects to
UI components.

#### [INTERNAL][I] Pass project to CollaborationUtils.leaveSession(...)

Adjusts CollaborationUtils.leaveSession(...) get the used project object
as a parameter instead of getting it from the plugin context.

Adjusts LeaveSessionButton to use the toolbar project object for the
call.

Adjusts IntelliJCollaborationUtilsImpl to use
UIProjectUtils.runWithProject(...) to obtain a project object.

#### [INTERNAL][I] Use UIProjectUtils in XMPPAuthorizationHandler

Replaces the usage of the project object contained in the session
context by using UIProjectUtils.runWithProject(...) to obtain a valid
project object.

Subsequently removes DialogUtils.oldShowConfirm(...) as it is no longer
used.

#### [INTERNAL][I] Use UIProjectUtils for incoming invitation wizards

Adjusts the Wizard base class to get the used project object as a CTOR
parameter instead of using the project object contained in the plugin
context.

Uses UIProjectUtils in NegotiationHandler to obtain a project object to
run the JoinSessionWizard and AddProjectToSessionWizard with.

#### [INTERNAL][I] Use UIProjectUtils in ProgressFrame

Uses UIProjectUtils in ProgressFrame to set the correct position of the
created progress frame without using the project object contained in the
plugin context.

#### [INTERNAL][I] Use UIProjectUtils in IntelliJDialogManager

Uses UIProjectUtils in IntelliJDialogManager to set the correct position
of the created dialog without using the project object contained in the
plugin context.

#### [INTERNAL][I] Adjust NotificationPanel to use the shared project

Adds the method UIProjectUtils.getSharedProject() which returns the
project object contained in the current session context if present. This
method is used to determine the project object to use in Notification
panel.

With this logic, notifications created through the NotificationPanel
will be displayed in all projects that are part of the session or all
open projects if there is no running session.

#### [REFACTOR][I] Do minor refactoring

Does some minor refactoring in the classes changed by the last few
commits.

Replaces anonymous inner class declarations with lambda expressions or
method references.

Reduces the visibility of classes, CTORs, methods, and parameters to the
required minimum.

Adds annotations to ignore listeners that could be declared inline.
These listeners are kept as explicit declarations in the class body to
improve readability.

Fixes typos. This includes the spelling of IntelliJ (ending on a capital
J), as such a camel case is not correctly recognized by the Intellij
code inspection.

Removes unnecessary generic type declarations when initializing a
generic data type.

Adds NotNull annotations that are implicitly inherited from the super-
class.

</details>

### DIFF
--- a/intellij/src/saros/core/ui/eventhandler/XMPPAuthorizationHandler.java
+++ b/intellij/src/saros/core/ui/eventhandler/XMPPAuthorizationHandler.java
@@ -1,10 +1,12 @@
 package saros.core.ui.eventhandler;
 
 import com.intellij.openapi.application.ApplicationManager;
+import com.intellij.openapi.project.Project;
 import java.text.MessageFormat;
 import org.apache.log4j.Logger;
 import saros.intellij.ui.Messages;
 import saros.intellij.ui.util.DialogUtils;
+import saros.intellij.ui.util.UIProjectUtils;
 import saros.net.xmpp.JID;
 import saros.net.xmpp.subscription.SubscriptionHandler;
 import saros.net.xmpp.subscription.SubscriptionListener;
@@ -13,34 +15,42 @@ import saros.net.xmpp.subscription.SubscriptionListener;
 public class XMPPAuthorizationHandler {
 
   private static final Logger LOG = Logger.getLogger(XMPPAuthorizationHandler.class);
+
+  private final UIProjectUtils projectUtils;
+
   private final SubscriptionListener subscriptionListener =
       new SubscriptionListener() {
 
         @Override
         public void subscriptionRequestReceived(final JID jid) {
 
-          ApplicationManager.getApplication()
-              .invokeLater(
-                  new Runnable() {
-                    @Override
-                    public void run() {
-                      handleAuthorizationRequest(jid);
-                    }
-                  });
+          projectUtils.runWithProject(
+              project ->
+                  ApplicationManager.getApplication()
+                      .invokeLater(
+                          new Runnable() {
+                            @Override
+                            public void run() {
+                              handleAuthorizationRequest(project, jid);
+                            }
+                          }));
         }
       };
   private final SubscriptionHandler subscriptionHandler;
 
-  public XMPPAuthorizationHandler(final SubscriptionHandler subscriptionHandler) {
+  public XMPPAuthorizationHandler(
+      final SubscriptionHandler subscriptionHandler, UIProjectUtils projectUtils) {
     this.subscriptionHandler = subscriptionHandler;
     this.subscriptionHandler.addSubscriptionListener(subscriptionListener);
+
+    this.projectUtils = projectUtils;
   }
 
-  private void handleAuthorizationRequest(final JID jid) {
+  private void handleAuthorizationRequest(Project project, final JID jid) {
 
     boolean accept =
-        DialogUtils.oldShowConfirm(
-            null,
+        DialogUtils.showConfirm(
+            project,
             Messages.SubscriptionManager_incoming_subscription_request_title,
             MessageFormat.format(
                 Messages.SubscriptionManager_incoming_subscription_request_message,
--- a/intellij/src/saros/core/ui/eventhandler/XMPPAuthorizationHandler.java
+++ b/intellij/src/saros/core/ui/eventhandler/XMPPAuthorizationHandler.java
@@ -3,7 +3,6 @@ package saros.core.ui.eventhandler;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.project.Project;
 import java.text.MessageFormat;
-import org.apache.log4j.Logger;
 import saros.intellij.ui.Messages;
 import saros.intellij.ui.util.DialogUtils;
 import saros.intellij.ui.util.UIProjectUtils;
@@ -13,11 +12,9 @@ import saros.net.xmpp.subscription.SubscriptionListener;
 
 /** Handler for accepting or rejecting incoming XMPP subscription requests */
 public class XMPPAuthorizationHandler {
-
-  private static final Logger LOG = Logger.getLogger(XMPPAuthorizationHandler.class);
-
   private final UIProjectUtils projectUtils;
 
+  @SuppressWarnings("FieldCanBeLocal")
   private final SubscriptionListener subscriptionListener =
       new SubscriptionListener() {
 
@@ -27,15 +24,10 @@ public class XMPPAuthorizationHandler {
           projectUtils.runWithProject(
               project ->
                   ApplicationManager.getApplication()
-                      .invokeLater(
-                          new Runnable() {
-                            @Override
-                            public void run() {
-                              handleAuthorizationRequest(project, jid);
-                            }
-                          }));
+                      .invokeLater(() -> handleAuthorizationRequest(project, jid)));
         }
       };
+
   private final SubscriptionHandler subscriptionHandler;
 
   public XMPPAuthorizationHandler(
--- a/intellij/src/saros/core/ui/eventhandler/XMPPAuthorizationHandler.java
+++ b/intellij/src/saros/core/ui/eventhandler/XMPPAuthorizationHandler.java
@@ -39,7 +39,7 @@ public class XMPPAuthorizationHandler {
   private void handleAuthorizationRequest(final JID jid) {
 
     boolean accept =
-        DialogUtils.showConfirm(
+        DialogUtils.oldShowConfirm(
             null,
             Messages.SubscriptionManager_incoming_subscription_request_title,
             MessageFormat.format(
--- a/intellij/src/saros/intellij/ui/util/SafeDialogUtils.java
+++ b/intellij/src/saros/intellij/ui/util/SafeDialogUtils.java
@@ -65,15 +65,12 @@ public class SafeDialogUtils {
     final AtomicReference<String> response = new AtomicReference<>();
 
     application.invokeAndWait(
-        new Runnable() {
-          @Override
-          public void run() {
-            String option =
-                Messages.showInputDialog(
-                    project, message, title, Messages.getQuestionIcon(), initialValue, null);
-            if (option != null) {
-              response.set(option);
-            }
+        () -> {
+          String option =
+              Messages.showInputDialog(
+                  project, message, title, Messages.getQuestionIcon(), initialValue, null);
+          if (option != null) {
+            response.set(option);
           }
         },
         ModalityState.defaultModalityState());
@@ -92,12 +89,7 @@ public class SafeDialogUtils {
     LOG.info("Showing error dialog: " + title + " - " + message);
 
     application.invokeLater(
-        new Runnable() {
-          @Override
-          public void run() {
-            Messages.showErrorDialog(project, message, title);
-          }
-        },
+        () -> Messages.showErrorDialog(project, message, title),
         ModalityState.defaultModalityState());
   }
 
--- a/intellij/src/saros/intellij/ui/util/SafeDialogUtils.java
+++ b/intellij/src/saros/intellij/ui/util/SafeDialogUtils.java
@@ -9,7 +9,6 @@ import java.util.concurrent.atomic.AtomicReference;
 import org.apache.log4j.Logger;
 import saros.SarosPluginContext;
 import saros.exceptions.IllegalAWTContextException;
-import saros.repackaged.picocontainer.annotations.Inject;
 
 /**
  * Dialog helper used to show messages in safe manner by starting it on the AWT event dispatcher
@@ -29,8 +28,6 @@ public class SafeDialogUtils {
 
   private static final Application application;
 
-  @Inject private static Project project;
-
   static {
     application = ApplicationManager.getApplication();
 
@@ -40,9 +37,14 @@ public class SafeDialogUtils {
   private SafeDialogUtils() {}
 
   /**
-   * Shows an input dialog. This method must not be called from a write safe context as it needs to
-   * be executed synchronously and AWT actions are not allowed from a write safe context.
+   * Synchronously shows an input dialog. This method must not be called from a write safe context
+   * as it needs to be executed synchronously and AWT actions are not allowed from a write safe
+   * context.
    *
+   * @param project the project used as a reference to generate and position the dialog
+   * @param message the text displayed as the message of the dialog
+   * @param initialValue the initial value contained in the text field of the input dialog
+   * @param title the text displayed as the title of the dialog
    * @return the <code>String</code> entered by the user or <code>null</code> if the dialog did not
    *     finish with the exit code 0 (it was not closed by pressing the "OK" button)
    * @throws IllegalAWTContextException if the calling thread is currently inside a write safe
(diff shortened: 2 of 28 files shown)


de.fu_berlin.inf.dpp.intellij/src/de/fu_berlin/inf/dpp/intellij/ui/util/UIActiveProjectProvider.java
```diff
@@ -0,0 +1,78 @@
...
+    if (currentSession != null) {
+      return sarosSession.getComponent(Project.class);
+    }
+
+    Promise<DataContext> promisedDataContext =
+        DataManager.getInstance().getDataContextFromFocusAsync();
+    Promise<Project> promisedProject = promisedDataContext.then(DataKeys.PROJECT::getData);
+
+    try {
+      return promisedProject.blockingGet(1000);
```
tobous: This should be replaced with a static final variable.

The maximum waiting period of one second (1000 milliseconds) was chosen purely by gut feeling (by guessing how long a UI action can be delayed without the user getting annoyed). This value seem to hold (I have only done minor testing) and the waiting does not seem to cause any noticeable slowdown (especially since this mechanism is only used while there is no session, where Saros does not have a lot of  UI interactions besides the Session negotiation windows).

tobous:
Rebased onto current version of the base PR.

Fixed some variable names.

tobous:
Moved timeout for project request to static final variable.

tobous:
Rebased onto the base PR without any interaction.

tobous:
Rebased onto current base PR. Fixed a lot of conflicts dealing with imports due to the recent refactoring of the package structure.

intellij/src/saros/intellij/ui/util/UIActiveProjectProvider.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * Returns a current project object.
+   *
+   * <p>If there is currently a session, the session project will be returned. Otherwise, the
+   * currently focused project will be requested and returned. In this case, <code>null</code> is
+   * possible.
+   *
+   * <p><b>NOTE:</b> The project returned by this method should only be used to for UI purposes.
```
srossbach: I guess this method should be similar to its Eclipse counterpart https://github.com/saros-project/saros/blob/master/eclipse/src/saros/ui/util/SWTUtils.java#L202

But I do not understand

1. The Saros Part (I mean it is for UI purposes, so why do fetch it from the session) ?
2. Doesn't the getDataContextFromFocusAsync always produces a timeout when called from the Swing/AWT dispatch thread ?

tobous: 1. In an upcoming change, I am going to move the project object (probably contained in a wrapper) to the session context. The held project will be initialized as part of the negotiation, where the user chooses a project module combination to share (instead just a module).
I would prefer to use this project object for UI purposes to ensure that the notification/dialogs are displayed in the project that is actually shared and not the focused project. (This should not matter for dialogs but is should matter for notifications as those are only shown in the specific project.)
2. I am not sure. It does seem to work in general for the UI components, but I have only done minor testing. But it's a good hint. Will look into it.

tobous: Yes, you were absolutely right. This method always times out when called from the AWT event dispatcher thread. The method internally calls [`doWhenFocusSettlesDown`](https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/openapi/wm/impl/FocusManagerImpl.java#L178), which executes asynchronously on the AWT event dispatcher thread.

Will move the project requests out of the dispatcher thread blocks and add an assertion ensuring that the method is not called from the dispatcher thread.

tobous: Ok, that does not work. Will have to think about this more.

tobous:
Wupsi. This version was not supposed to be pushed. I forgot that I already amended the commits and only figured out that the solution had problems afterwards. :sweat_smile: Ignore the changes made by the latest push.

tobous:

tobous:
I completely reworked the patch to move the UI components away from using the project object stored in the project context.

The new solution now partly avoids the problem by using the project object passed when the base toolbar is initialized by Intellij.
I solved  the issues with deadlocks when requesting the project object from the AWT dispatcher thread by making the implementation asynchronous.

intellij/src/saros/intellij/ui/util/UIProjectUtils.java
```diff
@@ -0,0 +1,105 @@
...
+   */
+  @Nullable
+  private Project getProjectFromSession() {
+    ISarosSession currentSession = sarosSession;
+
+    if (currentSession == null) {
+      return null;
+    }
+
+    return sarosSession.getComponent(Project.class);
```
tobous: Use `currentSession` object.

tobous:

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -386,14 +388,18 @@ public void canceled(final NegotiationTools.CancelLocation location, final Strin
    *
    * @param negotiation The IPN this wizard handles
    */
-  public AddProjectToSessionWizard(Window parent, AbstractIncomingProjectNegotiation negotiation) {
+  public AddProjectToSessionWizard(
```
m273d15: Update javadoc?

intellij/src/saros/intellij/ui/wizards/JoinSessionWizard.java
```diff
@@ -53,12 +54,14 @@ public void cancel() {
    *
    * @param negotiation The negotiation this wizard displays
    */
-  public JoinSessionWizard(Window parent, IncomingSessionNegotiation negotiation) {
+  public JoinSessionWizard(Project project, Window parent, IncomingSessionNegotiation negotiation) {
```
m273d15: update javadoc?

intellij/src/saros/intellij/ui/wizards/Wizard.java
```diff
@@ -71,17 +69,17 @@ public void actionPerformed(ActionEvent e) {
...
-
+  protected Project project;
   /**
    * Constructor creates wizard structure.
    *
    * @param title window title
    * @param headerPanel
    */
-  public Wizard(Window parent, String title, HeaderPanel headerPanel) {
+  public Wizard(Project project, Window parent, String title, HeaderPanel headerPanel) {
```
m273d15: update javadoc?

intellij/src/saros/intellij/ui/util/UIProjectUtils.java
```diff
@@ -0,0 +1,117 @@
...
+   * <p>If there is currently a session, the given method is run immediately using the project
+   * contained in the session context. Otherwise, the currently focused project is requested and the
+   * method is run asynchronously when the request is done. In this case, the used project object
+   * can be <code>null</code> if the request fails.
+   *
+   * <p><b>NOTE:</b> This method should only be used to for UI purposes. To interact with the
+   * project whose module is shared as part of a Saros session, please use the Project object
+   * contained in the session context.
+   */
+  public void runWithProject(@NotNull ProjectRunner projectRunner) {
```
m273d15: javadoc: `@param` description?

intellij/src/saros/intellij/ui/util/UIProjectUtils.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    Promise<DataContext> promisedDataContext =
+        DataManager.getInstance().getDataContextFromFocusAsync();
+    Promise<Project> promisedProject = promisedDataContext.then(DataKeys.PROJECT::getData);
+
+    promisedProject.onProcessed(projectRunner::run);
+  }
+
+  /** Interfaced used to define tasks that need a project to run. */
```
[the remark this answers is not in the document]
tobous: Done

intellij/src/saros/intellij/ui/util/SafeDialogUtils.java
```diff
@@ -129,15 +142,19 @@ public static String showPasswordDialog(final String message, final String title
...
+   * @param project the project used as a reference to generate and position the dialog
+   * @param message the text displayed as the message of the dialog
+   * @param title the text displayed as the title of the dialog
    * @return the value {@link Messages#YES} if "Yes" is chosen and {@link Messages#NO} if "No" is
    *     chosen or the dialog is closed
    * @throws IllegalAWTContextException if the calling thread is currently inside a write safe
    *     context
    */
-  public static Integer showYesNoDialog(final String message, final String title)
+  public static Integer showYesNoDialog(Project project, final String message, final String title)
```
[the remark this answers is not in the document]
tobous: The method just passes along the result of the IntelliJ dialog, which returns an Integer. The IntelliJ dialog only returns 1 as 'Yes' and 0 as 'No', so it is used as a boolean. But I don't think adjusting this is worth it. The current usage of the method just checks the result against the constants defined by IntelliJ.

intellij/src/saros/intellij/ui/util/UIProjectUtils.java
tobous: I use the additional variable to ensure that I always work on the same object. The variable `sarosSession` is volatile and can be changed asynchronously through the ISessionLifecycleListener defined in the beginning of the class when the session state changes (session starts/ends). If I would work on the volatile reference directly, it could happen that the session ends after my check but before I am working with the variable, which would lead to a NPE as the variable would have been set to `NULL` by the session listener in accordance with the session end.

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
tobous: Done

intellij/src/saros/intellij/ui/wizards/JoinSessionWizard.java
tobous: Done

intellij/src/saros/intellij/ui/wizards/Wizard.java
tobous: done

intellij/src/saros/intellij/ui/util/UIProjectUtils.java
tobous: done

tobous:
Adjusted PR according to review.

intellij/src/saros/core/ui/eventhandler/XMPPAuthorizationHandler.java
```diff
@@ -1,46 +1,48 @@
...
+
   private final SubscriptionHandler subscriptionHandler;
 
-  public XMPPAuthorizationHandler(final SubscriptionHandler subscriptionHandler) {
+  public XMPPAuthorizationHandler(
+      final SubscriptionHandler subscriptionHandler, UIProjectUtils projectUtils) {
     this.subscriptionHandler = subscriptionHandler;
     this.subscriptionHandler.addSubscriptionListener(subscriptionListener);
+
+    this.projectUtils = projectUtils;
```
stefaus: Even if its actual potential should be low, this could lead to a race condition:
If the `subscriptionListener` is triggered while `projectUtils` is not initialized / visible, it will fail. Reordering the statements won't help.

From the [Java Language Specification](
https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.5):
>An object is considered to be completely initialized when its constructor finishes. A thread that can only see a reference to an object after that object has been completely initialized is guaranteed to see the correctly initialized values for that object's final fields.


A Solution would be to make the `anonymous class` a `static final` one and move all references into it. Or register the Listener outside the constructor.

srossbach: You could also make the variable volatile and assign it first. But my guess is that this class is a singleton used inside the PicoContainer ? and so this variable is always initialized before the listener is triggered.

tobous: Yes, the XMPPAuthorizationHandler is instantiated by the PicoContainer plugin context. Furthermore, the handler reacts to incoming XMPP Subscription events, which require a running connection to an XMPP account.

I could let the class implement `Startable` and move the subscription to the `start()` call. But I am not sure whether this is really necessary. What are your opinions?

I have seen (and used) this pattern multiple times before, so if it is an issue, we should also aim to adjust all other places where listeners with dependencies on other instance variables are subscribed in the CTOR.

srossbach: The main container is not started at all because in former times some components reinjected themselves and so it was prohibited by PicoContainer to call start on such a container. I guess that can be changed now. But I do not know if it is worth the effort. And yes this pattern is used multiples times all over the code (registering listeners inside the CTOR).

tobous: > The main container is not started at all because in former times some components reinjected themselves and so it was prohibited by PicoContainer to call start on such a container.

Good to know. I will have to make sure that I don't rely on Startable in such cases then.

Will the PicoContainer just 'silently' not run the start/stop methods or will it complain about Startable components in the main container?

srossbach: IIRC the reinject fails if the container was already started. However we have no components now that want to reinject themselves. I think you have to play around a little bit and make sure that https://github.com/saros-project/saros/blob/master/core/src/saros/context/ContainerContext.java#L250 is still working if you called start on the main container.

stefaus: So this is not optimal, but as long as no other thread will call the listener during the startup (like a ui thread or something) it's ok.

> You could also make the variable volatile and assign it first. 

This is by Specification not a safe approach, because the anonymous class still gets a `this` reference to the incomplete object.

srossbach: Looking at the code I have my doubts that all final fields would not be written at the time you call addSubscriptionListener (assuming that this would be the last statement in the CTOR). But yes it would be only confusing to declare each variable as volatile to ensure that the callback would not see a half initialized class.

tobous: I think I will go ahead an merge this PR as the issue is not pressing in this case and a more general solution approach can be discussed in #534.

PR #457, opened 2019-02-26
[INTERNAL][I] Remove RemoteWriteAccessManager

tobous:
Removes RemoteWriteAccessManager. The class was copied over from Saros/E
but was never fully integrated into Saros/I, meaning it provided no
functionality in its  current form.

As it still has to be discussed whether Saros should handle write
permissions at all, the class will be removed for now. If it is decided
that such a functionality is wanted, it can be re-created and correctly
integrated or implemented from scratch.

Subsequently removes AutoHashMap as it was only used by
RemoteWriteAccessManager. Furthermore, the functionality of the
AutoHashMap can now be replaced through the usage of
Map.computeIfAbsent(...).


tobous:
Rebased onto current master. Fixed a lot of conflicts dealing with imports due to the recent refactoring of the package structure.

tobous:
Resolved a minor merge conflict in `EditorManager`.

Issue #421, opened 2019-02-04
Move eslint format check into the travis format checking stage


m273d15:
We introduced a stage that contains the format checks (currently only java check).
The javascript code is also checked but this is still located in the build scripts.
It is necessary to move the check into the other stage. The challenge is to install only the packages which are required for format checking (instead of installing all packages)

AbdBarho:
fixed #458

PR #458, opened 2019-02-28
Move eslint format check into the travis format checking stage

AbdBarho:
addresses the issue #421


travis/script/lint/lint_js.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+project_root_dir=$1
+
+echo "Linting the HTML GUI"
+node ./lint.js project_root_dir/ui.frontend/html
```
m273d15: This file has a few issues:
1. **Usage of relative path `./lint.js`**
    You defined in `travis.yml` that this script is executed from within in the project root (repository root dir) as "current work directory". The notation `./lint.js` always refers to a file in the current work directory. Therefore if this line is interpreted the following file is referenced:`<project root>/lint.js`. You want to reference `<project root>/travis/script/lint/lint.js` therefore use `$project_root_dir/travis/script/lint/lint.js`.
`node <project root>/lint.js project_root_dir/ui`
2. **Missing $ sign before variable**
  Instead of writing `project_root_dir/ui.frontend/html` you have to write `$project_root_dir/ui.frontend/html`

The correct file would be
```bash
#!/bin/bash

project_root_dir=$1

echo "Linting the HTML GUI"
node $project_root_dir/travis/script/lint/lint.js $project_root_dir/ui.frontend/html

echo "Linting the HTML Whiteboard"
node $project_root_dir/travis/script/lint/lint.js $project_root_dir/whiteboard/frontend
```

Btw. As far as I know it is possible to execute bash scripts with windows (with the git bash). In the git bash (with node) you could simply call `travis/script/lint/lint_js.sh $PWD` in the project root in order to test your script.

AbdBarho: Thank you for updating the permissions on the file, I knew the problem but I did not know that you can actually fix it through git bash (because in cmd or powershell it does not work).

I have updated the files to address your comment

PR #461, opened 2019-03-01
Adjust UI on host side of project negotiation to handle multiple open projects

tobous:
#### [INTERNAL][I] Add project selection to ContactPopMenu

Adjusts the ContactPopMenu to show all open modules grouped by their
project. This is needed to make the plugin context project independent.
Instead of setting the project at plugin initialization, we can instead
use the project selected by the user and set it in the session context.

#### [INTERNAL][I] Instantiate saros project object in ContactPopMenu

Replaces a call to IWorkspaceImpl.getProject(String) with directly
instantiating a new IntelliJProjectImpl object as the needed module
object is already present.

#### [INTERNAL][I] Move text shown to user to messages.properties

Moves text shown to user when interacting with the ContactPopMenu to the
messages.properties file.

#### [INTERNAL][I] Disable menu entry stating that no module was found

Disables the ContactPopMenu menu entry shown when Saros could not find
any (valid) module in the project the host is trying to share.

Otherwise, the user might get confused as it looks like they could still
interact with the menu entry, even though it is only meant as a way to
convey information.

### DIFF
--- a/intellij/src/saros/intellij/ui/tree/ContactPopMenu.java
+++ b/intellij/src/saros/intellij/ui/tree/ContactPopMenu.java
@@ -17,9 +17,11 @@ import javax.swing.JMenuItem;
 import javax.swing.JPopupMenu;
 import org.apache.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import saros.core.ui.util.CollaborationUtils;
 import saros.filesystem.IProject;
 import saros.filesystem.IResource;
+import saros.intellij.context.SharedIDEContext;
 import saros.intellij.filesystem.IntelliJProjectImpl;
 import saros.intellij.ui.Messages;
 import saros.intellij.ui.util.IconManager;
@@ -144,7 +146,7 @@ class ContactPopMenu extends JPopupMenu {
       }
 
       JMenuItem moduleItem = new JMenuItem(moduleName);
-      moduleItem.addActionListener(new ShareDirectoryAction(moduleName, wrappedModule));
+      moduleItem.addActionListener(new ShareDirectoryAction(project, moduleName, wrappedModule));
 
       shownModules.add(moduleItem);
     }
@@ -179,10 +181,14 @@ class ContactPopMenu extends JPopupMenu {
 
   /** Action that is executed, when a project is selected for sharing. */
   private class ShareDirectoryAction implements ActionListener {
+    private final Project project;
     private final String moduleName;
     private final IProject module;
 
-    private ShareDirectoryAction(String moduleName, IProject module) {
+    private ShareDirectoryAction(
+        @NotNull Project project, @NotNull String moduleName, @Nullable IProject module) {
+
+      this.project = project;
       this.moduleName = moduleName;
       this.module = module;
     }
@@ -213,7 +219,7 @@ class ContactPopMenu extends JPopupMenu {
       List<JID> contacts = new ArrayList<>();
       contacts.add(user);
 
-      // TODO set the correct project for the session context
+      SharedIDEContext.preregisterProject(project);
       CollaborationUtils.startSession(resources, contacts);
     }
   }
--- a/intellij/src/saros/intellij/ui/tree/ContactPopMenu.java
+++ b/intellij/src/saros/intellij/ui/tree/ContactPopMenu.java
@@ -163,11 +163,15 @@ class ContactPopMenu extends JPopupMenu {
                   : "complying with our current release restrictions ")
               + "were found");
 
-      projectMenu.add(
+      JMenuItem noModulesFoundMenuItem =
           new JMenuItem(
               nonCompliantModules.isEmpty()
                   ? Messages.ContactPopMenu_menu_entry_no_modules_found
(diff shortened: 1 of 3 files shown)


tobous:

tobous:

tobous:
Accidentally deleted the branch. The 'd' and 'f' keys are too close together.

tobous:
Rebased onto the current base PR without any interaction (except fixing a typo).

intellij/src/saros/intellij/ui/tree/ContactPopMenu.java
```diff
@@ -3,56 +3,87 @@
...
+  /**
+   * Prepares the JMenu for the given project. The menu contains all sharable modules contained in
+   * the project.
+   *
+   * @param project the project whose menu to prepare
+   * @return a <code>Pair</code> containing the prepared JMenu (first value) and a list of
+   *     incompatible modules (second value) or <code>null</code> if the ModuleManager could not be
+   *     instantiated
+   */
+  private Pair<JMenu, List<String>> prepareProjectMenu(@NotNull Project project) {
```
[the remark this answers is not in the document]
m273d15: @tobous  We already talked about codacy as an alternative to SonarCloud which also reviews "external" PRs. As discussed I enabled codacy, but for now I did not marked this check as "required before merge". I will check how we can configure our checks.

tobous:
Changed the target of the PR to master in order to separate the UI changes from the changes made to remove project dependencies from the plugin context.
This was done as these patches are mostly independent. Furthermore, these UI patches should only be merged once the project dependencies are correctly handled by the session context.

tobous:
Rebased onto current master with minimal interaction to fix merge-conflicts dealing with import statements.

intellij/src/saros/intellij/ui/tree/ContactPopMenu.java
```diff
@@ -26,33 +29,67 @@
...
   private final ContactTreeRootNode.ContactInfo contactInfo;
 
   ContactPopMenu(ContactTreeRootNode.ContactInfo contactInfo) {
     this.contactInfo = contactInfo;
 
-    if (workspace == null || project == null) {
-      SarosPluginContext.initComponent(this);
+    SarosPluginContext.initComponent(this);
+
+    JMenu menuShareProject = new JMenu("Work together on...");
```
m273d15: Why do you use a constant string here instead of storing the menu title in messages.properties?

intellij/src/saros/intellij/ui/tree/ContactPopMenu.java
```diff
@@ -26,33 +29,67 @@
...
+  /**
+   * Prepares the JMenu for the given project. The menu contains all sharable modules contained in
+   * the project.
+   *
+   * @param project the project whose menu to prepare
+   * @return a <code>Pair</code> containing the prepared JMenu (first value) and a list of
+   *     incompatible modules (second value) or <code>null</code> if the ModuleManager could not be
+   *     instantiated
+   */
+  private Pair<JMenu, List<String>> prepareProjectMenu(@NotNull Project project) {
```
m273d15: You don't have to do this now: Restructuring the method could improve the readability. The "happy" path is hard to find. I would recommend to split the method into multiple smaller methods.

srossbach:
From a UX perspective: I do not know how the Notification panel looks like, but getting "spammed" with possible issues when you right click on a contact is not the best experience. How about disabling the entries in the context menu and adding a tooltip for such modules describing the isssue ?

tobous:
@srossbach I was also not happy with the solution, but I couldn't think of a good alternative at that time. Notifications are pop-ups that are displayed for a time (I think 10 seconds) on the bottom right corner of the IDE and can be viewed in a notification log afterwards, so this is not *that* annoying, but it is still unnecessarily distracting.

Your approach is an improvement, but also has the drawback that the menu gets hard to navigate if there are multiple modules that can not be shared. Furthermore, I will have to think about whether to group the entries into "shareable" and "not shareable" or whether to still just sort them alphabetically (or both). (Any opinions? I am leaning towards grouping and sorting the groups alphabetically.)

Nevertheless, I think your approach still offers better usability than the current solution. I will create a separate PR for this as the adjustment is independent to what this PR is aiming to do. @m273d15 As I will have to adjust the menu creation logic anyways, I will include the restructuring of `prepareProjectMenu` in that PR.

intellij/src/saros/intellij/ui/tree/ContactPopMenu.java
tobous: It was implemented this way from the start and I just forgot about it. Fixed.

intellij/src/saros/intellij/ui/tree/ContactPopMenu.java
tobous: See my reply below; will be done in a separate PR.

tobous:
@srossbach @m273d15 I created the PR #541 to change the behavior when encountering non-shareable modules and also restructure the method `prepareProjectMenu`.

tobous:

tobous:
Adjusted logic to add shared project to SharedIDEContext on host side.

PR #464, opened 2019-03-05
Prepare Saros/E release infrastructure

m273d15:
* Use `lib/` entry instead of explicit list of libraries. This list is already maintained in the `MANIFEST.MF` and in the gradle build description.
* Add `saros.ui` to the saros feature


srossbach:
The lib directory also contains the whole source code of the libraries. Do you want to ship this too ? I do not think so.

m273d15:
The `lib` dir does not contain source code anymore. Some source jars (I think only `swt-browser-improved-src.zip`) are in the `libs` directory, but gradle only copies the "binary" jars to `lib`.

Issue #110, opened 2018-07-04
Switch to semantic versioning


m273d15:
Semantic versioning has one big advantage... the semantic.
It is possible to express whether the new release contains:
* just a few minor bug fixes or
* a totally reworked business logic.

We have to discuss:
- [ ] whether we want the new versioning scheme
-------------------------------------------------------------
If the answer of the previous question is 'yes' then discuss:
- [ ] which form of semantic versioning is applicable (e.g. [SemVers](https://semver.org/))
- [ ] what is the right date to switch the versioning scheme
- [ ] what is the current semantic version
- [ ] what are necessary steps to change the versioning scheme

PR #465, opened 2019-03-05
[BUILD] Change to new semantic version 15.0.0

m273d15:
The version number `15` is more or less arbitrary.
The new versions are:

| component | version |
|-----|----|
| eclipse | 15.0.0
| intellij | 0.1.0
| core | 0.1.0
| ui | 0.1.0
| whiteboard | 0.1.0

Intellij versioning as [proposed by tobous](https://github.com/saros-project/saros/pull/465#issuecomment-470280748). This means the `0.1.0` version of Intellij is not constant as of the internal components (core,ui,whiteboard) and will be increased.
#110


m273d15:
@srossbach Do you see any issues with this versioning/version?

tobous:
I would prefer to leave IntelliJ at `0.1.0` for the first alpha. For the first somewhat stable releases I would continue with `1.0.0`, `2.0.0`, etc.
Once we reach compatibility with Saros/E (and are at least close to a feature parity), we can think about jumping to the current major release number used by Saros/E.

But we will have to decide in general how we want to handle the versioning between different Saros "derivatives" (Saros/E, Saros/I, Saros/S).

When switching to a new general versioning scheme, we should also discuss whether we want to include the IDE type (Eclipse, Intellij, etc.) as part of the string used for the version compatibility check. This would be very useful to ensure that different IDE versions aren't detected as compatible only because the version schemes collided. Including the IDE allows us to specify it as part of the compatibility list.

As a format, I would suggest to just prepend the version string with the IDE followed by some kind of separator, like '/'. Modifying the logic would be quite simple as we can add the IDE string to the plugin context and just prepend it in the version manager. The suggested format would look like `eclipse/15.0.0` for the eclipse release.

To work with this new internal compatibility version, we would also have to adjust the version comparison algorithm in [VersionManager](https://github.com/saros-project/saros/blob/master/core/src/saros/versioning/VersionManager.java).

intellij/resources/META-INF/plugin.xml
```diff
@@ -12,7 +12,7 @@
         This is the first alpha version of Saros for IntelliJ. Use at your own risk.
     </change-notes>
 
-    <version>0.1.0</version>
+    <version>15.0.0.DEVEL</version>
```
tobous: Please leave Saros/I at `0.1.0` for the alpha release. Otherwise, this might create unnecessarily high expectations towards the available functionality/stability/quality for the used.

tobous:
Also, I am not that fond of the `DEVEL` suffix.

My guess would be that we are using this to ensure that development builds are not automatically compatible with the last release. Or am I wrong in this assumption? Is there another reason for the suffix?

If we want to keep it, I would prefer to use a different separator char for the suffix. It looks weird to switch from an integer to a char using the same separator. But that is only my subjective opinion. If we add the IDE version, we could use the same separator char that is used there and just parse the qualifier there (if it is present). This would avoid unnecessary additional work for switching this.

m273d15:
@tobous Ok, I adopted your proposal :+1: If we see the need of using an additional postfix during development we can still add it.

m273d15:
The server has no version and as long as we don't release it I see no need for a version.

stefaus:
The Server stores a Version to be compatible with Eclipse, so this has to change too.
https://github.com/saros-project/saros/blob/a7b9df1e840f8955bf819b76f4d5224d067004e4/server/src/saros/server/SarosServer.java#L18

For Consistency, whats about the unchanged Sonar Version Numbers?

srossbach:
Get me wrong but IIRC the last release already had a core component. What version was that component ?

m273d15:
@stefaus 
> The Server stores a Version to be compatible with Eclipse, so this has to change too.

I will adjust the version. Thank you :+1: 

> For Consistency, whats about the unchanged Sonar Version Numbers?

Changing the version would add a new tag in sonarqube. In fact this version change PR does not mark a new release it should just prepare the next release. Therefore I would change the sonarqube version after the release. Furthermore we think about a better integrated tool as [Codacy](https://app.codacy.com/project/kelvin.glass/saros/dashboard), but this is a different topic.

@srossbach 
> Get me wrong but IIRC the last release already had a core component. What version was that component ?

The versions of the last release are:
* `de.fu_berlin.inf.dpp.core` version 0.1.0
* `de.fu_berlin.inf.dpp` version 14.10.31
* `de.fu_berlin.inf.dpp.whiteboard` version 14.10.31

Do you ask because it is a problem to use the same version of a bundle in multiple releases? Does the eclipse update mechanism install only bundles with a new version?

I think for the next release this should not be a problem, because we renamed the bundles (to `saros.core` etc)., but it would be important for subsequent releases.

m273d15:
If nobody has objections I will rebase + merge this PR in the next week

PR #468, opened 2019-03-07
[BUILD] Rename feature,update -> eclipse.feature,update

m273d15:
The new project names make clear, that the projects
'feature' and 'update' are related to the eclipse plugin.
Everybody that is familiar with the concept of update-sites
and features is directly able to identify the corresponding projects.
Furthermore the name feature/update without the eclipse context
is to generic in the context of software development and causes confusion.


tobous:
This seems like a step in the right direction to me, but the new names are still a bit confusing. Is it at all possible to move the directories into the eclipse folder? It is weird to have them in the root directory as this would suggest to me that they have the same importance as the other directories in the repository root, which is definitely not the case. My guess is that for at least 90% of developers looking at our project, these directories and their content will be completely irrelevant as they don't contain any functionality but are rather a niche aspect of our build/deployment setup.

m273d15:
@tobous I rebased the branch and moved the feature/update directory in the eclipse directory.
I tested: building a feature / update site. It seems as if the whole PDE stuff works.

m273d15:
I tested the setup with eclipse neon:
* The setup process is not changed, because the eclipse configurations of the update/feature project are not generated by gradle.
* If you import the feature/update project into eclipse the visualization is as before. Both projects are shown on the same hierarchical level as the other projects. The PDE build work also.

PR #474, opened 2019-03-16
[INTERNAL][HTML] Update dependencies

AbdBarho:
a general clean up and update of the HTML GUI dependencies, with the required code modifications to fit the updated libraries.

A lot of useless deps have been removed, and the build process has been streamlined.

also the linting rules have been updated.


AbdBarho:
@m273d15  I have downgraded the version of mobx, it turns out that mobx uses internally Proxies which got pollyfilled but not well enough for it to function properly, the developers claim they will continue supporting the older versions.

I guess this wont be an issue when the JavaFx WebView is used, and I hope we can upgrade to latest version again.

PR #475, opened 2019-03-17
[FIX] Wrong error display for no saros support

stefaus:
When inviting a contact without Saros support, the user is shown a
confusing message about the contact JID not beeing a resource qualified
JID. This happens because the qualified JID is resolved later and
SarosSession.getUser() throws an IllegalArgumentException for bare JIDs
which is overriding the orginal error message about the contact not
having saros support. This easy fix adds a check to avoid the IAE.


chrismotian:
I want to test it. Do you have a JID account without Saros ?

srossbach:
You can just use an IM of your choice.

core/src/saros/negotiation/OutgoingSessionNegotiation.java
```diff
@@ -85,9 +85,11 @@ public OutgoingSessionNegotiation( //
 
   @Override
   protected void executeCancellation() {
-    final User user = sarosSession.getUser(getPeer());
+    if (getPeer().isResourceQualifiedJID()) {
+      User user = sarosSession.getUser(getPeer());
```
srossbach: What is the point of removing the final keyword ? I normally prefix every variable with final if this variable should not be changed later.

stefaus:
I think the real fix would be to cage the JIDs in the net package and handle XMPP stuff there, but for now this trivial small fix does the trick.

>  forbid starting a negotiation with a non FQ JID.

Then I would have to move the blocking `DiscoveryManager.getSupportingPresence()` call into `SarosSessionManager.invite()` and change all the code acompying it.

core/src/saros/negotiation/OutgoingSessionNegotiation.java
stefaus: No offense but I think it is just clutter in this case, because there is no further meaning in marking a user explicitly final within 5 lines of code.

srossbach:
DiscoveryManager.getSupportingPresence() is not needed. The UI already have all information. However I looked at the code and the UI does not retrieve the resource part.  I can fix this in Eclipse. But we must ask @tobous how IntelliJ is managing the JID stuff in the UI.

srossbach:
@m273d15 do you really want to start adding logic to the core that has to distinguish between a JID and a FQJID ? Maybe it is the only place but this still feels weird, especially the JID replacment in the OPN.

See https://github.com/saros-project/saros/pull/477

I forgot to add you as reviewer.

stefaus:
> do you really want to start adding logic to the core that has to distinguish between a JID and a FQJID ?

I think this is a misleading comment, because the changed class already handles different JIDs (see code snippet of the existing "JID replacement") and nothing else of the core is touched.
https://github.com/saros-project/saros/blob/f25ba2e0564b46158fc86ad56d1970b4f723d898/core/src/saros/negotiation/OutgoingSessionNegotiation.java#L216-L227

Your approach in #477 adds more XMPP / JID code to the GUI like `roster.getPresences(...)` which is (as talked about in the Stand-Ups) not the way to go, because network related code should be handled in core. In addition, the core API should be able to respond correctly to exceptions, and we do not fix this by simply not calling these states.

To a certain extent I understand the discomfort with all this JID handling, but a major overhaul of these classes won't be done in the coming weeks, so please keep in mind that this little check now fixes it sufficiently.

srossbach:
Can you at least add a comment that points to line 226/227 stating that this comparison should not be necessary ?

And I do not understand why the UI or at least the part of the UI that is calling the Core should not use some logic offered by the Core to obtain the FQJID ? When you work in the UI it is nice that you can basically pass "broadcast addresses" into the Core passing the responsibility to someone else to retrieve the correct address.

I still not like the current state of the UI. Imagine you are off your PC and logged into the XMPP server with your mobile phone sitting in a meeting. Everyone can invite you. However you are not able to reject or accept the invitation because it is send to your PC. And the inviting side cannot even notice that because it will only see one contact.

stefaus:
> Can you at least add a comment that points to line 226/227 stating that this comparison should not be necessary ?

This comparison is necessary from the creation of a `OutgoingSessionNegotiation` object till updated by `setPeer()` in Line 227. I think a comment on this behavior could be outdated after changes and be confusing.

> And I do not understand why the UI or at least the part of the UI that is calling the Core should not use some logic offered by the Core to obtain the FQJID ?

What is your use-case for the FQJID in the UI and why shouldn't this be handled in the core?
Now the UI is 
1. asking the roster (which should be replaced by a abstraction) for all the contact / group entries
2. asking for all the presences / resources of a contact to get the FQJID (only a JID if offline)
3. ask the core if the (FQ)JIDs have Saros support
4. create entries and than filter for the one with saros support (otherwise just the available)

When we move all this logic to the core, all UI impls can use the same logic and we can create a better abstraction of the Smack Framework / XMPP.

> I still not like the current state of the UI. Imagine you are off your PC and logged into the XMPP server with your mobile phone sitting in a meeting. Everyone can invite you. However you are not able to reject or accept the invitation because it is send to your PC. And the inviting side cannot even notice that because it will only see one contact.

Neither do I, but adding logic to the handling should be done in the core.

So you're okay with committing this? We could also move the discussion to #477 or another issue if necessary.

srossbach:
> 
> 
> > Can you at least add a comment that points to line 226/227 stating that this comparison should not be necessary ?
> 
> This comparison is necessary from the creation of a `OutgoingSessionNegotiation` object till updated by `setPeer()` in Line 227. I think a comment on this behavior could be outdated after changes and be confusing.
> 
The comparison is NOT necessary IFF you can ensure that the OPN is already created with a FQJID. I do not get the reason not to add a comment to your changed lines. Of course the comment will be outdated at a give time.

> > And I do not understand why the UI or at least the part of the UI that is calling the Core should not use some logic offered by the Core to obtain the FQJID ?
> 
> What is your use-case for the FQJID in the UI and why shouldn't this be handled in the core?
> Now the UI is
> 
>     1. asking the roster (which should be replaced by a abstraction) for all the contact / group entries
> 
>     2. asking for all the presences / resources of a contact to get the FQJID (only a JID if offline)
> 
>     3. ask the core if the (FQ)JIDs have Saros support
> 
>     4. create entries and than filter for the one with saros support (otherwise just the available)
> 
> 
> When we move all this logic to the core, all UI impls can use the same logic and we can create a better abstraction of the Smack Framework / XMPP.
> 

What is the point of moving all of this logic to the Core package when there is already an UI package ? Obtaining the FQJID for a given (plain) JID with Saros support is a single call to the discovery manager.

stefaus:
Ok, I will not discuss further if the core should be able to respond to incorrect input.

The point is that this is bad architecture.
We have the Eclipse, IntelliJ, HTML UI and how knows if we will get others and it makes no sense to implement the same control logic in all of them.

PR #476, opened 2019-03-19
[INTERNAL][HTML] Add a context menu

AbdBarho:
the menu is currently just a container with some elements for testing purposes, but will be expanded later.

in the patch two new components have been added: `FocusableItem` and `FocusableMenu` which provide the ability to control the focus with the mouse as well as with the keyboard.
these components have been used in the `Running Session` component to make it also more accessible.

also some stylistic changes

### DIFF
--- a/ui.frontend/html/src/components/ContextMenu/Controller.jsx
+++ b/ui.frontend/html/src/components/ContextMenu/Controller.jsx
@@ -1,0 +1,123 @@
+import FocusableItem from '../Focusable/FocusableItem';
+import FocusableMenu from '../Focusable/FocusableMenu';
+import React from 'react';
+import { Text } from 'react-localize';
+
+
+export default class ContextMenuController {
+  constructor(core) {
+    this.core = core;
+    // reference the actual context menu component
+    this.ctxMenu = null;
+    // setup event listeners
+    // disable original context menu
+    document.addEventListener('contextmenu', e => e.preventDefault());
+    // hide menu on any click
+    document.addEventListener('click', () => this.ctxMenu.hide());
+  }
+
+  /**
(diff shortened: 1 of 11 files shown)


ui.frontend/html/src/components/ContextMenu/Controller.jsx
```diff
@@ -0,0 +1,123 @@
...
+  /**
+   * shows a menu when a contact has been right-clicked
+   * @param {string} jid jid of the contact that has been clicked
+   * @param {any} source the element which would recieve focus after the menu is closed
+   * @param {MouseEvent} event used to set position of the menu
+   */
+  showContactMenu(jid, source, event) {
+    this.ctxMenu.setContent(
+      <this.Menu source={source}>
+        {/* react does not support nested dropdowns, we change the contents again*/}
```
m273d15: Probably we find a better solution in the future. I think updating the current context menu is not very intuitive.

AbdBarho: agree, I do not have a clear idea about everything that should go in the context menu, that is why I decided to start like this, so it is really easy to change what the menu shows.

ui.frontend/html/src/template.ejs
```diff
@@ -1,13 +1,18 @@
...
-    <!-- initial page gets injected here -->
-    <script>
-      window.initialPage = '<%= htmlWebpackPlugin.options.title %>'
-    </script>
-    <div id="root"></div>
-  </body>
-</html>
\ No newline at end of file
+
+<head>
```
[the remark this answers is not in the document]
AbdBarho: it was not intended, my editor auto-formatted it, I changed it back

PR #477, opened 2019-03-20
[UI][E] Prevent inviting invalid JID

srossbach:
Contacts with an invalid JID will no longer be invited. This ensures
that the Core can assume that it operates on FQJID all the time.

### DIFF
--- a/eclipse/src/saros/ui/actions/OpenChatAction.java
+++ b/eclipse/src/saros/ui/actions/OpenChatAction.java
@@ -100,7 +100,7 @@ public class OpenChatAction extends Action implements Disposable {
 
         return user.getJID();
       } else {
-        return contacts.get(0);
+        return contacts.get(0).getBareJID();
       }
     }
 
--- a/eclipse/src/saros/ui/model/roster/RosterEntryElement.java
+++ b/eclipse/src/saros/ui/model/roster/RosterEntryElement.java
@@ -28,6 +28,10 @@ public class RosterEntryElement extends TreeElement {
 
   private final boolean hasSarosSupport;
 
+  /* TODO away mode is can be displayed wrong if multiple resources for a given JID exist. This
+   * currently does not matter as the content provider will only return one RosterEntryElement for a JID.
+   */
+
   public RosterEntryElement(Roster roster, JID jid, boolean hasSarosSupport) {
 
     this.roster = roster;
@@ -48,7 +52,7 @@ public class RosterEntryElement extends TreeElement {
     final RosterEntry rosterEntry = getRosterEntry();
 
     if (rosterEntry == null) {
-      styledString.append(jid.toString());
+      styledString.append(jid.getBase().toString());
       return styledString;
     }
 
--- a/eclipse/src/saros/ui/util/CollaborationUtils.java
+++ b/eclipse/src/saros/ui/util/CollaborationUtils.java
@@ -72,6 +72,8 @@ public class CollaborationUtils {
 
     final Map<IProject, List<IResource>> newResources = acquireResources(resources, null);
 
+    final List<JID> contactsToInvite = ensureFQJID(contacts);
+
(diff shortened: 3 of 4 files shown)


srossbach:
Any reason why this patch was not reviewed ?

tobous:
To be honest, I though that this was superseded by #494. Will have a look at it.

srossbach:
I guess that will be superseded when we chance to Core API from JID to Contacts, however this is a good start to eliminate some bareJid vs fqJid stuff in the Core.

tobous:
For my understanding, what exactly is the difference between a fully qualified JID (fqjid) and a "bare" JID?

srossbach:
Hmm, today review ... 😴  

alice@foo is a bare JID (think of an broadcast ip like 192.168.0.255)
alice@foo/Saros is a FQJID like an ip 192.168.0.27

eclipse/src/saros/ui/actions/OpenChatAction.java
```diff
@@ -100,7 +100,7 @@ private JID getSelectedJID() {
 
         return user.getJID();
       } else {
-        return contacts.get(0);
+        return contacts.get(0).getBareJID();
```
tobous: Why is it important that this uses the bare JID?

eclipse/src/saros/ui/model/roster/RosterEntryElement.java
```diff
@@ -28,6 +28,10 @@
 
   private final boolean hasSarosSupport;
 
+  /* TODO away mode is can be displayed wrong if multiple resources for a given JID exist. This
```
tobous:
```suggestion
  /* TODO away mode can be displayed incorrectly if multiple resources for a given JID exist. This
```

eclipse/src/saros/ui/model/roster/RosterEntryElement.java
```diff
@@ -48,7 +52,7 @@ public StyledString getStyledText() {
     final RosterEntry rosterEntry = getRosterEntry();
 
     if (rosterEntry == null) {
-      styledString.append(jid.toString());
+      styledString.append(jid.getBase().toString());
```
tobous: Why are we only interested in the base JID?

eclipse/src/saros/ui/util/CollaborationUtils.java
```diff
@@ -460,4 +464,27 @@ private static void refreshProjects(
...
   }
+
+  private static List<JID> ensureFQJID(final Collection<JID> jids) {
+
+    final List<JID> result = new ArrayList<>();
+
+    for (final JID jid : jids) {
+      if (jid.isResourceQualifiedJID()) {
+        result.add(jid);
+        continue;
```
tobous: Use else block instead of this `continue` to make control flow more apparent? Probably a matter of taste.

eclipse/src/saros/ui/actions/OpenChatAction.java
srossbach: I already explained this during the review that we just use bare JID for the chat.

1. If you want to send a message to an offline contact you must use the bare JID
2. If you use a bare JID the server is responsible to route this to the correct presence.

eclipse/src/saros/ui/util/CollaborationUtils.java
srossbach: I guess it more useful to display all invalid JID at once in case you are trying to invite 30. Each invalid contact will spawn a dialog, one after another.

eclipse/src/saros/ui/actions/OpenChatAction.java
tobous: My bad, didn't notice the class name.

stefaus:
The merge of `XMPPContact`s has superseded these changes.

PR #480, opened 2019-03-21
Add and use JQueryHelper

m273d15:
I am currently working on substituting the swt-browser-improved by the swt browser in order to allow us to try other browsers as the jfxbrowser and simplify the HTML-GUI integration.

The only additional functionality of the swt-browser-improved is the JQueryBrowser function `.containsElement(selector)`. Therefore this change substitutes the additional function with a plain JQuery command via the javascript execution api and avoids code duplication of common JQuery commands.

This PR allows a simple switch to a new browser interface.

The subsequent steps are:
* Introduce browser interface and swt browser implementation
* Use the new interface in production and test code


ui/test/framework/stf/src/saros/stf/server/rmi/htmlbot/widget/impl/RemoteHTMLMultiSelect.java
```diff
@@ -31,9 +30,7 @@ public static RemoteHTMLMultiSelect getInstance() {
 
   @Override
   public void select(List<String> value) throws RemoteException {
-    String name = BotUtils.getSelectorName(selector);
-
-    browser.syncRun(String.format("view.setFieldValue('%s', '%s')", name, value.toString()));
+    new JQueryHelper(browser).setFieldValue(selector, value.toString());
```
AbdBarho: is it possible not to create a new instance of the JQueryHelper every time? it feels like a waste.

maybe make the browser a parameter in JQueryHelper? or maybe create an instance of JQueryHelper for every instance of browser and attach it directly to the browser?

m273d15: I will move the helper instance into the abstract class `HTMLSTFRemoteObject`

ui/src/saros/stf/server/bot/jquery/JQueryHelper.java
```diff
@@ -0,0 +1,100 @@
+package saros.stf.server.bot.jquery;
+
+import de.fu_berlin.inf.ag_se.browser.extensions.IJQueryBrowser;
+import de.fu_berlin.inf.ag_se.browser.html.ISelector;
+import de.fu_berlin.inf.ag_se.browser.html.ISelector.NameSelector;
+import java.util.ArrayList;
+import java.util.List;
+
+ public class JQueryHelper {
```
tobous: Please add javadoc to the class and its methods.

ui/src/saros/stf/server/rmi/htmlbot/impl/HTMLBotImpl.java
```diff
@@ -36,13 +36,13 @@ public IRemoteHTMLView view(View view) {
 
   @Override
   public List<String> getAccountList() throws RemoteException {
-    return BotUtils.getListItemsText(getBrowser(MainPage.class), SELECTOR_ACCOUNT_ENTRY);
+    return new JQueryHelper(getBrowser(MainPage.class)).getListItemsText(SELECTOR_ACCOUNT_ENTRY);
```
tobous: How often is this called and can the browser for `MainPage.class` ever change? Would it be feasible/sensible to instantiate the `JQueryHelper` in the CTOR and hold it as an instance variable instead of always creating new instances?

ui/src/saros/stf/server/rmi/htmlbot/impl/HTMLBotImpl.java
```diff
@@ -36,13 +36,13 @@ public IRemoteHTMLView view(View view) {
...
   }
 
   @Override
   public List<String> getContactList(View view) throws RemoteException {
-    return BotUtils.getListItemsText(
-        getBrowser(view.getPageClass()), SELECTOR_CONTACT_ITEM_DISPLAY_NAME);
+    return new JQueryHelper(getBrowser(view.getPageClass()))
```
tobous: Same as in `getAccountList()`. Move instantiation to CTOR and hold as instance variable?

ui/src/saros/stf/server/bot/jquery/JQueryHelper.java
```diff
@@ -0,0 +1,100 @@
...
+   public JQueryHelper(IJQueryBrowser browser) {
+    this.browser = browser;
+  }
+
+   public void clickOnSelection(ISelector selector) {
+    this.browser.run(String.format("%s[0].click()", selector.getStatement()));
+  }
+
+   public String getTextOfSelection(ISelector selector) {
+    return (String)
```
tobous: The usage of this call in [RemoteHTMLTextElement](https://github.com/saros-project/saros/pull/480/commits/077a0fe619f8bf0c2a3ef6c4ac9b4d4fdd441481#diff-95f5543c9abb70f2995c396dbcfddaa8R18) still treats the returned value as an object of type `Object`. Either don't cast it here and adjust the return type or adjust the usage of the method.

ui/src/saros/stf/server/rmi/htmlbot/widget/impl/RemoteHTMLView.java
```diff
@@ -179,19 +176,6 @@ private void ensureExistence(ISelector selector) throws RemoteException {
...
-              .get(BotPreferences.SHORT_TIMEOUT, TimeUnit.MILLISECONDS);
-    } catch (InterruptedException e) {
-      Thread.currentThread().interrupt();
-    } catch (ExecutionException e) {
-      log.error("could not determine whether element exists", e);
-    } catch (TimeoutException e) {
-      log.error("could not determine whether element exists", e);
-    }
-    return foundIt;
+    return new JQueryHelper(getBrowser()).selectionExists(selector);
```
tobous: Use `jQueryHelper` object held in instance variable of super class `HTMLSTFRemoteObject` instead?

ui/src/saros/stf/server/rmi/htmlbot/impl/HTMLBotImpl.java
m273d15: The browser could change (based on the functionality of `BrowserManager`). I am not really sure why the browser of a page should change. I think this is a topic we should keep in mind. Probably, the replacement of a browser is just a function which is never necessary (currently the browser is set only once in `BrowserCreator`) and just increases the complexity.

m273d15: Moving the instantiation into the CTOR might cause issues in the feature. Therefore I would prefer not to change it. What do you think?

ui/src/saros/stf/server/bot/jquery/JQueryHelper.java
```diff
@@ -0,0 +1,100 @@
...
+    String code =
+        "var strings = [];"
+            + selector.getStatement()
+            + ".each(function (i) { "
+            + "strings[i] = $(this).text().trim(); }); return strings; ";
+
+     return this.getListItems(selector, code);
+  }
+
+   private List<String> getListItems(ISelector selector, String code) {
```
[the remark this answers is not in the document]
m273d15: Done

ui/src/saros/stf/server/rmi/htmlbot/widget/impl/RemoteHTMLView.java
```diff
@@ -179,19 +176,6 @@ private void ensureExistence(ISelector selector) throws RemoteException {
...
-    boolean foundIt = false;
-    try {
-      foundIt =
-          getBrowser()
-              .containsElement(selector)
-              .get(BotPreferences.SHORT_TIMEOUT, TimeUnit.MILLISECONDS);
-    } catch (InterruptedException e) {
-      Thread.currentThread().interrupt();
-    } catch (ExecutionException e) {
-      log.error("could not determine whether element exists", e);
```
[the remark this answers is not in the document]
m273d15: Done

ui/src/saros/stf/server/bot/jquery/JQueryHelper.java
m273d15: Done

ui/src/saros/stf/server/bot/jquery/JQueryHelper.java
m273d15: Adjusted the call.

m273d15:
@tobous Thank you for reviewing the PR.

ui/src/saros/stf/server/rmi/htmlbot/impl/HTMLBotImpl.java
tobous: Fine with me. This is only a small optimization anyway.

ui/src/saros/stf/server/rmi/htmlbot/widget/impl/RemoteHTMLView.java
tobous: Ok, this seems to be a problem in my understanding of the class. I assumed that `getBrowser()` would just return the browser object held by this object, in which case this could be replaced with the usage of the `jQueryHelper` held by the object. But, in actuality, this call gets a specific browser object from the `browserManager` (see [here](https://github.com/saros-project/saros/blob/b4f129c972ade993cb4307d324bee7bcbb37d8a9/ui/src/saros/stf/server/rmi/htmlbot/widget/impl/RemoteHTMLView.java#L170)).

This could be avoided by adjusting the name of the method to better reflect the functionality. But this does not have to be done as part of this PR.

m273d15: In my first version I always created a new instance of JQueryHelper. I understood [the point of @AbdBarho](https://github.com/saros-project/saros/pull/480#discussion_r269471682) and just created a new instance if the browser changes. If you like I can change it. I see no big improvement in readability.

ui/src/saros/stf/server/bot/jquery/JQueryHelper.java
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  /**
+   * Verifies whether an element defined by a selector exists.
+   *
+   * @param selector selector whose existence has to be verified
+   * @return <code>true</code> if the element defined by the selector exists
+   */
+  public boolean selectionExists(ISelector selector) {
+    return (Boolean)
```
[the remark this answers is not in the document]
tobous: What about this `Boolean`?

ui/src/saros/stf/server/rmi/htmlbot/widget/impl/RemoteHTMLView.java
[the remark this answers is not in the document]
tobous: The unused logger is still present. See line 27.

m273d15:
@tobous Removed the unused logger and added the null check to JQueryHelper. My first idea was to move the null check in the class calling the method, but after checking the JS-snipped I changed my mind.  

I think the whole approach of using JQuery for testing (it is never used in production) should be reworked. Maintaining the ids in the Java code might lead to very brittle stf tests.

tobous:
> I think the whole approach of using JQuery for testing (it is never used in production) should be reworked. Maintaining the ids in the Java code might lead to very brittle stf tests.

How about opening an issue for this so that we don't forget about it?

PR #481, opened 2019-03-22
[INTERNAL][HTML] Introduce IBrowser and SWTBrowser

m273d15:
* IBrowser: an interface to abstract
  from the browser logic.
* SWTBrowser: an implementation of
  IBrowser for the SWT Browser
* JavascriptFunction and NoCheckedExceptionCallable
  as in the swt browser improved

Extracted the used logic of the swt-browser-improved

### DIFF
--- a/ui/src/saros/ui/browser/IBrowser.java
+++ b/ui/src/saros/ui/browser/IBrowser.java
@@ -1,0 +1,78 @@
+package saros.ui.browser;
+
+public interface IBrowser {
+
+  /*
+   * Try to set the focus on the browser widget
+   *
+   * @return a boolean that is false if it was not possible to set the focus
+   */
+  public boolean setFocus();
+
+  /*
+   * Change the size of the browser
+   *
+   * @param width the new width of the browser widget
+   *
+   * @param height the new height of the browser widget
+   */
+  public void setSize(final int width, final int height);
+
+  /*
+   * Returns the currently URL which is currently loaded by the browser
+   *
+   * @return a string that represents the currently loaded URL
+   */
+  public String getUrl();
+
+  /*
+   * Load a specified URL and block until the page is loaded or the timeout is
+   * exceeded
+   *
+   * @param url the URL which has be loaded
+   *
+   * @param timeout the timeout that specified the maximal load time
+   *
+   * @return a boolean that is false if the URL cannot be loaded (within the
+   * timeout)
+   */
+  public boolean loadUrl(final String url, final int timeout);
+
+  /*
+   * Load a specified HTML content into the browser
+   *
+   * @param html the HTML that has to be loaded
+   */
+  public void loadHtml(final String html);
+
+  /*
+   * Execute a JavaScript code that returns an object.
+   *
+   * @param jsCode the JavaScript code that has to be executed
+   *
+   * @return an object that represents an arbitrary return value of jsCode
+   */
+  public Object evaluate(final String jsCode);
+
+  /*
+   * Execute a JavaScript code without a return value.
+   *
+   * @param jsCode the JavaScript code that has to be executed
+   */
+  public boolean execute(final String jsCode);
+
+  /*
+   * Add the functionality a JavascriptFunction class to the browser
+   *
+   * @param jsFunction the function that has to be injected into the browser
+   */
+  public void addBrowserFunction(final AbstractJavascriptFunction jsFunction);
+
+  /*
+   * Add runnable that is executed after the browser widget is disposed
+   *
+   * @param run a runnable that is executed after the browser widget is
+   * disposed
+   */
+  public void runOnDisposal(Runnable run);
+}
--- a/ui/src/saros/ui/browser/IBrowser.java
+++ b/ui/src/saros/ui/browser/IBrowser.java
@@ -58,6 +58,7 @@ public interface IBrowser {
    * Execute a JavaScript code without a return value.
    *
    * @param jsCode the JavaScript code that has to be executed
+   * @return true if the execution was successful
    */
   public boolean execute(final String jsCode);
 
--- a/ui/src/saros/ui/browser/SWTBrowser.java
+++ b/ui/src/saros/ui/browser/SWTBrowser.java
@@ -5,6 +5,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+import org.apache.log4j.Logger;
 import org.eclipse.swt.browser.Browser;
 import org.eclipse.swt.browser.BrowserFunction;
 import org.eclipse.swt.browser.ProgressEvent;
@@ -19,6 +20,8 @@ public class SWTBrowser implements IBrowser {
     LOADING,
     COMPLETE
   }
+
+  private static final Logger log = Logger.getLogger(SWTBrowser.class);
 
   private final Browser browser;
   private ReadyState state = ReadyState.LOADING;
@@ -42,10 +45,14 @@ public class SWTBrowser implements IBrowser {
     browser = new Browser(parent, style);
 
     browser.addDisposeListener(
-        (DisposeEvent e) -> {
+        (DisposeEvent event) -> {
           synchronized (runOnDisposalList) {
             for (Runnable runnable : runOnDisposalList) {
-              runnable.run();
+              try {
+                runnable.run();
+              } catch (RuntimeException exception) {
+                log.error("error running on disposal: " + exception.getMessage(), exception);
+              }
             }
           }
         });
@@ -53,7 +60,7 @@ public class SWTBrowser implements IBrowser {
 
   @Override
   public boolean setFocus() {
-    return this.syncExec(
+    return syncExec(
         () -> {
           return browser.setFocus();
         });
@@ -61,12 +68,12 @@ public class SWTBrowser implements IBrowser {
 
   @Override
   public void setSize(final int width, final int height) {
-    this.syncExec(() -> browser.setSize(width, height));
+    syncExec(() -> browser.setSize(width, height));
   }
 
   @Override
   public String getUrl() {
-    return this.syncExec(
+    return syncExec(
         () -> {
           return browser.getUrl();
         });
@@ -74,22 +81,23 @@ public class SWTBrowser implements IBrowser {
 
   @Override
   public boolean loadUrl(final String url, final int timeout) {
-    browser.getShell().open();
+    syncExec(() -> browser.getShell().open());
 
-    this.state = ReadyState.LOADING;
-    browser.addProgressListener(readyStateListener);
-
-    this.syncExec(
+    state = ReadyState.LOADING;
+    syncExec(
         () -> {
+          browser.addProgressListener(readyStateListener);
           browser.setUrl(url);
         });
+
     boolean successful = waitForPageLoaded(timeout);
 
-    browser.removeProgressListener(readyStateListener);
+    syncExec(() -> browser.removeProgressListener(readyStateListener));
 
     return successful;
   }
 
+  /** This method waits as long as defined in <code>timeout</code> */
   private boolean waitForPageLoaded(final int timeout) {
     AtomicBoolean waiting = new AtomicBoolean(true);
     Display display = browser.getDisplay();
@@ -103,37 +111,35 @@ public class SWTBrowser implements IBrowser {
             display.wake();
           }
         };
+    syncExec(() -> display.timerExec(timeout, wakeUpTimeout));
 
-    display.timerExec(timeout, wakeUpTimeout);
-
-    this.syncExec(
+    syncExec(
         () -> {
           while (waiting.get() && state != ReadyState.COMPLETE) {
             if (!display.readAndDispatch()) display.sleep();
           }
         });
 
-    return this.state == ReadyState.COMPLETE;
+    return state == ReadyState.COMPLETE;
   }
 
   @Override
   public void loadHtml(final String html) {
-    this.syncExec(() -> browser.setText(html));
+    syncExec(() -> browser.setText(html));
   }
 
   @Override
   public Object evaluate(final String jsCode) {
-    return this.syncExec(
+    return syncExec(
         () -> {
           Object rv = browser.evaluate(jsCode);
-          System.out.println(rv);
           return rv;
         });
   }
 
   @Override
   public boolean execute(String jsCode) {
-    return this.syncExec(
+    return syncExec(
         () -> {
           return browser.execute(jsCode);
         });
@@ -141,7 +147,7 @@ public class SWTBrowser implements IBrowser {
 
   @Override
   public void addBrowserFunction(final AbstractJavascriptFunction function) {
-    this.syncExec(
+    syncExec(
         () -> {
           BrowserFunction swtFunction =
               new BrowserFunction(browser, function.getName()) {
@@ -162,6 +168,8 @@ public class SWTBrowser implements IBrowser {
   private <V> V syncExec(final NoCheckedExceptionCallable<V> callable) {
     final AtomicReference<V> r = new AtomicReference<V>();
     final AtomicReference<RuntimeException> exception = new AtomicReference<RuntimeException>();
+    final AtomicReference<Error> error = new AtomicReference<Error>();
+
     browser
         .getDisplay()
         .syncExec(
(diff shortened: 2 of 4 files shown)


ui/src/saros/ui/browser/SWTBrowserWrapper.java
```diff
@@ -0,0 +1,192 @@
...
+   * @param url
+   * @param timeout
+   * @return
+   */
+  private boolean waitForPageLoaded(final String url, final long timeout) {
+
+    Runnable waitForLoaded =
+        () -> {
+          while (this.state != ReadyState.COMPLETE) {
+            if (!browser.getDisplay().readAndDispatch()) {
```
srossbach: I do not think that this is working at all. Do you even know what you are doing here ? You are trying to dispatch UI events from another thread. I guess that this thread always crashes due to an illegal access exception.

ui/src/saros/ui/browser/SWTBrowserWrapper.java
```diff
@@ -0,0 +1,192 @@
...
+              browser.getDisplay().sleep();
+            }
+          }
+        };
+
+    Thread t = new Thread(waitForLoaded);
+    t.setDaemon(true);
```
srossbach: Why daemon ?

ui/src/saros/ui/browser/SWTBrowserWrapper.java
```diff
@@ -0,0 +1,192 @@
...
+    t.start();
+
+    try {
+      t.join(timeout);
+    } catch (InterruptedException e) {
+      return false;
+    }
+    return true;
```
srossbach: This method always returns true unless the caller somehow interrupts the thread that his method was calling.

ui/src/saros/ui/browser/SWTBrowserWrapper.java
```diff
@@ -0,0 +1,192 @@
...
+                public Object function(Object[] arguments) {
+                  return function.function(arguments);
+                }
+              };
+          return swtFunction;
+        });
+  }
+
+  @Override
+  public void runOnDisposal(Runnable runnable) {
```
srossbach: Race conditions. As it seems that this class can be used from any thread this list is accessed by multiple threads and ArrayLists are not thread safe.

ui/src/saros/ui/browser/SWTBrowserWrapper.java
```diff
@@ -0,0 +1,192 @@
...
+    runOnDisposalList.add(runnable);
+  }
+
+  private <V> V syncExec(final NoCheckedExceptionCallable<V> callable) {
+    final AtomicReference<V> r = new AtomicReference<V>();
+    final AtomicReference<RuntimeException> exception = new AtomicReference<RuntimeException>();
+    Display.getDefault()
```
srossbach: Display.getDefault() should not be called in Eclipse. You can "revive" / create a new display during the shutdown of the Eclipse IDE because the display is already disposed when the plugin tears down.

ui/src/saros/ui/browser/IBrowserWrapper.java
```diff
@@ -0,0 +1,78 @@
+package saros.ui.browser;
+
+public interface IBrowserWrapper {
```
srossbach: What is the problem with IBrowser ?

m273d15: You mean the name or the interface of the swt-browser? if you mean the name: sure we can call it IBrowser.

ui/src/saros/ui/browser/SWTBrowserWrapper.java
m273d15: I only moved this logic from the swt-browser-improved and forgot to adjust this call. `browser.getDisplay()` should be right.

ui/src/saros/ui/browser/SWTBrowserWrapper.java
m273d15: I will improve this so that the method returns only **true** if the loading process was successful.

ui/src/saros/ui/browser/IBrowserWrapper.java
srossbach: The name of the interface. I cannot remember that I have seen an IXyzWrapper in any application yet.

ui/src/saros/ui/browser/SWTBrowserWrapper.java
m273d15: It seems to work, but I know what you mean. I would expect a `SWTException: Invalid thread access`. I already had different versions. Which would you prefer. I am not sure which is better:
* Just wrap the if-block it into a `syncExec`
* Remove the display operations and use a Semaphore (initial permits = 0) which blocks in `acquire` here until it is released in the ProgressListener.

ui/src/saros/ui/browser/SWTBrowserWrapper.java
m273d15: Can be removed.

ui/src/saros/ui/browser/SWTBrowserWrapper.java
m273d15: I will adjust it

ui/src/saros/ui/browser/SWTBrowserWrapper.java
srossbach: We have to clarify what thread invoked the listener. If this is always not the SWT thread then simply wait (with a timeout) for the result. I would also recommend to block here even it is called from the SWT because if you call readAndDispatch you are may expect that setUrl can be called again. I do not know how long it takes to load the pages but a small UI freeze might be acceptable in this situation.

m273d15:
@srossbach Thanks for your review. This PR is still in a raw state I will improve the whole implementation + documentation and want to write tests to verify whether the whole Browser integration works.
I will be away until the 4.4. Therefore I will not answer or improve this PR in the next days.

ui/src/saros/ui/browser/IBrowserWrapper.java
m273d15: The postfix `Wrapper` is not very informative. I will remove it.

ui/src/saros/ui/browser/SWTBrowserWrapper.java
m273d15: The execution thread of the listeners seems to depend on the implementation:
* IE: It seems like the listeners are executed [directly](https://github.com/eclipse/eclipse.platform.swt/blob/6bdf4f12c125b636189a4de55260df5a71775f2e/bundles/org.eclipse.swt/Eclipse%20SWT%20Browser/win32/org/eclipse/swt/browser/IE.java#L656)
* GTK-WebKit1: [it is executed directly](https://github.com/eclipse/eclipse.platform.swt/blob/6bdf4f12c125b636189a4de55260df5a71775f2e/bundles/org.eclipse.swt/Eclipse%20SWT%20WebKit/gtk/org/eclipse/swt/browser/WebKit.java#L2464),  
GTK-WebKit2: it is executed in the ui thread 
Cocoa-WebKit: it is [executed directly](https://github.com/eclipse/eclipse.platform.swt/blob/6bdf4f12c125b636189a4de55260df5a71775f2e/bundles/org.eclipse.swt/Eclipse%20SWT%20WebKit/cocoa/org/eclipse/swt/browser/WebKit.java#L821)

srossbach: It seems that this is always called in the SWT EDT. So if you make sure waitForPageLoaded is invoked in the EDT (seems loadUrl is the only method that is not executed on the SWT) you can simply spin wait via.

    while (!completed)
      if (!display.readAndDispatch()) Thread.yield();

Or call display.sleep() instead. But I do not know when the display is waked up again. (of course you can spawn a thread that calls Display#wake() after the timeout.

Just one question. As far as I remember there was an issue with the completed event and so the improved SWTBrowser injected some JS code into the browser to get notified when the page is really loaded. Is this no longer the case, i.e was is fixed in the Eclipse SWTBrowser ?

m273d15: The issue never appeared during my tests, but I am aware that the possibility exist. For now the SarosApi and the DOM were always available. 
I thought the issue could be related to [BUG 433400](https://bugs.eclipse.org/bugs/show_bug.cgi?id=433400) which was solved by [this commit](https://github.com/eclipse/eclipse.platform.swt/commit/0658d789bfb0cb2ccab2ad6b8af953cadf626c1e).

m273d15: > It seems that this is always called in the SWT EDT. So if you make sure waitForPageLoaded is invoked in the EDT (seems loadUrl is the only method that is not executed on the SWT) you can simply spin wait via.

Thanks for clarifying.

ui/src/saros/ui/browser/SWTBrowser.java
```diff
@@ -0,0 +1,199 @@
...
+
+  private enum ReadyState {
+    LOADING,
+    COMPLETE
+  }
+
+  private final Browser browser;
+  private ReadyState state = ReadyState.LOADING;
+  private List<Runnable> runOnDisposalList =
+      Collections.synchronizedList(new ArrayList<Runnable>());
```
m273d15: Used synchronized list in order to avoid race-condition issues.

ui/src/saros/ui/browser/SWTBrowser.java
```diff
@@ -0,0 +1,199 @@
...
+    Runnable wakeUpTimeout =
+        () -> {
+          // avoid execution if url is already loaded
+          if (state != ReadyState.COMPLETE) {
+            waiting.set(false);
+            display.wake();
+          }
+        };
+
+    display.timerExec(timeout, wakeUpTimeout);
```
m273d15: The timeout related wake-up task and the event-loop are running on the ui thread. Therefore race-conditions between should be avoided. I know it is not perfect, because the `wakeUpTimeout` runnable is still executed (as NOP) even if the event-loop is already terminated. Other synchronization methods always allowed the case that `display.wake()` is executed after the check of the while loop, but before `display.sleep()`.
I tested this with our current html page (no timeout) and a html containing a `for (;;) {}` script (caused timeout).

tobous:
I think you made a typo with the the tag for the commit. It is probably supposed to be `[UI][HTML]` and not `[I][HTML]`.

m273d15:
You are right. I had "INTERNAL" in mind, but used the wrong tag :sweat_smile:

ui/src/saros/ui/browser/SWTBrowser.java
srossbach: Just two questions.

Why must the class handle the UI thread access ? You are doing more work than neccessary.

If you call runAndDispatch you are able to call loadUrl again. I think it is better to abort with an illegalstate exception here because you start to load another html page.

Just create some async runnables that call loadUrl in these runables. They will not be executed one after another but all during the recursion.

See https://github.com/saros-project/saros/blob/master/core/src/saros/session/internal/ActivityHandler.java#L262

and assume modal context = readAndDispatch

ui/src/saros/ui/browser/IBrowser.java
```diff
@@ -0,0 +1,78 @@
...
+   * @return an object that represents an arbitrary return value of jsCode
+   */
+  public Object evaluate(final String jsCode);
+
+  /*
+   * Execute a JavaScript code without a return value.
+   *
+   * @param jsCode the JavaScript code that has to be executed
+   */
+  public boolean execute(final String jsCode);
```
srossbach: return value is ?

ui/src/saros/ui/browser/SWTBrowser.java
```diff
@@ -0,0 +1,199 @@
...
+      };
+
+  public SWTBrowser(Composite parent, int style) {
+    browser = new Browser(parent, style);
+
+    browser.addDisposeListener(
+        (DisposeEvent e) -> {
+          synchronized (runOnDisposalList) {
+            for (Runnable runnable : runOnDisposalList) {
+              runnable.run();
```
srossbach: When one runnable raises a runtime exception nothing will be executed afterwards. Is this intended ?

ui/src/saros/ui/browser/SWTBrowser.java
```diff
@@ -0,0 +1,199 @@
...
+  public boolean setFocus() {
+    return this.syncExec(
+        () -> {
+          return browser.setFocus();
+        });
+  }
+
+  @Override
+  public void setSize(final int width, final int height) {
+    this.syncExec(() -> browser.setSize(width, height));
```
srossbach: Isn't the coding convention like: omit "this" when it is not needed ?

ui/src/saros/ui/browser/SWTBrowser.java
```diff
@@ -0,0 +1,199 @@
...
+  public String getUrl() {
+    return this.syncExec(
+        () -> {
+          return browser.getUrl();
+        });
+  }
+
+  @Override
+  public boolean loadUrl(final String url, final int timeout) {
+    browser.getShell().open();
```
srossbach: Just wondering if it is possible to call this on a different thread than the EDT ? If it is not possible then the syncCall on line 82 is not necessary.

srossbach: Ok as I thought: ERROR_THREAD_INVALID_ACCESS - if not called from the thread that created the receiver

ui/src/saros/ui/browser/SWTBrowser.java
```diff
@@ -0,0 +1,199 @@
...
+  public void loadHtml(final String html) {
+    this.syncExec(() -> browser.setText(html));
+  }
+
+  @Override
+  public Object evaluate(final String jsCode) {
+    return this.syncExec(
+        () -> {
+          Object rv = browser.evaluate(jsCode);
+          System.out.println(rv);
```
srossbach: ...

ui/src/saros/ui/browser/SWTBrowser.java
```diff
@@ -0,0 +1,199 @@
...
+          return swtFunction;
+        });
+  }
+
+  @Override
+  public void runOnDisposal(Runnable runnable) {
+    runOnDisposalList.add(runnable);
+  }
+
+  private <V> V syncExec(final NoCheckedExceptionCallable<V> callable) {
```
srossbach: Looking at https://github.com/saros-project/saros/blob/master/eclipse/src/saros/ui/util/SWTUtils.java#L138 

Do you really want to swallow "Error" ?

ui/src/saros/ui/browser/SWTBrowser.java
```diff
@@ -0,0 +1,199 @@
...
+              }
+            });
+
+    if (exception.get() != null) {
+      throw exception.get();
+    }
+    return r.get();
+  }
+
+  private void syncExec(final Runnable runnable) {
```
srossbach: syncExec( () -> {runnable.run(); return null;});

m273d15: What do mean?

srossbach: This one-liner is a replacement for the current method body.

ui/src/saros/ui/browser/SWTBrowser.java
m273d15: I created a proper testing environment on my PC and found a lot of sources of an `ERROR_THREAD_INVALID_ACCESS`. I will add the missing syncExec.

ui/src/saros/ui/browser/SWTBrowser.java
m273d15: Ah, you mean `SWTUtils.runSWTSync(() -> {runnable.run(); return null;})` ? Yes, should be the more convenient solution.

srossbach: You do not have access to SWTUtils. Simply insert the code, it will call the syncExec method of this class.

m273d15: Sorry, it is to hot. I forgot that I already defined another syncExec version.
**Done**

ui/src/saros/ui/browser/SWTBrowser.java
m273d15: Done

ui/src/saros/ui/browser/SWTBrowser.java
m273d15: Done :see_no_evil:

ui/src/saros/ui/browser/SWTBrowser.java
m273d15: Done

ui/src/saros/ui/browser/SWTBrowser.java
m273d15: Done

ui/src/saros/ui/browser/SWTBrowser.java
m273d15: Done

ui/src/saros/ui/browser/IBrowser.java
m273d15: Added doc.
Done

ui/src/saros/ui/browser/SWTBrowser.java
```diff
@@ -0,0 +1,200 @@
...
+    browser = new Browser(parent, style);
+
+    browser.addDisposeListener(
+        (DisposeEvent event) -> {
+          synchronized (runOnDisposalList) {
+            for (Runnable runnable : runOnDisposalList) {
+              try {
+                runnable.run();
+              } catch (RuntimeException exception) {
+                log.error("error running on disposal: " + exception.getMessage(), exception);
```
srossbach: I really do not find that kind of logging useful. You will find the message of the exception anyway.

E.G

0 [main] ERROR saros.activities.ActivityOptimizerTest  - error wile reading file: \foo (The system cannot find the file specified)
java.io.FileNotFoundException: \foo (The system cannot find the file specified)

PR #483, opened 2019-03-24
Adjust UI on client side of project negotiation to handle multiple open projects

tobous:
PR completely reworking the UI to accept the project negotiation and choose the local representation of a shared module.

#### [UI][I] Add boilerplate code for new module selection page

Adds the boilerplate code for the new classes added to re-implement the
module selection page. The newly added class
SelectLocalModuleRepresentationPage will eventually replace
SelectProjectPage.

The new implementation splits the different responsibilities into
different classes to make it easier to modify the logic to allow for
multiple shared module and to make the logic easier to understand.

#### [INTERNAL][I] Add the basic layout for the module tabs

Adds the logic to initialize a basic layout for the ModuleTab class.

Adds TODOs for the UI logic that still needs to be implemented.

#### [INTERNAL][I] Add logic to fill the project combo-box

Adds the logic to fill the project combo-box in ModuleTab.
The combo-box is filled with all currently open IntelliJ projects. The
entries are sorted alphabetically.

#### [INTERNAL][I] Add logic to radio buttons

Adds the logic that makes the radio buttons in ModuleTab exclusive to
each other and that disables all fields of the option that is not
chosen.

#### [INTERNAL][I] Add a file chooser to the base path field

Adds a file chooser to the field to enter the base path of the new
module in the ModuleChooser.

#### [INTERNAL][I] Add logic to set default values for module selection

Adds the logic to set default values and the default selection for the
dialog to choose a local module representation, implemented in
ModuleTab.
The selection logic checks if there is an open project that contains a
module with a matching name. If such a project module combination is
found, the matching options are selected in the dialog. If multiple
entries are found, the first one is used. If no such entries are found,
the first project of the list is selected and the option to create a new
module is selected.

#### [INTERNAL][I] Add project combo-box listener to module selection dialog

Adds a project combo-box listener to ModuleTab that updates the other
values and the default selection of the dialog when a new project is
chosen.

#### [INTERNAL][I] Add logic to update the validity state on input changes

Adds the logic to update the validity state of the ModuleTab on input
change to the rest of the dialog fields.

#### [INTERNAL][I] Implement logic determining the current validity state

Implement the logic to determine the current validity state of the
ModuleTab. The logic is modeled after current module restrictions,
meaning the new/existing module name has to match the remote module
name.

#### [INTERNAL][I] Add logic to exclude project modules in ModuleTab

Adds the logic to exclude project modules from the choice of existing
modules. This is done to enforce the current restrictions that project
modules can not be shared.

#### [INTERNAL][I] Use new module selection page in AddProjectToSessionWizard

Replaces SelectProjectPage with SelectLocalModuleRepresentationPage.
Subsequently adjusts the logic to use all components provided by the
page:
 - Adjusts the module creation logic to use the selected project and
    base path
 - Adjust the logic to use an existing module to instantiate the
    IProject object directly instead of using
    IWorkspace.getProject(String)

Adjusts the visibility of some components of the newly added module
selection logic to make them accessible from AddProjectToSessionWizard.

#### [REFACTOR][I] Remove SelectProjectPage

Removes SelectProjectPage as it is no longer used. Removes a reference
to the class from Wizard.

Removes the messages only used in SelectProjectPage.
Renames the messages that are actually used by
AddProjectToSessionWizard.

#### [UI][I] Adjust size of AddProjectsToSessionWizard

Adjusts the default size of AddProjectToSessionWizard to fit the
dimensions of the new local module representation selection page.

#### [INTERNAL][I] Split validity check in ModuleTab

Splits the validity check of the different components in ModuleTab to
allow for separate evaluation. This is needed to implement the logic
highlighting  invalid fields.

#### [UI][I] Add red border to invalid fields in ModuleTab

Adds the logic to display a red border around fields whose input is
invalid. Also adds a tool tip to such fields stating why the field is
invalid.

This makes it much more clear to the user why they are not able to
proceed with the dialog.


srossbach:
I do not recommend to use the "tabish" style as it was reviewed as UX disaster. Alexander Waldmann had done a rework of this wizard. It never got it. I picked it up again ( https://saros-build.imp.fu-berlin.de/gerrit/#/c/471/ ). But it was never reviewed.

tobous:
From what I understand, the UI change proposed in the Gerrit change would remove the option to create a new module/project as part of the incoming project negotiation. Instead, the dialog would just give the user a combo box to choose a local representation for each shared module/project. Is that correct?
If so, is that really a restriction that we want to introduce while we still allow module/project sharing? (The option likely has to be removed/adjusted once we move to sharing reference points, but that still has to be discussed further.)

Can you elaborate what you mean by "UX disaster"? I don't see a problem with using a tabbed view. We have the header text box to explain the usage if it is not completely intuitive.

For the current functionality, I don't see a good alternative to a tabbed view as we have to present multiple options. For the new page, the following layout was planned (I hope you enjoy my ghetto ascii art :wink:):
```
|––––––––––––––|––––––––––––––|
| Module tab 1 | ...          |
|––––––––––––––––––––––––––––––––––––––––––––––––––––––––|
|  –––––––––––––––––––––––––––––––––––––                 |
|  | v | Choose open project combo box |                 |
|  –––––––––––––––––––––––––––––––––––––                 |
|                                                        |
|  [ ] Create new module radio button                    |
|    –––––––––––––––––––––––––––––––                     |
|    | Module name text field      |                     |
|    –––––––––––––––––––––––––––––––                     |
|    –––––––––––––––––––––––––––––––  –––––––––––––––––  |
|    | Module base path text field |  | Browse button |  |
|    –––––––––––––––––––––––––––––––  –––––––––––––––––  |
|                                                        |
|  [ ] Use existing module radio button                  |
|    ––––––––––––––––––––––––––––––––––––––––            |
|    | v | Choose existing module combo box |            |
|    ––––––––––––––––––––––––––––––––––––––––            |
|––––––––––––––––––––––––––––––––––––––––––––––––––––––––|
```

Displaying all of this for every shared module on a single page is not feasible. And making multiple pages, one for each shared module, seems more confusing to me. Furthermore, it would complicate the logic as we would have to give the user the option to go back to previous modules, etc.

srossbach:
The tabbed pane is really cumbersome if you get more than 2 projects/modules. You have to navigate through the tabs and remember what you have typed in. The basic idea is to align the combo horizontally and the each module vertically.

As for the UX stuff. You get a combo box dialog, you will it out and now you wonder why you cannot finish the dialog unless you realize that there are multiple tabs. For a first time use, that is not the best experience.

See also https://www.inf.fu-berlin.de/inst/ag-se/theses/Waldmann12-saros-usability-evaluation.pdf starting on page 87.

tobous:
Hmm. I still don't think that the layout proposed in the evaluation scales that well either, especially with the need for a second text field for the base path when creating a new module. If there are too many modules, the wizard still gets cumbersome as you won't be able to display them all on one screen and will have to deal with a vertical scroll-bar.

The best alternative I can think of would probably be to use a separate page for each module where each page holds the panel that was previously represented by a tab (see ascii art above).
But, since Saros/I currently does not support sharing of multiple modules, I won't bother to much with building support for it (for now). I will still adjust my design to best integrate with the wizard-page design.

Any concerns with my proposal?

srossbach:
It really depends if you wait for the HTML UI. In that case this is no longer your duty.

Maybe this topic needs a bit more discussion anyway. I do not think that we have any good data on how many projects/modules are normally shared on session startup.

What still annoys me is that we still have to select the destination of each project over and over again on every session startup. I think I wrote once a patch that remembered the mapping.

tobous:
I had a look at the current wizard implementation, and, while it is possible to move to a wizard-page based approach, it requires more work than I thought.
So, for now, I would like to keep the tabbed-approach. If we at some point decide against the approach (and against the HTML UI), the current design should still be flexible enough that the main logic can be kept.

As to choosing everything at the startup: For Saros/I, I implemented a basic logic to enter useful default values. If a module with a matching name is found, the option to use it for the project sharing is selected by default. Otherwise, the option to create a new module is selected by default using the default name of the shared module. As the current Saros/I implementation still requires that shared modules have the same name of all sides (I haven't had a look at this requirement in a while; I am not sure if it is still necessary), the default options are the only two valid options for the dialog, meaning the user basically never has to (or even can) do anything.
This gets somewhat harder now that we are adjusting the logic to work with multiple projects, but it should still work in most cases. (If there are multiple modules with a matching name located in different projects, the first one found will be chosen. Each time the user changes the selected project, the default pre-selection will be checked again to see if the project contains a matching module. For the module creation, the project base path is entered as the default module base path, meaning the modules will be created in the project directory.)

(Talking about this also made me notice that, in an application context, the module name is no longer a valid unique identifier for a module. I have opened #488 for this.)

tobous:
Wow... these UI patches got a lot bigger than I thought. This PR is still WIP as I need to do some more cleanup, improve commit messages, etc. But the logic should basically be done.

srossbach:
@tobous  can you configure the bot in such a way that only one email is send after the review. I am currently getting an email for each issue that is found by the bot which is annoying.

tobous:
@srossbach Yes, I found that quite annoying as well. We are still working on a good default Config for codacy. There is a summary option, but we will have to check if it still displays all the necessary information.

tobous:
@srossbach @m273d15 Yeah, the summary option is kind of useless as it gives no information and does not allow for a good discussion of the found issues (like a comment at the affected line).

I am not sure what the better option is... It would be nice if Codacy would use the "review as comment" functionality to post all comments at once to avoid flood of notifications...

tobous:
Rebased on current base PR and added a refactoring patch to fix a typo in a method name in `Wizard`.

tobous:
Rebased onto current base PR with minimal interaction to fix merge-conflicts dealing with import statements.

tobous:
Rebased onto current base PR. Fixed some minor merge conflicts dealing with imports.

tobous:
Replaced the combobox wrapper with a custom renderer showing the module/project name as the combobox entry.

tobous:
Add logic to highlight invalid fields in the dialog.

tobous:

tobous:

intellij/src/saros/intellij/ui/wizards/pages/moduleselection/ModuleTab.java
```diff
@@ -408,6 +481,93 @@ private boolean hasValidExistingModule() {
     return selectedExistingModule != null && selectedExistingModule.getName().equals(moduleName);
   }
 
+  /**
+   * Updates whether the field for the new module name is marked as invalid. The new state is based
+   * on the returned value of {@link #hasValidNewModuleName()}.
+   */
+  private void updateNewModuleNameValidityIndicator() {
```
m273d15: This code duplication does not hurt, but we can think about a better design. This might be a good candidate for a Saros-eat-your-own-dog-food-session :relaxed:

tobous:

PR #484, opened 2019-03-25
[INTERNAL][HTML] Re-implement Choose File Step

AbdBarho:
the old implementation used a library that was buggy and caused problems.

this is a commit contains an implementation of the features needed to select file and folders.


AbdBarho:
I have tested the performance of this implementation using the ProjectTree of Saros as an example (since it is relatively big about  ~ 100 000 files/folders, node_modules included).

loading and rendering the start-session-wizard on Chrome out of eclipse took about 3 seconds.
while the same took in eclipse about 15 seconds.

this huge difference can be attributed to :
 - When opening the standalone version, the ProjectTree is already mocked up and in the code of the GUI, but when using the eclipse version, there is an additional step of creating the actual JSON which needs to be sent to the browser, that itself takes some time (I did not measure how long does it take to create the JSON, at least not yet).
 - Chrome is faster than IE in parsing and rendering (hopefully webkit is also as good)
 - The polyfills we are using are not optimized, they are there just to fill the holes in IE, taking them out when the browser actually supports the required features will improve the performance

tobous:
See #695

PR #487, opened 2019-03-27
[UI][E] Update ConnectionState on account changes

srossbach:
This patch ensures that the connection widget is updated when either the
account file changes or accounts are added.


srossbach:
1. If you start Saros/E via Eclipse in STF you will see "Add Account First" although the account store have accounts.
2. (Not sure about this) After a fresh plugin install go through the Configuration Wizard but make sure you do have auto connect disabled. After you finished the wizard it should still display the information that you should add an account first.

Issue #491, opened 2019-04-11
Rework serialization of ColorIDSet and XMPPAccount


m273d15:
In order to persist the configured `XMPPAccount` and the `ColorIDSet` we have currently the following implementation:
* The `ColorIDSetStorage` class uses the default java serialization in the method `save` to serialize a list of `ColorIDSet`s and save the result in the preference store. 
* The `XMPPAccountStore` class uses the default java serialization in the method `saveAccounts` to serialize the current `activeAccount` and the configured `accounts`. The result of the serialization is encrypted and written to a file.

The disadvantage of this approach is that the saved data are highly coupled with the corresponding class implementation. Therefore changing the class name or package name would lead to invalidated data. Furthermore saros just throws an exception if the saved data cannot be deserialized.

m273d15:
We already use XStream. We could use this library in order to avoid the rename issues with using [class aliasing](https://x-stream.github.io/alias-tutorial.html#class_aliasing) (and if required also [field aliasing](https://x-stream.github.io/alias-tutorial.html#field_aliasing))

srossbach:
I have already a patch for the XMPP Account stuff. It is not that hard to serialize data by hand. There is no need to use XStream to generate "enterprise" XML files.

m273d15:
@srossbach Oh sorry. I saw this post directly after pushing #492. If you want to push your solution I will close my PR.

PR #492, opened 2019-04-11
Rework serialization of XMPPAccount and ColorIDSet

m273d15:
* Cleaned up the tests of XMPPAccountStore
* Used XStream in order to (de)serialize XMPPAccount and ColorIDSet
#491

### DIFF
--- a/core/src/saros/account/XMPPAccountStore.java
+++ b/core/src/saros/account/XMPPAccountStore.java
@@ -562,7 +562,11 @@ public final class XMPPAccountStore {
     return null;
   }
 
-  // class which is used for serialization of account information
+  /**
+   * class which is used for serialization of account information
+   *
+   * <p>WARNING: If you change this class you may change the XML format of the account file.
+   */
   private static class AccountStoreInformation {
     public int activeAccountIndex;
     public ArrayList<XMPPAccount> configuredAccounts;
--- a/core/src/saros/account/XMPPAccountStore.java
+++ b/core/src/saros/account/XMPPAccountStore.java
@@ -1,15 +1,12 @@
 package saros.account;
 
+import com.thoughtworks.xstream.XStream;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
-import java.io.DataInputStream;
-import java.io.DataOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
 import java.io.UnsupportedEncodingException;
 import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
@@ -32,6 +29,8 @@ import javax.crypto.NoSuchPaddingException;
 import javax.crypto.spec.IvParameterSpec;
 import javax.crypto.spec.SecretKeySpec;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.log4j.Logger;
 import saros.annotations.Component;
 import saros.net.xmpp.JID;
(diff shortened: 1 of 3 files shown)


core/src/saros/account/XMPPAccountStore.java
```diff
@@ -224,31 +200,18 @@ private synchronized void saveAccounts() {
...
-
-      byte[] activeAccount = Crypto.encrypt(out.toByteArray(), secretKey);
-      out.reset();
-
-      oos = new ObjectOutputStream(out);
-      oos.writeObject(accounts);
-      oos.flush();
+      dataOut = new FileOutputStream(accountFile);
+      // use a pair in order to create a artificial xml root node
+      xStream.toXML(new ImmutablePair(activeAccount, accounts), out);
```
srossbach: List<XMPPAcounts> accountsToSave = new ArrayList<>(accounts)
int activeAccountIdx = accountsToSave.indexOf(acivteAccount)

new ImmutablePair(activeAccountIdx, accountsToSave)

core/src/saros/account/XMPPAccountStore.java
```diff
@@ -594,7 +566,7 @@ public XMPPAccount findAccount(String jidString) {
    * save the result in the user home directory which should be default only be accessible by the
    * user itself or administrators.
    */
-  private static class Crypto {
+  static class Crypto {
```
srossbach: Why did you remove private ?

core/src/saros/account/XMPPAccountStore.java
```diff
@@ -150,44 +149,21 @@ private synchronized void loadAccounts() {
...
-              new ObjectInputStream(new ByteArrayInputStream(Crypto.decrypt(buffer, secretKey)))
-                  .readObject();
-
-      size = dataIn.readInt();
+      dataIn = new FileInputStream(accountFile);
 
-      if (size <= 0 || size > MAX_ACCOUNT_DATA_SIZE)
-        throw new IOException(
-            "account data seems malformed, refused to load " + size + " bytes into memory");
+      byte[] encryptedAccountData = IOUtils.toByteArray(dataIn);
```
srossbach: So what happens if you point to a garbage file with about several 100 MBytes in size ?

core/src/saros/account/XMPPAccountStore.java
```diff
@@ -264,6 +227,15 @@ private synchronized void saveAccounts() {
     LOG.debug("saved " + accounts.size() + " account(s)");
   }
 
+  private XStream createXStream() {
+    XStream xStream = new XStream();
+    xStream.alias("accounts", ImmutablePair.class);
+    xStream.aliasField("activeAccount", ImmutablePair.class, "left");
```
srossbach: Isn't there another way to retrieve the field name left & right in a compile safe fashion ?

core/src/saros/editor/colorstorage/ColorIDSetStorage.java
```diff
@@ -182,12 +177,15 @@ private void load() {
...
-              new ObjectInputStream(new ByteArrayInputStream(serializedData.getBytes(CHARSET)))
-                  .readObject();
+      currentAvailableSets = (List<ColorIDSet>) createXStream().fromXML(serializedData);
     } catch (Exception e) {
       LOG.error("error while loading color sets", e);
     }
   }
+
+  private XStream createXStream() {
+    XStream xStream = new XStream(new DomDriver(CHARSET));
```
srossbach: Is this the same behavior as before ? To be more specific, is there any need for a charset at all ? Because you are producing already a string.

core/src/saros/account/XMPPAccountStore.java
m273d15: I like the idea. This will simplify the XML structure (without a reference)

core/src/saros/account/XMPPAccountStore.java
m273d15: For testing purposes. Otherwise it is not possible to test the accountFile format.

core/src/saros/account/XMPPAccountStore.java
m273d15: I will add a check whether the file is smaller than MAX_ACCOUNT_DATA_SIZE

core/src/saros/account/XMPPAccountStore.java
m273d15: It seems as if xstream does not support a compile safe way.

core/src/saros/editor/colorstorage/ColorIDSetStorage.java
m273d15: I will check this. Thank you, for the review

m273d15: Done

core/src/saros/account/XMPPAccountStore.java
m273d15: Done

core/src/saros/account/XMPPAccountStore.java
m273d15: Done

core/test/junit/saros/account/XMPPAccountStoreTest.java
```diff
@@ -89,26 +89,16 @@ private void writeAccountFile(File accountFile, String key, String content) thro
   }
 
   private String createAccountFileContent(
-      XMPPAccount activeAccount, List<XMPPAccount> otherAccounts) {
+      XMPPAccount activeAccount, ArrayList<XMPPAccount> configuredAccounts) {
```
srossbach: I do not get idea behind this test. Do you not trust XStream ? What is wrong with creating an XMPPAccountStore, add some accounts and save it. Afterwards you create another XMPPAccountStore and use the existing file. I guess such a test already exists.

core/test/junit/saros/account/XMPPAccountStoreTest.java
```diff
@@ -89,26 +89,16 @@ private void writeAccountFile(File accountFile, String key, String content) thro
...
 
+    Integer index = configuredAccounts.indexOf(activeAccount);
```
srossbach: It is "int" and not "integer" unless you want to return null. I have seen this multiple times in this patch set.

core/test/junit/saros/account/XMPPAccountStoreTest.java
m273d15: The idea was to inform a developer if he/she changes the XML format/content of the file and therefore invalidates our old files.

core/test/junit/saros/account/XMPPAccountStoreTest.java
m273d15: Done

core/test/junit/saros/account/XMPPAccountStoreTest.java
srossbach: You can also use Aliases to allow variable renaming in the future. And I also think adding a comment to the affected classes (e.g if you change XYZ you will broke ABC) is much more helpful than only have one test case that is failing.

m273d15: I will add these additional information to the corresponding lines.

m273d15: Done

PR #493, opened 2019-04-11
[INTERNAL][CORE] Reworked XMPPAccount serialization

srossbach:
The accounts are now serialized and deserialized using a custom
mechanism instead of the Java Serializable option. This ensures that the
account data can be restored even if the class names or their packages
changes.

### DIFF
--- a/core/src/saros/account/XMPPAccount.java
+++ b/core/src/saros/account/XMPPAccount.java
@@ -1,5 +1,10 @@
 package saros.account;
 
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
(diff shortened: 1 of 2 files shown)


srossbach:
This patch is not final. You can chose if your patch fits better to solve the problem or this one.

m273d15:
@srossbach If it is ok for you I would use my PR. Your solution is minimalistic, but using the XML will reduce the complexity of the code (I can substitute the write size, write payload, write size, write payload code with write xml).

srossbach:
I would still prefer to remove the redundant saving of the active account and use and index instead that points to the active account.

m273d15:
#491 is merged.

PR #494, opened 2019-04-11
[INTERNAL][CORE] Move User Preferences to User Object

stefaus:
Integration of the User Preference Store into the User Object.
1. this removes the duplicate storage of user color
2. removing (unlikely but possible) data races where a user has already been added to a session but the preference store is not set or user already removed
3. removes some code duplication

### DIFF
--- a/core/src/saros/negotiation/NegotiationFactory.java
+++ b/core/src/saros/negotiation/NegotiationFactory.java
@@ -213,7 +213,7 @@ public final class NegotiationFactory {
       throw new IllegalStateException("User <" + user + "> is not part of the session.");
     }
 
-    String type = session.getUserProperties(user).getString(ProjectNegotiationTypeHook.KEY_TYPE);
+    String type = user.getPreferences().getString(ProjectNegotiationTypeHook.KEY_TYPE);
     if (type.isEmpty()) {
       throw new IllegalArgumentException("Missing TransferType for User: " + user);
     }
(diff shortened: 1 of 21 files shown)


stefaus:
> The default value for favoriteColorId is -1 and not 0 (which is returned by the preference store)

Yes this is changed to 0 now, but did you find a problem with this different handling?

> The User class now has references to some "external" negotiation hooks packages.

The negotiation hook is in the same package (which is questionable) and the User also provides color information. Before the session had these references, which in my opinion is more questionable. I think this approach is more OOP. 

> You can now just override the color properties without any warning / internal use hint.

But you would probably look at references of `ColorNegotiationHook.KEY_INITIAL_COLOR` first. Yes working with a key/value store can lead to errors, but that's why we do reviews..

stefaus:
> Sorry for the late response but I do not got notified about the V15 release. Can you please change at least the default values to -1 ?

Okay, I'll look into it, but this could take a while. I'll be busy in the next weeks.

stefaus:
@srossbach You're pointing at this default setting, right?
https://github.com/saros-project/saros/blob/fe4f99e48cd77267e161ef746aa1149b3b8a3b4d/core/src/saros/preferences/PreferenceInitializer.java#L36
As far as I've found a `User` is always created with colors provided by the `ColorNegotiationHook` / `UserListEntry`.
And  `ColorNegotiationHook` is using the default from `PreferenceInitializer`.

So from my point of view this patch doesn't change behavior, only now some test cases use 0.
I could add `this.preferences.setDefault` to the User CTOR but as the default is already loaded I don't see the benefit (only for testcases that doesn't use the hooks). What do you think? Did I miss something here?

stefaus:
rebased, only updating picocontainer import

PR #499, opened 2019-04-16
[FIX][I] #479 Correctly trigger an active negotiation cancel

### LINKED ISSUE #479: Don't display an error notification when the remote user canceled the negotiation
Currently, the host displays an error notification when the remote user actively cancels the negotiation. This is confusing as it is an expected state and not an error. A notification of type 'information' (or 'warning' to draw more attention) would be more suitable.

The logic dispatching the notifications is located in [`NegotiationHandler`](https://github.com/saros-project/saros/blob/master/intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java) in [l.157](https://github.com/saros-project/saros/blob/master/intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java#L157) and [l.228](https://github.com/saros-project/saros/blob/master/intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java#L228). The logic has to be adjusted to differentiate between the different remote error causes and the dispatch the right kind of notification.

tobous:
Adjusts the logic to pass 'null' as the reason when the cancellation was
actively triggered by the user.

Fixes #479.

### DIFF
--- a/intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
+++ b/intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 import org.apache.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import org.picocontainer.annotations.Inject;
 import saros.filesystem.IChecksumCache;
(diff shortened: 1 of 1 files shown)


tobous:
Commit currently also contains a name change of a package-private method to check if the codacy rule disallowing the package-private modifier is still active.

tobous:
Removed test changes.

tobous:

PR #504, opened 2019-04-18
[BUILD][CORE] Upgrade EasyMock, PowerMock and jUnit

tobous:
Update to PowerMock 2.0.0. This should theoretically add support for
java 9, but from initial testing it looks like the java 9 support is
still a bit buggy.

Updates EasyMock to 3.5.1 as this is the latest release officially
supported by PowerMock 2.0.0.

Subsequently updates jUnit to 4.12. This could not be done previously as
the old mocking libraries required 4.11 or earlier.

Adjusts the test logic initializing Capture objects as they now require
a 'newInstance()' call instead of calling the CTOR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -99,7 +99,7 @@ configure(projectsToConfigure) { projectToConf ->
     }
   }
 
-  projectToConf.ext.junitVersion = 'junit:junit:4.11'
+  projectToConf.ext.junitVersion = 'junit:junit:4.12'
   projectToConf.ext.log4jVersion = 'log4j:log4j:1.2.15'
 
   dependencies {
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -104,8 +104,10 @@ configure(projectsToConfigure) { projectToConf ->
     }
 
     testConfig junitVersion
-    testConfig 'org.easymock:easymock:3.1'
-    testConfig 'org.powermock:powermock-easymock-release-full:1.5.1'
+    testConfig 'org.easymock:easymock:3.5.1'
+    testConfig 'org.powermock:powermock-core:2.0.0'
```
tobous: To be honest, I am not really sure if all of these powemock imports are really needed. I just added the ones that looked like they would be required....

PR #507, opened 2019-04-25
Release Notes for Saros 15.0.0

m273d15:


m273d15:
This PR contains my first draft of a release note of 15.0.0.
I want to keep these notes as short as possible. Otherwise, nobody reads them.

docs/release-notes/saros-e_15.0.0.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Upgrade Instruction
+We changed the structure of our configuration files and the location of our update-site.
+Therefore, if you used the previuous Saros version, please:
+* Remove your saros directory at `$HOME/.saros`
+* Change the update site to: <UPDATE-SITE>
+
+## Where is the whiteboard?
+You may know the Saros whiteboard from the previous release.
+Due to stability issues we did not released the Whiteboard feature of Saros, which allowed to share drawings between Saros users.
```
m273d15: Typo: `released`

docs/release-notes/saros-e_15.0.0.md
```diff
@@ -0,0 +1,29 @@
...
+
+Since the last release of Saros for Eclipse we mostly restructured, refactored and reworked the project in more than 1000 commits. From now on we want to use a lot shorter release cycles in order to
+provide fixes and features faster.
+
+Saros for IntelliJ 0.1.0 is the first alpha release version of our IntelliJ version. Please, see the [Saros for IntelliJ release notes](saros-i_0.1.0.md) before you use the new plug-in.
+
+## Upgrade Instruction
+We changed the structure of our configuration files and the location of our update-site.
+Therefore, if you used the previuous Saros version, please:
+* Remove your saros directory at `$HOME/.saros`
```
srossbach: Well I guess we will have to convert the account store as we did always when we changed things.

docs/release-notes/saros-e_15.0.0.md
```diff
@@ -0,0 +1,29 @@
...
+The Saros Team is happy to announce Saros for Eclipse 15.0.0 and Saros for IntelliJ 0.1.0.
+
+Since the last release of Saros for Eclipse we mostly restructured, refactored and reworked the project in more than 1000 commits. From now on we want to use a lot shorter release cycles in order to
+provide fixes and features faster.
+
+Saros for IntelliJ 0.1.0 is the first alpha release version of our IntelliJ version. Please, see the [Saros for IntelliJ release notes](saros-i_0.1.0.md) before you use the new plug-in.
+
+## Upgrade Instruction
+We changed the structure of our configuration files and the location of our update-site.
+Therefore, if you used the previuous Saros version, please:
```
m273d15: Add information that the old plugin has to be de-installed.

docs/release-notes/saros-e_15.0.0.md
m273d15: Yes, this is the solution which is cleaner, but I am not sure whether it is worth the effort. I would expect that the benefit of the users is small. In any case, the users have an increased migration effort with this release. The old Saros plugin also has to be de-installed (we changed the names of the OSGI bundles) before the new version can be installed. Otherwise, the eclipse instance contains two installations.

stefaus: I'm Ok with not converting the config this time and as nobody volunteered at the last Stand-Up it's more or less up to @srossbach if he wants to implement this.

Nevertheless (on my system) if Saros can't read the config file it just rans into a silent exception and nothing else happens. So is it really needed to delete the file for Upgrading?
While locating the folder on Linux is easy, I think some Windows people will have to search for it.

docs/release-notes/saros-e_15.0.0.md
```diff
@@ -0,0 +1,29 @@
+---
+title: Saros Release Notes
+---
+
+The Saros Team is happy to announce Saros for Eclipse 15.0.0 and Saros for IntelliJ 0.1.0.
+
+Since the last release of Saros for Eclipse we mostly restructured, refactored and reworked the project in more than 1000 commits. From now on we want to use a lot shorter release cycles in order to
+provide fixes and features faster.
+
+Saros for IntelliJ 0.1.0 is the first alpha release version of our IntelliJ version. Please, see the [Saros for IntelliJ release notes](saros-i_0.1.0.md) before you use the new plug-in.
```
tobous: It seems weird to mention Saros/I in the Saros/E release notes. I think this remark should be removed.

tobous: Also, the link does not work.
```suggestion
Saros for IntelliJ 0.1.0 is the first alpha release version of our IntelliJ version. Please, see the [Saros for IntelliJ release notes](saros-i_0.1.0.html) before you use the new plug-in.
```

docs/release-notes/index.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Release Notes
+---
+
+## Latest:
+* **Saros for Eclipse (Saros/E)**: [15.0.0](saros-e_15.0.0.md)
```
tobous: This link does not work in my local setup. Change to `html` instead of `md`? Same goes for link to Saros/I release notes.
```suggestion
* **Saros for Eclipse (Saros/E)**: [15.0.0](saros-e_15.0.0.html)
```

docs/release-notes/saros-e_15.0.0.md
m273d15: Ok, it is not necessary to delete the file. The user only sees that the account store is empty/no account is configured. If the user configured a new account, the invalid file is removed and the new account store is valid. Therefore no action is required, but I will add the information that the accounts have to be re-configured.

docs/releases/saros-e_15.0.0.md
```diff
@@ -0,0 +1,27 @@
+---
+title: Saros Release Notes
+---
+
+The Saros Team is happy to announce Saros for Eclipse 15.0.0 and Saros for IntelliJ 0.1.0.
```
tobous: Also remove mention of Saros/I?

m273d15: Ah, sorry I forgot this comment. I will remove this.

PR #509, opened 2019-05-02
[INTERNAL][I] Remove implementation of IWorkspace implementation

tobous:
Removes the implementation of IntelliJWorkspaceImpl.getProject(String)
and IntelliJWorkspaceImpl.getLocation() as they don't make sense in an
IntelliJ context. Replaces them with throwing an unsupported operation
exception and detailed explanations as part of the javadoc. Also marks
the methods as deprecated.

### DIFF
--- a/core/src/saros/filesystem/IWorkspace.java
+++ b/core/src/saros/filesystem/IWorkspace.java
@@ -35,7 +35,19 @@ public interface IWorkspace {
   public void run(IWorkspaceRunnable runnable, IResource[] resources)
       throws IOException, OperationCanceledException;
 
+  /**
+   * @deprecated the concept of a central workspace for the IDE instance does not apply to all IDE
+   *     models (see the IntelliJ project/module model). This method will be removed from the
+   *     interface in a future patch.
+   */
+  @Deprecated
   public IProject getProject(String project);
 
+  /**
+   * @deprecated the concept of a central workspace for the IDE instance does not apply to all IDE
+   *     models (see the IntelliJ project/module model). This method will be removed from the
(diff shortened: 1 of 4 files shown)


tobous:
Rebased onto current base PR without interaction.

tobous:

tobous:
With this PR I would like to work towards removing the methods `getProject` and `getLocation` from the IWorkspace interface. The concept does not work very well for IntelliJ and the methods are currently only used by the Server implementation.

@m273d15 @srossbach @Drakulix Opinions?

Drakulix:
> With this PR I would like to work towards removing the methods `getProject` and `getLocation` from the IWorkspace interface. The concept does not work very well for IntelliJ and the methods are currently only used by the Server implementation.
> 
> @m273d15 @srossbach @Drakulix Opinions?

If it is not used by any implementation other then the Server, why not mark it `@deprecated` in the core as well? This way no one should rely on this by accident, because it emits at least a warning.

I think I would even prefer this PR to remove those methods globally and make this something server-specific to avoid migration issues, but I get why you want to do this step-by-step.

tobous:
Marking the methods as deprecated is a good idea.

I already had a look at what it would take to remove the methods from the interface a while back. The server uses the methods in multiple places, always accessing the object through the IWorkspace interface, which makes it kind of messy to remove the methods from the interface. A restructuring of the affected classes is probably needed to remove the methods from the interface cleanly.

tobous:

tobous:

tobous:

PR #510, opened 2019-05-02
Prepare to remove project from plugin context (4 of X)

tobous:
Continuation of #451, #452, and #453.

#### [INTERNAL][I] Remove EditorAPI dependency on the shared project

Adjusts the EditorAPI methods used to modify editor contents to use the
passed project reference for the command processor instead of the
project object held by the session context.

Adjusts the usage of the adjusted methods to request the matching
project object from the module the resource belongs to.

#### [REFACTOR][I] Make EditorAPI static
Makes the class EditorAPI static as it is only a utility class that does
not hold a state. Subsequently removes the class from the session
context.


tobous:
Rebased multiple times onto different stages of base PR with minor interaction to solve import related merge conflicts.

tobous:

PR #511, opened 2019-05-02
Prepare to remove project from plugin context (5 of X)

tobous:
Continuation of #451, #452, #453, and #510.

#### [REFACTOR][I] Move session context factory to intellij package

Moves SarosIntellijSessionContextFactory to the package
'intellij.context'. This was done as the class contains Saros/I specific
dependencies, meaning it can not be moved to the core. For this reason,
it should not remain in the 'core' package.


tobous:
Rebased multiple times onto different stages of base PR with minor interaction to solve import related merge conflicts.

tobous:
Rebased onto master without any interaction.

PR #515, opened 2019-05-08
Redirect update-site via GitHub Pages

m273d15:
Instead of hosting the update-site in our main documentation
the usage of a p2 composite repository allows us to redirect
to the location of our update-site. Therefore the official
update-site becomes:
www.saros-project.org/update-site/eclipse


docs/update-site/eclipse/site.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<site>
+   <description url="https://www.saros-project.org/update-site/eclipse">
+      Site to install the Saros Feature for Distributed Party Programming
+   </description>
+   <feature url="https://raw.githubusercontent.com/saros-project/update-site-artifacts/master/update/saros.feature_14.10.31.jar" id="saros.feature" version="14.10.31">
```
m273d15: urls are still wrong!

docs/update-site/eclipse/compositeArtifacts.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<?compositeArtifactRepository version='1.0.0'?>
+<repository name='Saros Releases Repository' type='org.eclipse.equinox.internal.p2.artifact.repository.CompositeArtifactRepository' version='1.0.0'>
+    <properties size='0'>
+    </properties>
+    <children size='1'>
+      <child location='https://saros-project.github.io/update-site-artifacts'/>
```
tobous: We could benefit from having version specific update sites in the future, at which point it would be nice if the basic structure of the update site artifact repository would reflect that (e.g. by splitting it into folders for different versions). But this is nothing we have to worry about right now.

PR #516, opened 2019-05-08
[DOC] Add favicon for GitHub pages

tobous:


m273d15:
I am not sure whether this could cause an issue:
Our ico file has the size `232x256`. There is no standard favicon size, but usually, the favicon has a quadratic size (e.g. 16x16, 32x32).

I created an [alternative icon (I had zip this file in order to upload it)](https://github.com/saros-project/saros/files/3157030/saros_favicon.zip) which is compressed (file size of 6.3kb instead of 240kb) with a size of 256x256.

tobous:
Updated the PR to use the favicons generated by https://realfavicongenerator.net/ instead.

tobous:
Rebased onto current master and fixed a typo for a color in head.html

Issue #469, opened 2019-03-13
The "Work Together On.." dropdown list contains only the first project


m273d15:
The dropdown list of the "Work Together On.." context menu in the saros view contains only the first project instead of all available projects.

E.g. If the project tree contains three projects `A-project`, `B-project`, `C-project` only the project `A-project` is contained in the dropdown list.

srossbach:
https://stackoverflow.com/questions/12315749/e4-dynamic-menu-contributions

srossbach:
https://github.com/saros-project/saros/blob/master/eclipse/src/saros/ui/menuContributions/StartSessionWithProjects.java is the affected class. I checked the class and the logic is working. So I guess it is an Eclipse issue that does not properly show the context menu entries.

m273d15:
It seems as if #473 is required before using dynamic context menus in the eclipse 4 environment. For now I would remove the dynamic list and replace the "Work Together On.." entry with a "Share Project(s).." entry that opens the dialog for sharing projects.

m273d15:
Implemented a workaround in #521. After migrating to e4 we should re-implemented these dynamic menus!

PR #521, opened 2019-05-13
Avoid dynamic context menus

m273d15:
Workaround #469 (The "Share With" menu has the same issue)

During the migration to Eclipse 4.x, the dynamic context menus were broken.
Only the first entries of these menus are shown.
Therefore this change opens the dialog that allows to
select all project which can be shared.


srossbach:
How many STF tests are going to fail after this change is commited ?

m273d15:
No STF test is failing, but I have to adjust the `WorkTogetherOnContextMenu` which is used in the STF self-tests.

eclipse/plugin.xml
```diff
@@ -956,71 +956,27 @@
...
             allPopups="false"
             locationURI="popup:org.eclipse.ui.popup.any?after=additions">
-            <menu
-                  icon="icons/elcl16/session_tsk.png"
-                  label="Share With"
-                  tooltip="Share the Selected Projects With Your Contacts">
-                  <visibleWhen
+         <command
+               commandId="saros.ui.commands.ShareResource"
+               disabledIcon="icons/dlcl16/spacer.png"
```
tobous: Is there a reason to use this spacer instead of the real icon when the menu entry is disabled?

eclipse/plugin.xml
```diff
@@ -1046,9 +1002,12 @@
       <menuContribution
             allPopups="false"
             locationURI="popup:org.eclipse.ui.popup.any?before=menustart">
-         <menu
+         <command
+               commandId="saros.ui.commands.ShareResource"
+               disabledIcon="icons/dlcl16/spacer.png"
```
tobous: Same as above.

eclipse/plugin.xml
m273d15: I simply used the same configuration as in the main menu. I am not 100% when the disabledIcon is used. If our constrains (as a running session has to exist) are not satisfied, [the normal icon will simply appear greyed out](https://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fextension-points%2Forg_eclipse_ui_viewActions.html):

screenshot

PR #524, opened 2019-05-16
[DOC] Update Readme Text and add Animation

stefaus:


stefaus:
To watch a better preview goto https://github.com/saros-project/saros/tree/pr/update-readme

PR #527, opened 2019-05-21
Adding conditional testing in the test battery

suzanap:
The idea of conditional testing is to test first  Saros' main functionalities and then to test the rest. 
For Saros the main functionality is seting up a session with 2 users and 3 users. Some other important functionalities: FollowMode, ConcurrentEditing and EditDuringInvitation. If the tests for the main functionalities succeeded, then the rest will be tested; if not,the rest will be skipped. 

About implementation: 
- In the StfTestCase a TestWatcher is implemented, which monitors all results of the tests and stores in an arraylist the name of test classes, whose test methods failed. Some functions for checking if the tests for the main functionalities passed.
- In the "rest" of the test classes, in @BeforeClass it is checked if the tests for the main functionalities passed with the help of Assumptions. If the assumption is violated, then the test will be skipped. The tests will NOT be skipped, if the tests for the main functionalities are not run.


m273d15:
Your code does not comply with the formatting rules. Please, use the [google java formatter](https://www.saros-project.org/contribute/development-environment.html#configure).
Same for #528

stf/src/saros/stf/client/StfTestCase.java
```diff
@@ -27,600 +26,754 @@
...
+    }
 
-    throw exception;
-  }
+    /**
+     * Checks if the test for establishing a session between two users passed
+     */
+    public static boolean checkIfShare2UsersSequentiallySucceeded() {
+        for (String failedTest : allFailedTests) {
+            if (failedTest.contains("Share2UsersSequentiallyTest")) {
```
srossbach: I am deeply sorry but this is not a good solution because it is error prone by design.

suzanap: With the help of a TestWatcher, the name of the test class is written in allFailedTests if at least one of the tests in this test class failed.  You mean it is error-prone because you cannot rely on the TestWatcher? Could you please give more details?

m273d15:
@suzanap I see the following issues

As far as I understand your approach we have two aspects:
* **a)** The test conditions - If test A depends on B and B already failed then A is skipped.
* **b)** The test order - The previous technique works only if it is guaranteed that B is executed before A is executed.

### issues with a)
* The dependency implementation is not compile save. (e.g. Renaming the class `Share2UsersSequentiallyTest` breaks the logic in `checkIfShare2UsersSequentiallySucceeded`, because the class is referenced via a constant String. The compiler would not detect this issue because there is no relation between the actual class and the String)
* `Assume.assumeThat(checkIfShare2UsersSequentiallySucceeded())` is confusing and it is not really clear that you model dependencies with this method.

**Possible fixes:**
* Instead of collecting the return value of `getClassName()` you collect the values of `getTestClass()` in the `TestWatcher`.
* Instead of using `Assume.assumeThat(checkIfShare2UsersSequentiallySucceeded())` you could introduce a method that receives a list of test classes in order to define the dependencies.

### issues with b)
If you already model dependencies via a) it should not be necessary to maintain a second dependency model via the test class order within the test suites.

**Possible fixes:**
The dependencies used in a) should be used to define the execution order. 

### Usability
Even if the current test set works and all dependencies are defined, it is not obvious for developers that new tests should be embedded into this additional logic/dependency graph.

**Possible fixes:**
* Inform users during test execution if a test is executed that has no dependency information.
* Add a function like "noConditions" method that marks that test has no dependency information, because the test has no dependencies. This method should suppress the warning that no dependencies are listed in the test case.

----------------

Are these changes possible?
If the changes are not possible, could you explain why not?

suzanap:
@m273d15 

Point **a)** can be solved without a problem.
Point **b)** presents a problem that cannot be fixed. The dependencied defined via a) are defined in the test class itself. In order to define the order of the tests, you need to access the dependency within the class, but this dependency is not initialized yet. A possible workaround is:
- have a file with all dependencies: It could look like this: 
       `TestClassA:TestMethod or TestClassA:TestClassB `(depending if the test class depends on just a test 
        method or a test class that has more test methods)
-look for all tests in `stf.test` (the test name ends in "Test")
-use the dependencies defined in the file for defining the order of the tests in the TestSuite and for checking in the test class itself if the test, on which the test class depends on, passed or failed.

This way it is possible to have an overview over all the dependencies of the tests. 

To **Usability**: 
-An additional message with be logged when starting the test, saying if a test depends on another (Warning), or that it is an independent test (Info). 

Is this solution a good solution?

m273d15:
Maintaining an additional dependency file is also not compile save and introduces more complexity in our infrastructure.
Do you think the saved time during the test execution is worth the effort of implementing this handling as well as maintaining the additional complexity in the project?

suzanap:
Implementing this solution should not be a problem; the only problem is the complexity it adds to the project. From my prospective, it does not increase the complexity very much; but I might be biased, since I proposed the solution. The benefits of this solution are to be seen, if setting up a Saros session does not work properly, or adding files to a shared project does not work. However, I don't know exactly  how often this occurs in order to tell if this solution is worth it inspite of the increase of the complexity in the project.

m273d15:
@suzanap As discussed: I close this PR, because of the complexity which would be introduced by the change and the unknown benefit.

PR #528, opened 2019-05-21
set up equivalence classes

suzanap:
The idea of "set up equivalence classes" is to group some tests according to their setup and to set up & tear down the environment only once for the group. 
In Saros test battery there are at least 4 main groups, according to the number of testers needed. The set-up is mainly the Saros session.
However, not all tests from group "Alice&Bob", for example, need the same set up. Therefore in each group there will be independent tests (that need special set up, not just a Saros session), ONE test that will set up the environment properly (a Saros session with an empty project) and ONE test that will tear down the environment properly (clean up workspace, disconnect testers). In the group Alice&Bob&Carl there are two such subgroups: one for a session for all 3 testers and one for a session for only 2 testers. The runtime of the tests is 25% less. (27min vs 36 min) 

About implementation: 
- there are 2 methods for intializing the tests: selectFirst(for idependent tests), select (for the tests whose environment is already set up) and 2 methods for tearing down the
environment: tearDownSaros(for tests whose enviroment does not need to be torn down because another test depends on it) and tearDownSarosLast (for independent tests). 
- for the tests, whose environment is only once set up: before running the test methods, it is ensured that the testers are connected, the Saros session exists and the empty project is shared. After running the tests, it is proved if other tests were run in the test suite or if it is run alone. If it is run alone, then the tearDownLast method is called. This is done with the help of a test watcher that saves up in an arraylist the names of the test classes that were run. If the current test class is the only one in this array, then the tearDownLast method is called. 
- there is a test suite for each category: Alice, Alice&Bob, Alice&Bob&Carl, Alice&Bob&Carl&Dave. The order of the tests is important (since each test in one group should not tear down the environment.)

### DIFF
--- a/stf.test/test/saros/stf/test/AliceAndBobAndCarlAndDaveTestSuite.java
+++ b/stf.test/test/saros/stf/test/AliceAndBobAndCarlAndDaveTestSuite.java
@@ -1,0 +1,15 @@
+package saros.stf.test;
+
+import org.junit.runner.RunWith;
+import org.junit.runners.Suite;
+import saros.stf.test.invitation.ParallelInvitationWithTerminationByHostTest;
+import saros.stf.test.permissions.AllParticipantsFollowUserWithWriteAccessTest;
+
+@RunWith(Suite.class)
+@Suite.SuiteClasses({ AllParticipantsFollowUserWithWriteAccessTest.class,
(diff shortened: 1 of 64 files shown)


m273d15:
@suzanap As discussed: I close this PR. @srossbach already highlighted some issues.

PR #536, opened 2019-05-28
[FIX][I] #535 Use theme color for toolbar

### LINKED ISSUE #535: Toolbar color does not adjust to IDE theme
When changing the IDE theme, the color of the toolbar in the Saros view is not adjusted accordingly.
saros_i_wrong_toolbar_color

This issue is only present in IntelliJ IDEA 2019.1 and newer. With previous releases (tested in 2018.3.X), the color of the toolbar is adjusted correctly. This might be connected to the introduction of the improved theme engine with release 2019.1.

tobous:
Uses the theme colors provided by IntelliJ to set the foreground and
background color for the toolbar.

Fixes #535.

### DIFF
--- a/intellij/src/saros/intellij/ui/views/SarosToolbar.java
+++ b/intellij/src/saros/intellij/ui/views/SarosToolbar.java
@@ -1,6 +1,7 @@
 package saros.intellij.ui.views;
 
 import com.intellij.openapi.project.Project;
+import com.intellij.ui.JBColor;
 import java.awt.FlowLayout;
 import javax.swing.JToolBar;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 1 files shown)


srossbach:
Can you explain why you have to set the color manually. Looking at the issue entry and your provided picture every other UI widget fits the selected theme without any modifications.

tobous:
@srossbach No, I can't. I have no idea why this component behaves differently.

I first thought that maybe the color was set manually somewhere, but I couldn't find anything. Also, it is highly unlikely that this is the issue anyways because, as mentioned in the issue description, this issue only appeared after updating to IntelliJ IDEA 2019.1.

I was just having a short look at the available methods for JToolBar objects and the [JetBrains doc for the IntelliJ themes](https://www.jetbrains.org/intellij/sdk/docs/reference_guide/ui_themes/themes_metadata.html) and wanted to try out fixing it by setting the colors directly. It seems to work fine and also updates correctly when the theme is changed during runtime.

In general, it would be nicer to use the default button bar for IntelliJ plugin toolwindows instead of including a separate toolbar in our view, but I haven't figured out how to implement this yet (and couldn't find a documentation for this). This change would make this issue obsolete.

srossbach:
Could you just add a comment into the code that this is done this way because there is some unknown issue ?

tobous:
@srossbach done.

PR #539, opened 2019-06-03
[FIX][I] #488 Adjust module comparison for multiple open projects

### LINKED ISSUE #488: Adjust IntelliJProjectImpl to use project and module name to identify objects
Once Saros/I can work with multiple projects, the basic `toString`, `equals` and `hashCode` logic of [IntelliJProjectImpl](https://github.com/saros-project/saros/blob/master/intellij/src/saros/intellij/filesystem/IntelliJProjectImpl.java) has to be adjusted to also use project characteristics in addition to the module name to uniquely identify the object.

This is necessary as, while module names are unique in a specific project context, they are not unique in the application context. One IntelliJ application can have any number of open projects that all can contain a module with the same name. ~~The combination of project name and module name should still be unique in an application context.~~
**Update:** The project name + module name combination is also not unique as IntelliJ can have any number of projects with the same name open (which all could contain a module with the same name).
Module names are still exclusive inside a project, so we only have to add another criterion to uniquely identify a project. This can, for example, be done by using the path of the project file or the project base path.

tobous:
Adjusts IntelliJProjectImpl.equals(Object) and .hashCode() to directly
use the respective method of the held Module object. This is possible as
Module objects are singleton across the IDE lifecycle, i.e. a module
will always be represented by a single, unique Module object.

This change allows us to differentiate modules with the same name
belonging to different projects. This fixes #488.

Subsequently adjusts IntelliJProjectImpl.toString() to also include the
project the module belongs to.

### DIFF
--- a/intellij/src/saros/intellij/filesystem/IntelliJProjectImpl.java
+++ b/intellij/src/saros/intellij/filesystem/IntelliJProjectImpl.java
@@ -25,6 +25,7 @@ import saros.filesystem.IProject;
 import saros.filesystem.IResource;
 import saros.intellij.project.filesystem.IntelliJPathImpl;
 
+/** A <code>IntelliJProjectImpl</code> represents a specific module loaded in a specific project. */
 public final class IntelliJProjectImpl extends IntelliJResourceImpl implements IProject {
 
   /*
@@ -39,11 +40,11 @@ public final class IntelliJProjectImpl extends IntelliJResourceImpl implements I
 
   private static final Logger LOG = Logger.getLogger(IntelliJProjectImpl.class);
 
-  // Module names are unique (even among different projects)
+  private final Project project;
+
   private final String moduleName;
 
   private volatile Module module;
-
   private volatile VirtualFile moduleRoot;
 
   /**
@@ -66,6 +67,9 @@ public final class IntelliJProjectImpl extends IntelliJResourceImpl implements I
    */
   public IntelliJProjectImpl(@NotNull final Module module) {
     this.module = module;
+
+    this.project = module.getProject();
+
     this.moduleName = module.getName();
 
     moduleRoot = getModuleContentRoot(module);
@@ -216,8 +220,6 @@ public final class IntelliJProjectImpl extends IntelliJResourceImpl implements I
    */
   public boolean refreshModule() throws ModuleNotFoundException {
     if (module.isDisposed()) {
-      Project project = module.getProject();
-
       Module newModule = ModuleManager.getInstance(project).findModuleByName(moduleName);
 
       if (newModule == null) {
@@ -523,11 +525,29 @@ public final class IntelliJProjectImpl extends IntelliJResourceImpl implements I
     return null;
   }
 
+  /**
+   * {@inheritDoc}
+   *
+   * <p>This method operates under the assumption that module objects are handled as a singleton
+   * across the IDE lifecycle, i.e. a module will always be represented by a single, unique <code>
+   * Module</code> object.
+   *
+   * @return a hash code value for this object
+   */
   @Override
   public int hashCode() {
-    return moduleName.hashCode();
+    return module.hashCode();
   }
 
+  /**
+   * {@inheritDoc}
+   *
+   * <p>This method operates under the assumption that module objects are handled as a singleton
+   * across the IDE lifecycle, i.e. a module will always be represented by a single, unique <code>
+   * Module</code> object.
+   *
+   * @return whether the given objects is equal to this object
+   */
   @Override
   public boolean equals(final Object obj) {
 
@@ -539,11 +559,11 @@ public final class IntelliJProjectImpl extends IntelliJResourceImpl implements I
 
     IntelliJProjectImpl other = (IntelliJProjectImpl) obj;
 
-    return moduleName.equals(other.moduleName);
+    return module.equals(other.module);
   }
 
   @Override
   public String toString() {
-    return getClass().getSimpleName() + " : " + moduleName;
+    return getClass().getSimpleName() + " : " + project + " - " + moduleName;
   }
 }


tobous:
As the todo comment in the handler already suggests, this is still a work in progress. It relies on the migration to being an application level plugin, meaning this PR will have to be adjusted once the migration is finished.

intellij/src/saros/intellij/eventhandler/project/ProjectClosedHandler.java
```diff
@@ -0,0 +1,81 @@
...
+  private final ISarosSessionManager sarosSessionManager;
+  private final ISarosSession sarosSession;
+
+  private final MessageBusConnection messageBusConnection;
+
+  @SuppressWarnings("FieldCanBeLocal")
+  private final ProjectManagerListener projectManagerListener =
+      new ProjectManagerListener() {
+        @Override
+        public void projectClosing(@NotNull Project project) {
```
srossbach: Which thread calls this method ?

tobous:
Removed logging statement that was only added for testing purposes.

intellij/src/saros/intellij/eventhandler/project/ProjectClosedHandler.java
tobous: This is run on the swing dispatcher thread: It is triggered by the [ProjectManagerImpl](https://github.com/JetBrains/intellij-community/blob/4fbb6152ee1391b117b85b464107bb87de23f6ab/platform/platform-impl/src/com/intellij/openapi/project/impl/ProjectManagerImpl.java#L701), which [asserts that the current thread is the dispatcher thread](https://github.com/JetBrains/intellij-community/blob/4fbb6152ee1391b117b85b464107bb87de23f6ab/platform/platform-impl/src/com/intellij/openapi/project/impl/ProjectManagerImpl.java#L665). (I also double-checked this by adding the same assert statement to the listener method.)

But just checking the *current* implementation is somewhat risky as I couldn't find anything in the documentation guaranteeing that this will always be the case, meaning they could change this internal behavior at some point.

Maybe we should add such assertions to sections of our code as well if we rely on the execution being done on the dispatcher thread and the execution itself is not triggered by us directly. This would make it much easier to detect changes to the internal IntelliJ logic that would break parts of our logic.

srossbach: The problem is that stopSession should not be called on the dispatch thread. It might be possible that are activities are executed in a dead session. The shutdown is still problematic anyways as we do not know then the PicoContainer calls stop on the ActivityHandler.

tobous: Hmm ok, but we somehow have to deal with the project being closed. Even if there still are activities pending for a dead session, with the project closed, I don't think we have any way of applying them. (And even if there were a way of interacting with a closed module / a module of a closed project, I am not sure if that is something that we really want to use.)

I changed this to be a synchronous call as I was running into the mentioned problem with it being synchronous: The project object was already disposed (meaning the project was closed and torn down) but the session context was still alive and there were still pending activities. This caused exceptions as the pending activities were trying to use the disposed project object (as it is used for the hashcode/equals implementation of the module representation). This could be 'avoided' by saving the used values in the `IntelliJProjectImpl` directly instead of requesting them from the project, but the base of the problem (applying activities to closed modules) is still present. Furthermore, this causes problems when the used values are changed during runtime as the saved values would not be updated accordingly.

This change "solved" the issue in my trivial tests, but that is no indication that there isn't still the possibility of a race condition here.

The problem is that currently, I am not really able to test this well as Saros can't handle multiple open projects, meaning I can't test the long-term effects of the current implementation as the plugin is always torn down shortly after this is called (since the IntelliJ instance stopped). (Furthermore, doing manual tests for concurrency issues alone on one computer is somewhat complicated :wink:.)

srossbach: It is simply a design decision to start / stop the session from a non UI thread. Many components are coded with this contract. I know your intention to stop the session in this listener. As I do not have deep knowledge on this IntelliJ stuff there could be two approaches.

1. Shutdown the session in another thread and run the AWT/IntelliJ event loop in the meantime (I do not know if you are allowed to do that, i.e accessing the AWT EventQueue directly or IntelliJ offers such functionality). I also do not know how this work with different modal contexts. 

The other option would be to implement a IntelliJ callback for the ActivityHandler which discards activities for disposed projects and then shutdown the session in the background while you are performing a project switch.

tobous:
Rebased onto current master without any interaction. Did not address review comments yet. Still WIP.

intellij/src/saros/intellij/eventhandler/project/ProjectClosedHandler.java
tobous: I don't think using the first approach is an acceptable solution as blocking the AWT event dispatcher thread blocks the entire IDE.

I considered the second option, but this just moves the problem: At least with the current implementation, filtering for modules of disposed project requires us to compare IProject objects, which in turn tries to call the IntelliJ project of the held module (which is what we are trying to avoid in the first place).
Furthermore, the session uses `hashCode` on the contained `SPath` when receiving/parsing activities, which also leads to exception as it subsequently calls `hashCode` on the `IProject` object.

It seem like I'l have to adjust the implantation anyway to somehow make it (IntelliJ) project object independent, at which point filtering activities in the session should (probably) work.

tobous:
I re-worked the comparison logic and removed the project closed handler. I will re-introduce  the handler in a separate patch.

intellij/src/saros/intellij/filesystem/IntelliJProjectImpl.java
```diff
@@ -523,9 +527,28 @@ public VirtualFile findVirtualFile(final IPath path) {
...
+   *
+   * @return a hash code value for this object
+   */
   @Override
   public int hashCode() {
-    return moduleName.hashCode();
+    String projectName = project.getName();
+    String projectFilePath;
+
+    if (!project.isDisposed()) {
```
tobous: Hmmm. I just found out that this causes issues when the class is used in combination with hash-comparison-based data structures like hash sets. E.g. using `HashSet<T>.contains(T)` no longer works consistently as it compares the initial hash to the current hash, which can be different if the contained project object is disposed (or the project file path changed).

I am not sure what to do about this. Requiring ever other part of the code using other set implementations seems ...bad... as the hash-based implementations are usually the ones used by default and this issue is not directly apparent when using the class.

tobous: This is also an issue with the current eclipse implementation. The resource hash code of the Eclipse sources is solely base on the path of the resource, meaning if the path changes, the hashcode changes. This leads to issues when moving partially shared resources because of [this](https://github.com/saros-project/saros/blob/master/core/src/saros/session/internal/SharedProjectMapper.java#L307) check.

This is an even bigger issue if we are allowing shared projects to be moved during the session as the completely shared projects are also held as a HashSet.

@srossbach @m273d15 @stefaus Any idea how to handle this nicely, preferably without going through the whole codebase and checking the usage of hash sets (especially when used for long-term data storage)?

srossbach: I do not get the Eclipse stuff, see https://github.com/saros-project/saros/blob/master/core/src/saros/session/internal/SarosSession.java#L887

As for the hashcode. It does not matter what hashcode you return, however it has to be always the same value.

h(a) = h(b) does not imply that a = b, however if a = b => h(a) = h(b)

You simply could return 0 although this would mess up the hashmap. So I guess returning the hashed project name is fine.

srossbach: And BTW all Eclipse resource implementations are immutable so you cannot alter the hashcode in any way.

tobous: > I do not get the Eclipse stuff, see https://github.com/saros-project/saros/blob/master/core/src/saros/session/internal/SarosSession.java#L887

Ah, ok. I was not aware of that part of the logic.

> As for the hashcode. It does not matter what hashcode you return, however it has to be always the same value.

I interpreted the javadoc of `Object.hashCode` to mean that it always has to return the same value *as long as the state used for equals does not change*. But you are right, just using the name is more sensible as hash collisions in case where there are multiple project module combinations with the same name don't matter.

> And BTW all Eclipse resource implementations are immutable so you cannot alter the hashcode in any way.

Hmm, I guess I was a bit quick to conclude that the issue (moving stuff in a partial session does not work) is with the hash values.

tobous:
Removed file path from hashcode to ensure the returned value is always constant across the lifecycle of the object.

intellij/src/saros/intellij/filesystem/IntelliJProjectImpl.java
```diff
@@ -523,9 +527,18 @@ public VirtualFile findVirtualFile(final IPath path) {
...
+   * <p>This method operates under the assumption that disposed {@link Project} objects still allow
+   * requests for the project name. This method needs to be adjusted once this assumption no longer
+   * holds.
+   *
+   * @return a hash code value for this object
+   */
   @Override
   public int hashCode() {
-    return moduleName.hashCode();
+    return Objects.hash(project.getName(), moduleName);
```
srossbach: Looking at https://github.com/JetBrains/intellij-community/blob/master/platform/platform-impl/src/com/intellij/openapi/project/impl/ProjectImpl.java#L126 it is possible to change the name. Of course this is a corner case (and I have not checked if it is possible in Eclipse too, but I guess all objects are immutable) and I do not even know when this method is called (renaming the project ?).

The default ProjectImpl class also have no equals and hashcode override, so the question is: is there only one project object available for a given project (if yes just use hash on the project object), otherwise we have to find another method that always returns the same value. However the only method that is NotNullable is getLocationHash which itself uses getName ...

tobous: *Currently* project objects *seem* to be singleton from my limited testing. But I could not find any documentation explicitly mentioning this (see [project documentation](https://www.jetbrains.org/intellij/sdk/docs/basics/project_structure.html)).

I am not sure whether basing our logic on this assumption is the best idea, especially since this assumption is not easy to check. E.g. if the logic were to be changed tomorrow so that the project objects were just cached for a while but periodically re-generated, I am not sure how we could notice this internal change without a lot of manual testing/upstream source code searching. It is hard to construct comprehensive unit-tests for this. Furthermore, we are not always getting the project from the same source (sometimes we get it from the module, other times it is given to us by a listener, etc.), which makes testing all cases even harder.

Or am I just overthinking the whole thing?

Another option would be to handle this the same way we are handling moved resources, i.e. adding a listener that removes and re-adds the projec/module with the new hash code when it was moved. But we would have to apply this recursively as the module hashcode is based on the project hashcode and the file/folder hashcode is based on the module hashcode.

tobous: >Of course this is a corner case (and I have not checked if it is possible in Eclipse too, but I guess all objects are immutable) and I do not even know when this method is called (renaming the project ?).

You can just rename the project in the [settings](https://stackoverflow.com/a/21180363/11010012), which will immediately be reflected by the project object. Sadly, I could not find a listener for this change.

The change creates the file `.idea/.name` if the name differs from the name of the project base directory, so we could theoretically react to the creation/modification of the file and react with updating the hashset, but I am not sure whether this is the best solution.

tobous:
Did not continue working on this yet. Only rebased it on the current master without any interaction.

intellij/src/saros/intellij/filesystem/IntelliJProjectImpl.java
tobous: I could not find think of a better solution approach...

I posted a question regarding the topic to the IntelliJ forum: https://intellij-support.jetbrains.com/hc/en-us/community/posts/360004300240-Immutable-hashcode-for-project-and-module-objects

Hopefully somebody there (e.g. somebody of the IntelliJ development team) can suggest a better approach or confirm/guarantee that project and module objects are singleton across their lifecycle.

tobous:
Adjusted the patch to use delegate the `equals(Object)` and `hashCode()` calls to the module object. From the discussion in the IntelliJ forum, this seems like a valid approach. There are still some uncertainties, but once these have cleared up this PR should be ready to be reviewed again.

srossbach:
Seems you were looking for this ? https://github.com/JetBrains/intellij-community/blob/master/platform/projectModel-impl/src/com/intellij/openapi/module/impl/ModuleManagerImpl.java#L666

tobous:
I am not quite sure what you mean? Could you elaborate? Is the reference to the code that ensures that module objects are singleton across the lifecycle?

srossbach:
My guess is: yes . However I am not sure.

tobous:
I will have a more thorough look at it tomorrow. But I would still like to wait on the [forum discussion](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360004300240-Immutable-hashcode-for-project-and-module-objects) mentioned beforehand.

tobous:
> Seems you were looking for this ? https://github.com/JetBrains/intellij-community/blob/master/platform/projectModel-impl/src/com/intellij/openapi/module/impl/ModuleManagerImpl.java#L666

At least when using the `Modulemanager`, the logic guarantees that the `Module` objects are singleton. Together with the confirmation from the IntelliJ developers, I think we can proceed with this patch.

I was still toying with the idea of introducing an assertion that does a sanity check for the equals on this on runtime (using the name, project name and project location hash), but I think this would be overkill.

tobous:

srossbach:
Do you need an approvement because you do not have administrator permission or are you still testing this ?

tobous:
@srossbach No, I was just waiting to see if anybody still had any comments on the matter/PR.

intellij/src/saros/intellij/filesystem/IntelliJProjectImpl.java
```diff
@@ -523,11 +525,29 @@ public VirtualFile findVirtualFile(final IPath path) {
     return null;
   }
 
+  /**
+   * {@inheritDoc}
+   *
+   * <p>This method operates under the assumption that module objects are handled as a singleton
+   * across the IDE lifecycle, i.e. a module will always be represented by a singe, unique <code>
```
m273d15:
```suggestion
   * across the IDE lifecycle, i.e. a module will always be represented by a single, unique <code>
```

intellij/src/saros/intellij/filesystem/IntelliJProjectImpl.java
```diff
@@ -523,11 +525,29 @@ public VirtualFile findVirtualFile(final IPath path) {
...
   public int hashCode() {
-    return moduleName.hashCode();
+    return module.hashCode();
   }
 
+  /**
+   * {@inheritDoc}
+   *
+   * <p>This method operates under the assumption that module objects are handled as a singleton
+   * across the IDE lifecycle, i.e. a module will always be represented by a singe, unique <code>
```
m273d15:
```suggestion
   * across the IDE lifecycle, i.e. a module will always be represented by a single, unique <code>
```

tobous:
Rebased onto current master and fixed typos.

PR #541, opened 2019-06-03
[INTERNAL][I] Show non-shareable modules in contact pop menu

tobous:
Adjusts ContactPopMenu to also show non-shareable modules. Such modules
are disabled in the menu and are displayed in a separate group.
Furthermore, the carry a tooltip explaining that/why they can't be
shared.

Subsequently removes the user notification when encountering a
non-shareable module.

This re-structuring was done to avoid annoying the user with unnecessary
notifications.

### DIFF
--- a/intellij/src/saros/intellij/ui/tree/ContactPopMenu.java
+++ b/intellij/src/saros/intellij/ui/tree/ContactPopMenu.java
@@ -32,6 +32,7 @@ import saros.net.xmpp.JID;
  * Contact pop-up menu for selecting a project and module to share. Opened when right-clicking on a
  * contact.
  */
+// TODO remove the project module check once project module can be shared
 class ContactPopMenu extends JPopupMenu {
 
   private static final Logger LOG = Logger.getLogger(ContactPopMenu.class);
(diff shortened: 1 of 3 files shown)


intellij/src/saros/intellij/ui/tree/ContactPopMenu.java
```diff
@@ -42,49 +44,108 @@
...
+
+      noModulesFoundMenuItem.setEnabled(false);
+
+      projectMenu.add(noModulesFoundMenuItem);
     }
 
+    // Show project module as non-sharable
+    projectMenu.addSeparator();
+
+    JMenuItem projectItem = new JMenuItem(project.getName());
```
srossbach: I really do not know if this is an accepted way of building menu entries, i.e creating disabled menu entries describing the entries below such an entry.

tobous: I think you misunderstood. Yes, the entry is disabled, but it is not used to describe other entries below it. Since in IntelliJ, the base of a project (containing the project config) is also just a module, I also wanted to display this module in the list. This part of the code adds an entry for the project module that is disabled and contains a tooltip explaining that the project module currently can not be shared.

srossbach: Ah ok my bad. But as this is always disabled I guess it would not hurt to do not display this entry at all. Of course that would / will require that the developers read the manual :D

PR #543, opened 2019-06-08
Prepare to remove project from plugin context (7 of 7)

tobous:
Removes the last references to the project object held in the session context. This basically marks the end of the refactoring to remove the project object from the plugin context. The rest if just a minor refactoring that can be done when the UI stuff is merged as well.

The PR should by reviewed by going through the commits. Each commit features a commit message giving more details about the content.


tobous:

tobous:
Rebased onto current base PR.

tobous:

tobous:

tobous:
Fixed commit making VirtualFileConverter static. Class was still added to session context. I either forgot to remove it in the first place or made a mistake while rebasing this PR. Either way, it works now.

m273d15:
The PR looks good to me, but making classes static makes them harder to mock. I know this is currently not really important, because we are writing no tests.
I think we should talk about what we want to test and then start bit by bit with testing.

tobous:

PR #545, opened 2019-06-08
[REFACTOR] Replace deprecated ObjectUtils methods

stefaus:
The hashCode and equals method in the ObjectUtils class from Apache
Commons was marked deprecated and they advise to use the methods from
Java standard libary introduced in Java 7. This replacement was done
using a shell script.


stefaus:
According to the [guidelines](https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing) enforced by the google formatter: 
>imported names appear in ASCII sort order

PR #549, opened 2019-06-12
Remove unused AWT UISynchronizer implementation

tobous:
Removes an AWT UI synchronizer implementation located in the Eclipse
package. This class was added to allow for compatibility with IDEs using
AWT based UIs. This most likely was forgotten when splitting Saros into
Saros/E and the Saros core. It is no longer needed as the UI
synchronizer is now the responsibility of the separate IDE
implementations.

Also renames the eclipse UI synchronizer to match the general naming pattern and to make it easier to recognize that it is the eclipse implementation.


tobous:

PR #556, opened 2019-06-28
Fix issues with SharedIDEContext.dispose()

tobous:
#### [FIX][I] Drop preregistered project object on session end

Adjusts SharedIDEContext to drop the held preregistered project object.
Otherwise, starting a second session after the first session ended
causes an erroneous warning on the host side as the preregistered
project of the first session is still present.

#### [FIX][I] Add null check to SharedIDEContext.dispose()

Adds a null check for applicationEventHandlers and projectEventHandlers
to SharedIDEContext.dispose(). This check is needed as the context can
be disposed without the handlers being initialized when the project
negotiation was canceled on the client side.


tobous:

PR #560, opened 2019-07-01
[DOC] Split user documentation for IntelliJ and Eclipse

tobous:
Splits the sections 'installation', 'getting started', and 'screenshots'
to differentiate between Saros/I and Saros/E. Adds the new pages as
sub-items in the sidebar.
This was done to avoid confusion for new users when looking at the
documentation as the existing guides only applied to Saros/E.

Moves the current pages to [PREVIOUS_NAME]-eclipse.md.

Adds the new pages installation-intellij.md and
getting-started-intellij.md. The content of the two pages was taken
from the Saros/I 0.1.0 release notes. Subsequently shortens the release
notes to just link to the new sections.
The equivalent for the screenshots page for Saros/I was intentionally
left out as there currently is no such content. Creating such content is
not really sensible in the current development stage as the UI still
changes a lot.

Subsequently adjusts all affected links to match the new file names.


docs/documentation/getting-started-intellij.md
```diff
@@ -0,0 +1,215 @@
...
+The GUI used in Saros/I is only a placeholder providing the minimal, necessary functionality.
+We are currently working on a replacement in the form of an HTML GUI that can be used across all Saros versions.
+
+## Disclaimer
+
+Saros/I does not include sub-modules when sharing a module (see [Module Restrictions](#module-restrictions)).
+As a consequence, such sub-modules might not be present for all session participants.
+If a participant deletes a shared directory that contains a sub-module in the local setup of another participant, this sub-module will be deleted without any notice.
+
+## Features
```
m273d15: We have the separate page "Features" do you think it is useful to split this page (probably in the future)?

tobous: Maybe, but I am not sure if this is really sensible. The current page states the complete Saros feature set that Saros/I is working towards. If we see in the process that the feature set is not completely transferable to Saros/I, we can have this discussion again.

The only issue in the meantime is that users could look at the feature list and then complain that something is not available in Saros/I. But this is mentioned in the release notes and the getting started section for Saros/I.

tobous:
@m273d15 Any opposition to me merging this directly? I would like to get this out there as quick as possible to avoid creating more confusion with potential users. Furthermore, if it introduces a problem in the website, it should be easily fixable.

m273d15:
@tobous No, feel free to merge

tobous:
@m273d15 I decided to preserve more of the information specific to the release (like missing features, restrictions, etc.) by still keeping it in the release notes. This means the information is duplicated for now, but this should not matter to much.

PR #564, opened 2019-07-09
Build Saros server jar in gradle

m273d15:
The task `sarosServer` now creates an executable jar in `build/distribution/server`.
Fixed a wrong error message which I found during testing the executable jar.


server/build.gradle
```diff
@@ -16,3 +16,13 @@ sonarqube {
     property 'sonar.projectVersion', '0.1.0.DEVEL'
   }
 }
+
+jar {
+  manifest {
+    attributes 'Main-Class': 'saros.server.SarosServer'
+  }
+  from {
+    configurations.compile.collect { it.isDirectory() ? it : zipTree(it) }
```
tobous: What is included in this jar? Where do you get `it` from?

m273d15: The line `configurations.compile.collect { it.isDirectory() ? it : zipTree(it) }` does the following:
* selects the configuration `compile`. This configuration contains (as defined in line 2,3) the dependency to ':saros.core' and commons-collections4. Therefore these dependencies are included in the jar (and the classes of the server).
* `collect` is the gradle/groovy `map` equivalent. Collect is a list operation. Therefore the closure of `collect` has the default iteration variable `it` (similar to the perl `$_` variable). The alternative is to write something like `{ dependency -> ...}` or `{ Clazz dependency -> ...}`, but I think this is not very common in gradle.
* The closure of collect simply extracts the content of the dependent jars in order to create a fat jar (a jar which contains the dependent classes). The method [`zipTree`](https://docs.gradle.org/current/dsl/org.gradle.api.Project.html#org.gradle.api.Project:zipTree(java.lang.Object)) just returns a gradle internal representation of a file tree.

PR #566, opened 2019-07-09
Move "AddContactView" into a page and use consistent page name

m273d15:
* Use consistent name `ShareProjectPage` instead of multiple different names
* Use consistent name `AddAccountPage` instead of multiple different names
* Add a new page `AddContactPage` and the corresponding browser functions that allow to show and close the page
* Use the new browser functions to show/close the page and use the AddContactView in the AddContactPage


ui.frontend/html/src/components/App.jsx
```diff
@@ -6,12 +6,12 @@ import BasicWidgetTestView from './BasicWidgetTestView'
...
-import StartSessionWizardView from './StartSessionWizardView'
+import ShareProjectView from './ShareProjectView'
 
 const viewComponents = {
   [views.MAIN]: MainView,
-  [views.ADD_CONTACT]: AddContactView,
-  [views.START_SESSION_WIZARD]: StartSessionWizardView,
+  [views.ADD_CONTACT_PAGE]: AddContactView,
+  [views.SHARE_PROJECT_PAGE]: ShareProjectView,
   [views.CONFIGURATION_WIZARD]: ConfigurationWizardView,
```
tobous: Now it seems a bit weird that this is still named 'Wizard'. How about either also adjusting this as part of this PR or opening an issue to refactor this later on?

ui.frontend/html/src/constants.js
```diff
@@ -1,8 +1,8 @@
 // Here the different views are specified
 const views = {
   MAIN: 'main-page',
-  ADD_CONTACT: 'add-contact',
-  START_SESSION_WIZARD: 'start-session-wizard',
+  ADD_CONTACT_PAGE: 'add-contact-page',
+  SHARE_PROJECT_PAGE: 'share-project-page',
   CONFIGURATION_WIZARD: 'configuration-page',
```
tobous: Same as in `App.jsx`. Remaining usage of 'wizard', which was now generally replaced with 'page' as far as I can tell.

ui.frontend/html/src/components/App.jsx
m273d15: Thanks, I changed the name of the constant

ui/src/saros/ui/View.java
```diff
@@ -13,9 +14,9 @@
   /** The permanently accessible view of Saros, which provides access to most of its features. */
   MAIN_VIEW(MainPage.class, "main-page", "root"),
   /** The form to add a new contact with */
-  ADD_CONTACT(MainPage.class, "add-contact", "add-contact-form"),
+  ADD_CONTACT(AddContactPage.class, "add-contact", "add-contact-form"),
   /** The session wizard dialog */
-  SESSION_WIZARD(SessionWizardPage.class, "start-session-wizard", "session-wizard"),
+  SESSION_WIZARD(ShareProjectPage.class, "start-session-wizard", "session-wizard"),
```
tobous: What about this (const + descriptions)?

ui.frontend/html/src/components/App.jsx
```diff
@@ -6,13 +6,13 @@ import BasicWidgetTestView from './BasicWidgetTestView'
 import ConfigurationWizardView from './ConfigurationWizardView'
```
tobous: What about this?

ui.frontend/html/src/components/App.jsx
```diff
@@ -6,13 +6,13 @@ import BasicWidgetTestView from './BasicWidgetTestView'
...
+import ShareProjectView from './ShareProjectView'
 
 const viewComponents = {
   [views.MAIN]: MainView,
-  [views.ADD_CONTACT]: AddContactView,
-  [views.START_SESSION_WIZARD]: StartSessionWizardView,
-  [views.CONFIGURATION_WIZARD]: ConfigurationWizardView,
+  [views.ADD_CONTACT_PAGE]: AddContactView,
+  [views.SHARE_PROJECT_PAGE]: ShareProjectView,
+  [views.CONFIGURATION_PAGE]: ConfigurationWizardView,
```
tobous: What about the class (`ConfigurationWizardView`).

m273d15:
@tobous I added the requested changes. It might be still confusing, because the term `wizard` still appears in the `Views.java`.
This is because of: A page (as `add-contact-page.html`) is represented by a file that contains a view (as `AddContactView`) and this view is implemented by a component (in the case of the `AddContactView` a `form` ). In the case of the `ShareProjectView` the view is implemented with the component `Wizard`.
This implementation detail is still in the `Views.java`, because the class is used in stf testing and requires the `id` in order to work with the elements.

I have to create some follow up refactoring PRs that:
* Move the `Views.java` into the `ui/stf` package.
* Refactor the `Wizard` component to use a custom `id`. Otherwise each view using the component has the same root id (as e.g. `ConfigurationView` and `ShareProjectView`).
* Check whether the `ViewStore` is still required. The view store is used to allow the single-page-application workflow, which is not required in our context.

ui.frontend/html/src/components/AddContactView/index.jsx
```diff
@@ -5,7 +5,7 @@ import React from 'react'
 
 
 export default
-@inject('mainUI')
+@inject('core')
```
tobous: What causes the need for this change? I can't find a matching change that renaming something from 'mainUI' to 'core'. Or am I missing something? (I have no idea what the `@inject()` call does/is used for.)

m273d15: The previous version of the AddContactView was integrated in the main view. Therefore the mainUI store was used to trigger the change of the view. Now, the core store is needed in order to call the corresponding java method.

In an upcoming PR I will rework the whole store setup.

m273d15:
@tobous I will squash the PR

PR #570, opened 2019-07-15
Fix theme-dependent UI inconsistencies

### LINKED ISSUE #550: Color of Saros toolbar popup menu items is not adjusted dynamically on runtime
The color of the Saros toolbar popup menu items is not adjusted dynamically on runtime.
As a result, if the user changes the theme after the UI/plugin was initialized, the color of the menu items is not adjusted accordingly.

This applies to all classes in https://github.com/saros-project/saros/tree/master/intellij/src/saros/intellij/ui/views/buttons.

This might be related to #535 (#536).

### LINKED ISSUE #569: Border of the Saros toolbar buttons changes when the IDE theme changes
The buttons in the Saros toolbar do not have any border by default. If the theme is changed, they gain a border (the color is dependent on the theme).

The question is what the default behavior should be. Either, we have to reset the border every time the theme changes, or we have to add a default border to match the behavior when the theme changes.

tobous:
#### [FIX][I] #550 Add workaround for toolbar popup menu color

Manually sets the color of the Saros toolbar popup menu and its entries
using the correct dynamic color references. This reference automatically
changes when the IDE theme is changes, thereby causing the menu color to
change as well.

Also replaces the usage of default swing components in this context with
the jetbrains wrapper, as recommended in the IntelliJ platform SDK.

Fixes #550.

#### [FIX][I] #569 Add default border for toolbar buttons

Adds a default border to the Saros toolbar buttons. The color of the
border is determined by the current IDE theme.

Fixes #569.

### DIFF
--- a/intellij/src/saros/intellij/ui/views/buttons/FollowButton.java
+++ b/intellij/src/saros/intellij/ui/views/buttons/FollowButton.java
@@ -1,5 +1,7 @@
 package saros.intellij.ui.views.buttons;
 
+import com.intellij.openapi.ui.JBMenuItem;
+import com.intellij.openapi.ui.JBPopupMenu;
 import com.intellij.util.ui.UIUtil;
 import javax.swing.JButton;
 import javax.swing.JMenuItem;
@@ -106,7 +108,10 @@ public class FollowButton extends ToolbarButton {
   }
 
   private void createMenu() {
-    popupMenu = new JPopupMenu();
+    popupMenu = new JBPopupMenu();
+
+    popupMenu.setForeground(FOREGROUND_COLOR);
+    popupMenu.setBackground(BACKGROUND_COLOR);
 
     menuItemPrefix = "Follow ";
 
@@ -124,7 +129,11 @@ public class FollowButton extends ToolbarButton {
 
     popupMenu.addSeparator();
 
-    JMenuItem leaveItem = new JMenuItem("Leave follow mode");
+    JMenuItem leaveItem = new JBMenuItem("Leave follow mode");
+
+    leaveItem.setForeground(FOREGROUND_COLOR);
+    leaveItem.setBackground(BACKGROUND_COLOR);
+
     leaveItem.addActionListener(e -> followModeAction.execute(null));
     leaveItem.setEnabled(currentFollowModeManager.getFollowedUser() != null);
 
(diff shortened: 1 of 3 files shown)


intellij/src/saros/intellij/ui/views/buttons/ToolbarButton.java
```diff
@@ -1,17 +1,26 @@
...
+  static final Color FOREGROUND_COLOR = JBColor.foreground();
+  static final Color BACKGROUND_COLOR = JBColor.background();
+
   /** Creates a button with the specified actionCommand, Icon and toolTipText. */
   ToolbarButton(String actionCommand, String tooltipText, ImageIcon icon) {
     setActionCommand(actionCommand);
     setButtonIcon(icon);
     setToolTipText(tooltipText);
+
+    setBorder(BorderFactory.createLineBorder(JBColor.border(), 1, true));
```
srossbach: You could also try to override the paintBorder method (doing nothing) or override the setBorder method to do nothing if you do not want any border.

intellij/src/saros/intellij/ui/views/buttons/FollowButton.java
```diff
@@ -106,7 +107,10 @@ public FollowButton() {
   }
 
   private void createMenu() {
-    popupMenu = new JPopupMenu();
+    popupMenu = new JBPopupMenu();
+
+    popupMenu.setForeground(FOREGROUND_COLOR);
```
srossbach: Is this still necessary for a JBPopupMenu ?

intellij/src/saros/intellij/ui/views/buttons/FollowButton.java
```diff
@@ -141,6 +149,9 @@ private JMenuItem createItemForUser(User user) {
 
     JMenuItem menuItem = new JMenuItem(menuItemPrefix + userNameShort);
```
srossbach: In the other class you use a JBMenuItem. Any special reason to use a JMenuItem here ?

intellij/src/saros/intellij/ui/views/buttons/FollowButton.java
tobous: Yes, as otherwise it still retains the initial color when switching the IDE theme. This is visible between the entries.

As to why all of this is necessary, I have no idea. Same as in #535, this is more like a hackie workaround than a real fix.

intellij/src/saros/intellij/ui/views/buttons/FollowButton.java
tobous: I just forgot. Fixed.

intellij/src/saros/intellij/ui/views/buttons/ToolbarButton.java
tobous: Thanks for the hint. But after comparing the two alternatives, having a border is definitely preferable to not having a border. When the buttons are disabled, they have the same color as the IDE frame in some themes, meaning they can't be differentiated from the frame. This makes it hard to understand that they actually are buttons.

tobous:

Issue #572, opened 2019-07-16
WIP - Project-bound Saros GUI not handled correctly on lifecycle bounds


tobous:
Placeholder to ensure I won't forget about this issue. Will be fleshed out later.

Project-bound GUI is not initialized correctly if the project is opened during a running session.
-> check in CTOR and set correct state
-> improve architecture to move this check to super-class (session-buttons or something like that)

Project-bound GUI is not correctly torn down when the project is closed and there is no session running for the project.
- Objects just hand around and can't be garbage collected.
-> hold list of all project-bound gui, grouped by project. dispose when project closed (and no session running?)

PR #574, opened 2019-07-18
Correctly initialize project-bound Saros UI when there is a running session

tobous:
This fixes the first half (initialization) of #572.

#### [REFACTOR][I] Remove session lifecycle listener from FollowModeAction

Removes the session lifecycle listener from the FollowModeAction.
Instead, the needed session components are now passed to the execute
command.

This was done to reduce the complexity of the action and remove the
state held by the action, which led to unexpected behaviors.

#### [INTERNAL][I] Add AbstractSessionToolbarButton

Adds AbstractSessionToolbarButton to avoid code duplication. The class
can be used by all session buttons.

Also offers the functionality to set the initial state of the button in
case there is already a running session when the button is instantiated.
This is needed as such a case is not covered by the session lifecycle
listener.

#### [FIX][I] #572 Set initial state for session buttons

Uses AbstractSessionToolbarButtons for all session buttons.

Adds the call to set the initial button state to the button CTORs.

#### [FIX][I] #572 Set initial state for session tree

Sets the initial state for the session tree contained in the Saros view
when it is instantiated. This is needed to correctly handle case where
the UI is initialized while there is already a running session.

### DIFF
--- a/intellij/src/saros/intellij/ui/views/buttons/ConsistencyButton.java
+++ b/intellij/src/saros/intellij/ui/views/buttons/ConsistencyButton.java
@@ -9,7 +9,6 @@ import java.util.HashSet;
 import java.util.Set;
 import org.apache.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
-import saros.SarosPluginContext;
 import saros.activities.SPath;
 import saros.concurrent.watchdog.ConsistencyWatchdogClient;
 import saros.concurrent.watchdog.IsInconsistentObservable;
@@ -22,8 +21,6 @@ import saros.intellij.ui.util.NotificationPanel;
 import saros.observables.ValueChangeListener;
 import saros.repackaged.picocontainer.annotations.Inject;
 import saros.session.ISarosSession;
-import saros.session.ISarosSessionManager;
-import saros.session.ISessionLifecycleListener;
 import saros.session.SessionEndReason;
 
 /**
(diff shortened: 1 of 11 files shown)


intellij/src/saros/intellij/ui/views/buttons/ConsistencyButton.java
```diff
@@ -108,15 +80,33 @@ public ConsistencyButton(@NotNull Project project) {
...
     this.project = project;
 
-    setSarosSession(sessionManager.getSession());
-    sessionManager.addSessionLifecycleListener(sessionLifecycleListener);
+    setSarosSession(sarosSessionManager.getSession());
 
     addActionListener(actionListener);
     setEnabled(false);
+
+    setInitialState();
```
srossbach: I normally would expect by that this method name that it also sets the inconsistency state, however it does not. Add a TODO ?

Drakulix:
Because most of the code is justed moved and made more streamlined, I see nothing wrong with this PR.

intellij/src/saros/intellij/ui/views/buttons/ConsistencyButton.java
tobous: Thanks for the note. I just forgot. Fixed.

tobous:
Adjusted the logic to also set the initial consistency state correctly for `ConsistencyButton`.

tobous:
Rebased onto current master. Fixed minor merge conflict in `FollowButton`.

PR #575, opened 2019-07-18
Move button text to messages properties and add initial JID server value

tobous:
#### [REFACTOR][I] Move text shown to users in buttons to messages.properties

Moves all text shown to the users when interacting with the Saros
toolbar buttons to the messages.properties file.

Fixes some unnecessary string splits.

#### [UI][I] Set saros-con server address as default for adding accounts

Sets '@saros-con.imp.fu-berlin' as the default initial field value for
the account addition wizard.

Subsequently adjusts the SafeDialogUtils to add the option to specify a
selection range for initial inputs. This option is used to set the
cursor to the start of the input field. Otherwise, the default input
would be selected by default.


tobous:
Rebased onto current master and fixed minor merge conflicts caused by previous UI patches.

PR #576, opened 2019-07-19
Dispose project-bound Saros UI to avoid memory leaks

tobous:
Correctly disposes the project-bound Saros UI so that it can be garbage collected after the project was closed.

The objects are correctly garbage collected after this change (tested with looking at the heap with the debugger).

This fixes the second half of #572.

#### [FIX][I] #572 Add logic to dispose project-bound buttons

Makes AbstractSessionToolbarButton implement Disposable (IntelliJ
interface, not PicoContainer).

The basic logic to dispose the abstract button just removes the session
lifecycle listener. The class also offers an abstract method to allow
sub-classes to disposing their components when the button is disposed.

Registers the button to be disposed when its project is disposed.

#### [FIX][I] #572 Add logic to dispose project-bound tree view

Adjusts SessionAndContactsTreeView and its components to be disposed
when their project is disposed.

#### [DOC][I] Add teardown javadoc to project-bound UI

Adds javadoc to SarosMainPanelView stating and explaining the need to
correctly tear down contained UI components in order to avoid memory
leaks.

Adds references to this javadoc to all project-bound UI components to
(hopefully) ensure that this fact is not missed when implementing future
changes.

### DIFF
--- a/intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
+++ b/intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
@@ -14,6 +14,7 @@ import org.jivesoftware.smack.RosterEntry;
 import org.jivesoftware.smack.packet.Presence;
 import saros.SarosPluginContext;
 import saros.intellij.ui.util.IconManager;
+import saros.intellij.ui.views.SarosMainPanelView;
 import saros.net.xmpp.JID;
 import saros.net.xmpp.roster.IRosterListener;
 import saros.net.xmpp.roster.RosterTracker;
@@ -24,6 +25,9 @@ import saros.repackaged.picocontainer.annotations.Inject;
  *
  * <p>It registers as {@link IRosterListener} and keeps the displayed contacts in sync with the
  * server.
+ *
+ * <p><b>NOTE:</b>This component and any component added here must be correctly torn down when the
+ * project the components belong to is closed. See {@link SarosMainPanelView}.
  */
 
 /*
--- a/intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
+++ b/intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
@@ -1,5 +1,7 @@
 package saros.intellij.ui.tree;
 
+import com.intellij.openapi.Disposable;
+import com.intellij.openapi.util.Disposer;
 import com.intellij.util.ui.UIUtil;
 import java.util.Collection;
 import java.util.HashMap;
@@ -29,7 +31,8 @@ import saros.repackaged.picocontainer.annotations.Inject;
  * the same issue as the Eclipse counterpart, it cannot
  * handle multiple JIDs with different resources
  */
-public class ContactTreeRootNode extends DefaultMutableTreeNode implements IRosterListener {
+public class ContactTreeRootNode extends DefaultMutableTreeNode
+    implements IRosterListener, Disposable {
 
   private static final long serialVersionUID = 1L;
 
@@ -48,6 +51,9 @@ public class ContactTreeRootNode extends DefaultMutableTreeNode implements IRost
 
   public ContactTreeRootNode(SessionAndContactsTreeView treeView) {
     super(treeView);
+
+    Disposer.register(treeView, this);
+
     SarosPluginContext.initComponent(this);
 
     this.treeView = treeView;
@@ -56,6 +62,11 @@ public class ContactTreeRootNode extends DefaultMutableTreeNode implements IRost
 
     rosterTracker.addRosterListener(this);
     rosterChanged(rosterTracker.getRoster());
+  }
+
+  @Override
+  public void dispose() {
+    rosterTracker.removeRosterListener(this);
   }
 
   /**
--- a/intellij/src/saros/intellij/ui/tree/SessionTreeRootNode.java
+++ b/intellij/src/saros/intellij/ui/tree/SessionTreeRootNode.java
@@ -1,5 +1,7 @@
 package saros.intellij.ui.tree;
 
+import com.intellij.openapi.Disposable;
+import com.intellij.openapi.util.Disposer;
 import com.intellij.util.ui.UIUtil;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -21,7 +23,7 @@ import saros.session.User;
 import saros.ui.util.ModelFormatUtils;
 
 /** Session tree root node. */
-public class SessionTreeRootNode extends DefaultMutableTreeNode {
+public class SessionTreeRootNode extends DefaultMutableTreeNode implements Disposable {
   public static final String TREE_TITLE = "Session";
   public static final String TREE_TITLE_NO_SESSIONS = "No Session Running";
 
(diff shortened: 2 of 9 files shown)


tobous:
> Codacy Here is an overview of what got changed by this pull request:
> 
> ```diff
> Issues
> ======
> - Added 1
>            
> ```
> 
> See the complete overview on [Codacy](https://app.codacy.com/app/Saros/saros/pullRequest?prid=3884551&bid=13529374)

False positive caused by imports only being used for links in javadoc in combination with line breaks in the link statements.

intellij/src/saros/intellij/ui/views/buttons/AbstractSessionToolbarButton.java
```diff
@@ -62,6 +83,21 @@ void setInitialState() {
...
+    disposeComponents();
+  }
+
+  /**
+   * Method to dispose components of the button.
+   *
+   * <p>This method is called when the project the button belongs to is disposed. It should drop any
+   * internal state that would prevent the object from being garbage collected.
+   */
+  abstract void disposeComponents();
```
srossbach: I do not understand the reason for this method. If you have additional work to do in a subclass you can just override dispose(), either calling super.dispose() at the beginning or at the end of the method.

intellij/src/saros/intellij/ui/tree/SessionTreeRootNode.java
```diff
@@ -101,6 +116,21 @@ public SessionTreeRootNode(SessionAndContactsTreeView treeView) {
...
+    sessionManager.removeSessionLifecycleListener(sessionLifecycleListener);
+
+    ISarosSession currentSession = session;
+
+    if (currentSession != null) {
+      currentSession.removeListener(sessionListener);
+      removeSessionNode(currentSession);
+    }
+
+    sessionNodeList.clear();
```
srossbach: The job of a dispose method is either to free unmanaged resources / memory and to deregister stuff properly so that the object can be garbage collected. Of course you can set all references to null / clear datastructures etc. to help the garbage collector but this is just boiler plate code with little benefits.

tobous: Removed. Just un-registering the listeners now.

intellij/src/saros/intellij/ui/views/buttons/AbstractSessionToolbarButton.java
tobous: I decided to implement this kind of as a template method to
- ensure that the listener held by the abstract super class is always correctly unregistered (which is why the method `dispose()` in the abstract class is final) and
- hopefully remind the developer to consider whether there is something to dispose when implementing a new button (i.e. avoid error of omission).

This was an active design decision on my part to err on the side of caution (as more buttons will almost definitely have to be added in the future if this is not superseded by the HTML UI beforehand).

But if you think that this is unnecessary/excessively cautious, I can use your suggested approach instead.

tobous:
Trimmed down the dispose calls to only do the minimum cleanup necessary. Adjusted the added javadoc.

tobous:
Rebased onto current master solving minor merge conflicts.

intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
```diff
@@ -22,14 +25,18 @@
  *
  * <p>It registers as {@link IRosterListener} and keeps the displayed contacts in sync with the
  * server.
+ *
+ * <p><b>NOTE:</b>This component and any component added here must be correctly torn down when the
```
srossbach: Maybe it would be more helpful to put such information to a central place (documentation etc.).

intellij/src/saros/intellij/ui/views/buttons/AbstractSessionToolbarButton.java
srossbach: protected abstract void disposeComponents()

tobous: Hmmm...*Technically*, this is currently not needed as the class is package-private and all its implementations are contained in the package. (This also means that I should make the instance variables package private as well.)

But this still creates a potential problem with the implementing classes, which are not package private. If I were to create another class outside the package that extends such an implementation (e.g. the `LeaveSessionButton`), I would no longer be able to override these methods.

But I am not sure whether this is something to worry about as it (currently) makes sense to keep all implementations in this package.

intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
tobous: Yeah, that would probably be a good idea. We currently don't have a section where IDE specific implementation details/restrictions can be listed. Will look into adding something like that.

intellij/src/saros/intellij/ui/views/buttons/AbstractSessionToolbarButton.java
tobous: But is it really worth it to change the visibility? If this is done for this method, it should also subsequently be done for all package-private methods and variables in this class and `AbstractToolbarButton`. This could be done, but I don't really see the advantage of this as all implementations should most likely be located in this package. (As mentioned in my initial comment, this is only a problem when creating a class located outside the package that is extending an implementation.)

Furthermore, this also leads to IntelliJ complaining that the variables/methods could have a more strict access restrictions.
This in itself is not really a 'problem', but it it annoying when looking at the code (and can lead to reader suspecting issues where there are none). This can be avoided by adding annotations suppressing the warnings, but this has to be done for every instance and also adds clutter.

srossbach: Why is IntelliJ complaining about such an issue ? If I create a class with multiple protected modifiers (because the class should be subclassed somewhere in the future) it makes no sense to complain about that.

Of course it is fine for now to use package private but imagine you have

saros.*.ui.*.featureXYZ

now you add saros.*.ui.*.featureXYZ.subfeatureABC which should use classes in the parent package and now you are out of luck because the standard modifiers in Java are quite bad when it comes a more complex packet structure.

I guess this will change with Java 9, see https://openjdk.java.net/projects/jigsaw/

tobous: Yeah, the access modifiers in java are somewhat lacking when it comes to inheritance or complex package structures.

> Why is IntelliJ complaining about such an issue ?

Maybe "complaining" was the wrong word. It gets marked as an issue by the code analyzer in the editor as the analyzer notices that the access modifier could be more strict *with the current usage*. This generally makes sense in my opinion to remind developers to encapsulate the class as  well es possible. However, as this is a static analysis, it of course can't deal with planned/future developments that would require a less strict access modifier.

As already mentioned, the rule can be disabled or it can be overridden through annotations, but I would still prefer to avoid it in this case as it would require a lot of annotations (5 out of the 8 fields/methods would have to be annotated).
In general, I try to resolve any issues that the code analyzer has (by following the suggestion where appropriate or adding an annotation to exclude the piece of code from the rule). This allows me to better notice when I have accidentally introduces a new issue while changing parts of the code.

intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
tobous: I did not know where to put the section. We need a better separation for IDE specific topics. I created an issue for the separation issue (#593) and for this missing documentation (#594).

tobous:

PR #551, opened 2019-06-13
Introducing EclipseReferencePointManager

pdler:
In Saros/E the EclipseReferencePointManager is used for mapping reference points to eclipse projects and also determining eclipse resources given by reference point and relative paths, or SPath.
This PR contains the implementation of the EclipseReferencePointManager and unit tests. The EclipseReferencePointManager is created by the eclipse plugin context and will be filled with eclipse projects and reference points, when a Saros Session is started, like the CoreReferencePointManager.

Features:
- Creates ReferencePoints, given by eclipse projects
- Maps ReferencePoints to eclipse projects
- Returns eclipse projects, given by ReferencePoints
- Returns eclipse resources (IResource, IFile, IFolder),
  given by ReferencePoints and ReferencePointRelativePath or SPaths

PR #583, opened 2019-07-24
Adjusting SPath references in Saros/E with EclipseReferencePointManager

pdler:
This PR contains adjustings of SPath references in Saros/E, because the SPath class can't only return the IReferencePoint and the relative path to the saros resource, but also the saros resources itself (like IFolder, IFile, ...). 
The idea is, that the EclipseReferencePointManager gets the SPath object and determinates Eclipse resources (because in Saros/E the Eclipse resource is determined from the SPath object).
The FileActivityConsumerTest had also to be adjusted.

IMPORTANT: 
Because the PR https://github.com/saros-project/saros/pull/551 has not been merged to the BasedOnReferencePoints branch yet, so this PR shows the Commit (b42539e) "[E] IntroducingReferencePointManager". So please ignore this commit.

### DIFF
--- a/core/src/saros/activities/SPath.java
+++ b/core/src/saros/activities/SPath.java
@@ -96,7 +96,10 @@ public class SPath {
    * Returns a handle for an IFolder represented by this SPath.
    *
    * @return the IFolder contained in the associated IProject for the given project relative path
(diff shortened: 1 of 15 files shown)


m273d15:
@pdler Please, fix the codacy findings before merging!

PR #585, opened 2019-07-24
Fix initial user editor state

### LINKED ISSUE #581: Initial UserEditorState is not always send when session starts
Currently, Saros/I relies on [`sendAwarenessInformation()`](https://github.com/saros-project/saros/blob/master/intellij/src/saros/intellij/editor/EditorManager.java#L251), which is triggered by the session listener method [`userFinishedProjectNegotiation(User)`](https://github.com/saros-project/saros/blob/master/intellij/src/saros/intellij/editor/EditorManager.java#L229), to transfer the initial awareness information when a user joins the session.

But, it also relies on this call to transfer the initial state of the user that joined the session. This sometimes causes issues as the call in turn relies on the editor pool, which is initialized by [`addProjectResources(IProject)`](https://github.com/saros-project/saros/blob/master/intellij/src/saros/intellij/editor/EditorManager.java#L377), triggered by [`resourcesAdded(IProject)`](https://github.com/saros-project/saros/blob/master/intellij/src/saros/intellij/editor/EditorManager.java#L239).

The issue is that the order of `resourcesAdded(IProject)` and `userFinishedProjectNegotiation(User)` is not consistent. As a result, `userFinishedProjectNegotiation(User)` will sometimes be called first, causing the initial awareness information state not to be transmitted to the other participants as it has not been initialized yet.

#### Possible solution approach
A possible solution approach would be handle the two cases separately.

When a resource is added, all clients still initialize the resources and their state. But they also send the awareness information for the added resources.

When a user joins the session, all *other* participants send the awareness information for the shared resources. The awareness information for the joining user will be handled by the first case.

This would resolve the order dependence of the two listener methods.

tobous:
#### [FIX][I] #581 Only send user editor state when other user joined

Adjusts the logic to send the current user editor state to no longer
send the state of the joined user. This was done as the session of the
joined user might not have been set up completely at this point.

Instead, the state of the joined user will be send as part of the
resources added call. This change is made in a separate commit.

#### [REFACTOR][I] Move methods to send user editor state out of listener

Moves the methods to send the user editor state out of the session
listener as they are now also needed outside the listener.

#### [FIX][I] #581 Send user editor state when resource is added to session

Adjusts the logic of EditorManager#addProjectResources(IProject) to also
send the user editor state for any editors belonging to the added
module.

Fixes #581.

#### [INTERNAL][I] Correctly set active editor in user editor state

Ensures that the correct active editor is set after initializing the
user editor state for added modules.

### DIFF
--- a/intellij/src/saros/intellij/editor/EditorManager.java
+++ b/intellij/src/saros/intellij/editor/EditorManager.java
@@ -290,56 +290,6 @@ public class EditorManager extends AbstractActivityProducer implements IEditorMa
 
           sendEditorOpenInformation(localUser, activeEditorPath);
         }
-
-        private void sendEditorOpenInformation(@NotNull User user, @Nullable SPath path) {
-          EditorActivity activateEditor =
-              new EditorActivity(user, EditorActivity.Type.ACTIVATED, path);
-
-          fireActivity(activateEditor);
-        }
-
-        /**
-         * Sends the viewport information for the given editor if it is currently visible.
-         *
-         * @param user the local user
-         * @param path the path of the editor
-         * @param editor the editor to send the viewport for
-         * @param visibleFilePaths the paths of all currently visible editors
-         */
-        private void sendViewPortInformation(
-            @NotNull User user,
-            @NotNull SPath path,
-            @NotNull Editor editor,
-            @NotNull Set<String> visibleFilePaths) {
-
-          VirtualFile fileForEditor = DocumentAPI.getVirtualFile(editor.getDocument());
-
-          if (fileForEditor == null) {
-            LOG.warn(
-                "Encountered editor without valid virtual file representation - path held in editor pool: "
-                    + path);
-
-            return;
-          }
-
-          if (!visibleFilePaths.contains(fileForEditor.getPath())) {
-            LOG.debug(
-                "Ignoring "
-                    + path
-                    + " while sending viewport awareness information as the editor is not currently visible.");
-
-            return;
-          }
-
-          LineRange localViewPort = EditorAPI.getLocalViewPortRange(editor);
-          int viewPortStartLine = localViewPort.getStartLine();
-          int viewPortLength = localViewPort.getNumberOfLines();
-
-          ViewportActivity setViewPort =
-              new ViewportActivity(user, viewPortStartLine, viewPortLength, path);
-
-          fireActivity(setViewPort);
-        }
       };
 
   /**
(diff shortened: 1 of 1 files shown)


intellij/src/saros/intellij/editor/EditorManager.java
```diff
@@ -248,9 +248,13 @@ public void resourcesAdded(final IProject project) {
          * transmitted, a second editor activated activity is send for the locally active editor to
          * correctly set the active editor in the remote user editor state for the local user.
          */
-        private void sendAwarenessInformation() {
+        private void sendAwarenessInformation(@NotNull User joinedUser) {
           User localUser = session.getLocalUser();
 
+          if (localUser.equals(joinedUser)) {
```
Drakulix: The commit message explains this reasonably well, but I would appreciate a little comment here, stating why we explicitly ignore the `joinedUser` for any uninformed readers of this code.

intellij/src/saros/intellij/editor/EditorManager.java
```diff
@@ -378,9 +383,19 @@ private void sendSelectionInformation(
...
-    Project intellijProject = project.adaptTo(IntelliJProjectImpl.class).getModule().getProject();
+    Module module = project.adaptTo(IntelliJProjectImpl.class).getModule();
+    FileIndex moduleFileIndex = ModuleRootManager.getInstance(module).getFileIndex();
+    Project intellijProject = module.getProject();
 
-    VirtualFile[] openFiles = ProjectAPI.getOpenFiles(intellijProject);
+    Set<VirtualFile> openFiles = new HashSet<>();
+
+    for (VirtualFile openFile : ProjectAPI.getOpenFiles(intellijProject)) {
+      if (moduleFileIndex.isInContent(openFile)) {
```
Drakulix: What does the `isInContent` check do? Why do we care for this property for the `openFiles`-set?
I assume this is needed to be able to send the info of the `openFileMapping`-entries, which is built from `openFiles`?

tobous: `isInContent` checks whether the passed file is part of the content of the module the file index belongs to, i.e. whether the resource belongs to the module. I added this check to filter out any open shared files that belong to a different module. Technically, this is not needed yet, but it is a nice and simple optimization to avoid sending user editor state information for editors that were already shared beforehand as this information should already be available to all participants.

intellij/src/saros/intellij/editor/EditorManager.java
```diff
@@ -248,9 +253,13 @@ public void resourcesAdded(final IProject project) {
          * transmitted, a second editor activated activity is send for the locally active editor to
          * correctly set the active editor in the remote user editor state for the local user.
          */
-        private void sendAwarenessInformation() {
+        private void sendAwarenessInformation(@NotNull User joinedUser) {
```
srossbach: I would not use `joinedUser` . It suggest that the user has joined a session which is not the case. Either use `user` or `toUser`

intellij/src/saros/intellij/editor/EditorManager.java
tobous: Added explanation to javadoc.

intellij/src/saros/intellij/editor/EditorManager.java
tobous: Done.

I confused this for a second as, currently in Saros/I, the project negotiation always coincides with the session negotiation. But this is only the case as long as you can't add resources to a running session.

tobous:

Issue #580, opened 2019-07-23
Heartbeat introduced in #397 causes vectortime desync


tobous:
**Describe the bug**
The vector time desyncs during the running session, independently of whether the users actively create activities or not. This leads to the participants no longer being able to send editor activities after some time, thereby breaking the session.

This behavior can be replicated with both Saros/I and Saros/E.

The issue starts with the commit 11f3931aed7c03ed828c9e1286308eb37e69a99c (it can't be replicated on prior commits), which makes it almost guaranteed that this behavior is caused by the heartbeat patch introduced in #397.

**To Reproduce**
1. Start a Saros session. (You can also edit some files to ensure that the session is working at this point.)
2. Wait for some time to ensure a heartbeat activity was send (at least one minute).
3. Type something in a shared editor.

**Expected behavior**
The editor activities (text edits) should be transferred to the other participants. Instead, nothing happens as the activity transformation fails with an exception.

```
ERROR 16:28:57,837 [DispatchContext] (ConcurrentDocumentServer.java:138) failed to transform jupiter activity: JupiterActivity(timestamp: [5,4], operation: Insert(1622,'a',1622), source: intellij_test_1@saros-con.imp.fu-berlin.de/Saros)
saros.concurrent.jupiter.TransformationException: Precondition #3 violated (Vector time does not match): [5,4] , [4,4]
	at saros.concurrent.jupiter.internal.Jupiter.checkPreconditions(Jupiter.java:268)
	at saros.concurrent.jupiter.internal.Jupiter.receiveJupiterActivity(Jupiter.java:158)
	at saros.concurrent.jupiter.internal.JupiterDocumentServer.transformJupiterActivity(JupiterDocumentServer.java:65)
	at saros.concurrent.management.JupiterServer.transform(JupiterServer.java:104)
	at saros.concurrent.management.ConcurrentDocumentServer.receive(ConcurrentDocumentServer.java:136)
	at saros.concurrent.management.ConcurrentDocumentServer.transformIncoming(ConcurrentDocumentServer.java:112)
	at saros.session.internal.ActivityHandler.directServerActivities(ActivityHandler.java:386)
	at saros.session.internal.ActivityHandler.handleIncomingActivities(ActivityHandler.java:136)
	at saros.session.internal.SarosSession.exec(SarosSession.java:721)
	at saros.session.internal.ActivitySequencer$3.run(ActivitySequencer.java:362)
	at saros.util.ThreadUtils$1.run(ThreadUtils.java:35)
[...]

PR #588, opened 2019-07-25
[FIX][CORE] #580 Heartbeat causes vectortime desync 

srossbach:
The Heartbeat dispatcher forgot to register the listener and so the fireActivity call was also a NoOp as well.

### DIFF
--- a/core/src/saros/concurrent/management/HeartbeatDispatcher.java
+++ b/core/src/saros/concurrent/management/HeartbeatDispatcher.java
@@ -20,7 +20,7 @@ import saros.util.NamedThreadFactory;
  */
 public class HeartbeatDispatcher extends AbstractActivityProducer implements Startable {
   private ScheduledThreadPoolExecutor heartbeatScheduledExecutor;
-  private final ISarosSession sarosSession;
+  private final ISarosSession session;
   private final UISynchronizer uiSynchronizer;
   private final JupiterClient jupiter;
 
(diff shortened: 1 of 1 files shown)


core/src/saros/concurrent/management/HeartbeatDispatcher.java
```diff
@@ -48,6 +49,7 @@ public void start() {
 
   private void dispatchHeartbeats() {
     // clientDocs should only be accessed by the main thread
+    // FIXME syncExec not asyncExec
```
m273d15: Could you add a reason? I think it is hard to verify whether an old "FIXME"s without a reason is still relevant.

core/src/saros/concurrent/management/HeartbeatDispatcher.java
```diff
@@ -56,13 +58,15 @@ private void dispatchHeartbeats() {
...
+                              new NoOperation(), session.getLocalUser(), path);
                       fireActivity(heartbeat);
                     }));
   }
 
   @Override
   public void stop() {
+    session.removeActivityProducer(this);
     heartbeatScheduledExecutor.shutdown();
+    // FIXME wait here until everything is finished
```
m273d15: Could you add a reason?

srossbach: I thought the reason is obvious. After calling stop I would expect that the component is stopped. If you read the Javadoc and know the the behavior of asyncExec in combination with shutdown it is clear that this component can still do something (not the case anymore after the listener removal). However it is possible that a heartbeat is still broadcasted / executed in the ActivityHandler when you leave the stop method.

srossbach: I will address the FIXME in another patch. (Pushing it later today).

PR #589, opened 2019-07-25
[INTERNAL][E] Reworks ResourceChangeValidator

srossbach:
As the validator is working again in Eclipse 4.X the code is cleaned up
and split among multiple classes.

This patch also introduces logic to distinguish between deleting a
project or just moving / renaming it.


eclipse/src/saros/session/resources/validation/messages.properties
```diff
@@ -0,0 +1,3 @@
+ResourceChangeValidator_0=Participants with read-only access should not edit the resources of this shared project. If you wish to makes changes, ask the host to give you write-access.
+ResourceChangeValidator_1=You should leave the Saros session before deleting a shared project. Deleting a project has no effect on the other session participants, so the session will become inconsistent.
+ResourceChangeValidator_2=You should leave the Saros session before renaming or moving a shared project. Renaming or moving a project has no effect on the other session participants, so the session will become inconsistent.
```
srossbach: Have not tried out but I guess this is not true. Looking at the delta visitor moving the project also moves the files and files moves counts as delete if they are moved to a different  project. Same applies when deleting a project.

srossbach: Ok it seems to some logic issues that is not the case, however the log statement is more than confusing.

DEBUG 22:37:23,627 [ModalContext] (SharedResourcesManager.java:235) ignoring delta changes for project P/B as it was only opened

The project B was not opened, it was closed and does not exists anymore ...

eclipse/src/saros/session/resources/validation/messages.properties
```diff
@@ -0,0 +1,3 @@
+ResourceChangeValidator_0=Participants with read-only access should not edit the resources of this shared project. If you wish to makes changes, ask the host to give you write-access.
```
m273d15: You substituted the additional context (in the message names) by numbers. Don't you think it is more readable with extra context?

eclipse/src/saros/session/resources/validation/messages.properties
m273d15: > Have not tried out but I guess this is not true. Looking at the delta visitor moving the project also moves the files and files moves counts as delete if they are moved to a different project. Same applies when deleting a project.

Do you mean case 2 will never apply? I am not really sure what you are trying to test.

srossbach: All cases apply. My issue was if I delete project A, what will happen on the remote side, e.g will only inconsistencies occur or will the whole content (not the project itself) be deleted as well.

eclipse/src/saros/session/resources/validation/ResourceDeltaVisitor.java
```diff
@@ -0,0 +1,62 @@
...
+
+        return false;
+      }
+    }
+
+    isModifyingResources = true;
+    return false;
+  }
+
+  /** @return the isModifyingResources */
```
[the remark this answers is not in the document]
srossbach: Was autogenerated. Removed it.

eclipse/src/saros/session/resources/validation/messages.properties
srossbach: The was done by the Eclipse Tooling after I got issues with the existing Message.class. Will add more meaningful names.

eclipse/src/saros/session/resources/validation/messages.properties
m273d15: We should test the case, but if I remember correctly we already tested the case in our last release testing and had the expected behavior of an inconsistent state without deleting the resources on host side.

m273d15:
Travis still fails, because the formatting does not comply with the google-java-format.
Do you configured the [google-java-formatter](https://www.saros-project.org/contribute/development-environment.html#configure).

srossbach:
Yes I did but somehow the SaveAction did not trigger.

eclipse/src/saros/session/resources/validation/ResourceDeltaVisitor.java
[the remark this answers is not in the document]
srossbach: Done

eclipse/src/saros/session/resources/validation/messages.properties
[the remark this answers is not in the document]
srossbach: Done

srossbach: Done

eclipse/src/saros/session/resources/validation/ResourceChangeValidator.java
```diff
@@ -0,0 +1,97 @@
...
+ * realized this
+ */
+public class ResourceChangeValidator extends ModelProvider {
+  private static final Logger log = Logger.getLogger(ResourceChangeValidator.class);
+
+  /** Error code for internal use, but we don't need it. */
+  private static final int ERROR_CODE = 0;
+
+  private static final IStatus MODIFYING_RESOURCES_ERROR_STATUS =
+      new Status(
```
tobous: You could move this initialization to a static helper method to avoid code duplication as only the message ever changes.

eclipse/src/saros/session/resources/validation/ResourceChangeValidator.java
```diff
@@ -0,0 +1,97 @@
...
+          IStatus.ERROR,
+          Saros.PLUGIN_ID,
+          ERROR_CODE,
+          Messages.ResourceChangeValidator_MoveOrRenameProjectErrorMessage,
+          null);
+
+  private static volatile ISarosSession session;
+
+  /* The provider is created on the fly when needed so we cannot benefit from the initialize method and so this has to be static. */
```
tobous: Wouldn't it then be easier to hold a static reference to the session manager instead and just request the session object when needed? It would require a static block injecting the session manager, but would basically be the same logic and would avoid the usage of the support class (`ResourceChangeValidatorSupport`) and the volatile usage of the session (not using the static reference directly).

srossbach: This was the former approach. I changed it because now you have two options to disable this class. Either via the plugin.xml or via the context factory.

TBH I do not like this crappy plugin.xml file.

eclipse/src/saros/session/resources/validation/ResourceChangeValidator.java
srossbach: Done

eclipse/src/saros/session/resources/validation/ResourceChangeValidator.java
tobous: I am not sure I understand what you mean. Could you elaborate? Do you mean that you can also disable this by simply not setting the session (e.g. by not starting the support class)? What advantages does this provide?

srossbach:
Only added Tobias suggestion.

PR #590, opened 2019-07-26
Upgrade to IntelliJ 2019.2

tobous:
#### [FIX][I] Replace usage of DataKeys with LangDataKeys

Replaces the usage of DataKeys with LangDataKeys. This is necessary as
IntelliJ 2019.2 deprecates the class and largely removes its contents.

#### [BUILD][I] Update to IntelliJ 2019.2

Updates the version to build the plugin against to 2019.2.

Subsequently updates the gradle-intellij-plugin to the latest version.

The switch to 2019.2 causes an illegal reflexive access warning to be
displayed when running the plugin locally using 'runIDE'. This seems to
be an issue with IntelliJ and does not seem to influence our plugin.

### DIFF
--- a/.idea/gradle.xml
+++ b/.idea/gradle.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="GradleMigrationSettings" migrationVersion="1" />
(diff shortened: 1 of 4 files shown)


tobous:

PR #592, opened 2019-07-26
[INTERNAL][CORE] Addresses FIXMEs in HeatbeatDispatcher

srossbach:
This patch addresses the FIXMEs. In addition it adds a detailed
description from the original Jupiter paper why it is necessary to
generate NoOps.

Refactored the dispatchHeartbeats method to make it more readable.

### DIFF
--- a/core/src/saros/concurrent/management/HeartbeatDispatcher.java
+++ b/core/src/saros/concurrent/management/HeartbeatDispatcher.java
@@ -1,72 +1,106 @@
 package saros.concurrent.management;
 
+import java.util.Map.Entry;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import org.apache.log4j.Logger;
 import saros.activities.JupiterActivity;
+import saros.activities.SPath;
+import saros.concurrent.jupiter.Algorithm;
+import saros.concurrent.jupiter.internal.Jupiter;
 import saros.concurrent.jupiter.internal.text.NoOperation;
 import saros.repackaged.picocontainer.Startable;
 import saros.session.AbstractActivityProducer;
 import saros.session.ISarosSession;
+import saros.session.User;
 import saros.synchronize.UISynchronizer;
 import saros.util.NamedThreadFactory;
 
 /**
- * This class sends periodically JupiterOperations to ack other clients Activities.
+ * This class generates and sends periodically Jupiter-NoOperations to acknowledge operations.
  *
- * <p>This fixes a logical memory leak, that is present in the Jupiter-implementation, if a client
- * does not participate in the session actively (usually Saros/S). This causes other clients to hold
- * JupiterActivities indefinitely to be able to merge them at all times. Sending 'NoOperation's
- * fixes this issue by acking received activities though the operation's vectorTime.
+ * <p>See
+ * https://www.researchgate.net/publication/220876978_High-Latency_Low-Bandwidth_Windowing_in_the_Jupiter_Collaboration_System
+ *
+ * <p>Some fine points: In our system, messages must be saved until they are acknowledged by the
+ * other party, since they may be needed in order to fix up incoming messages. Normally, these
+ * acknowledgments are piggy-backed on traffic going the other way. However, it is possible for the
+ * traffic to a window to be one-sided (e.g., for a status display window being periodically
+ * updated). Therefore, each side must periodically generate explicit acknowledgments (i.e. no-op
+ * messages) to prevent the outgoing queues from growing forever.
  */
 public class HeartbeatDispatcher extends AbstractActivityProducer implements Startable {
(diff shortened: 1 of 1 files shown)


core/src/saros/concurrent/management/HeartbeatDispatcher.java
```diff
@@ -1,72 +1,106 @@
...
-  private ScheduledThreadPoolExecutor heartbeatScheduledExecutor;
+
+  private static final Logger log = Logger.getLogger(HeartbeatDispatcher.class);
+
   private final ISarosSession session;
   private final UISynchronizer uiSynchronizer;
-  private final JupiterClient jupiter;
+  private final JupiterClient jupiterClient;
+
+  private ScheduledThreadPoolExecutor heartbeatScheduledExecutor;
```
Drakulix: Renaming and moving variables and similar changes do make this a lot harder to review then necessary because functional differences are harder to spot.

core/src/saros/concurrent/management/HeartbeatDispatcher.java
```diff
@@ -1,72 +1,106 @@
...
-                          jupiter.generateJupiterActivity(
-                              new NoOperation(), session.getLocalUser(), path);
-                      fireActivity(heartbeat);
-                    }));
+
+    assert uiSynchronizer.isUIThread() : "invalid thread access";
+
+    final User localUser = session.getLocalUser();
+
+    for (Entry<SPath, Jupiter> entry : jupiterClient.getClientDocs().entrySet()) {
```
Drakulix: Why expand the `forEach` call into a verbose for loop?
This is much harder to read and I would strongly vote for the brevity of the first approach.

I cannot spot any functional differences here, except for moving the `uiSynchronizer`, so why even bother?

tobous: In general, I would agree with this. Especially when iterating over maps, lambdas are preferable in my mind as they allow you to directly access both key and value without the usage of a map entry.

In this case, the following would be the shortest version:
```java
    jupiterClients
        .getClientDocs()
        .forEach(
            (path, jupiterAlgorithm) ->
                fireActivity(
                    jupiterAlgorithm.generateJupiterActivity(
                        new NoOperation(), sarosSession.getLocalUser(), path)));
```
<details><summary>(In this case, explicitly declaring the activity first doesn't really improve the formatting.)</summary>

```java
    jupiterClients
        .getClientDocs()
        .forEach(
            (path, jupiterAlgorithm) -> {
              JupiterActivity heartbeat =
                  jupiterAlgorithm.generateJupiterActivity(
                      new NoOperation(), sarosSession.getLocalUser(), path);

              fireActivity(heartbeat);
            });
```

</details>

In this case, the indentation is not really optimal, so the readability is down to personal preference in my mind (I think I would use the lambda) . I am fine with both options (lambda or loop).

core/src/saros/concurrent/management/HeartbeatDispatcher.java
```diff
@@ -1,72 +1,106 @@
...
+  private final JupiterClient jupiterClient;
+
+  private ScheduledThreadPoolExecutor heartbeatScheduledExecutor;
 
   public HeartbeatDispatcher(
-      ISarosSession sarosSession,
-      UISynchronizer uiSynchronizer,
-      ConcurrentDocumentClient cdClient) {
-    this.session = sarosSession;
+      final ISarosSession session,
```
tobous: Is there a reason to declare these final or is it just a convention?

core/src/saros/concurrent/management/HeartbeatDispatcher.java
tobous: +1 In general, it would be nice to split off refactorings into separate commits.

core/src/saros/concurrent/management/HeartbeatDispatcher.java
srossbach: Well the Google Formatter can format things in not such a nice way and I do not like it when the code starts to "drawing arrows".

On the other hand I really do not like it if Lambas (which should represent the art of functional programming) calling non idempotent methods.

Beside that the following call would be more of a functional style:

    this.jupiter
        .getClientDocs()
        .entrySet()
        .stream()
        .map(
            (entry) ->
                entry
                    .getValue()
                    .generateJupiterActivity(
                        new NoOperation(), session.getLocalUser(), entry.getKey()))
        .forEach((activity) -> fireActivity(activity));

srossbach:
As for the syncExec and asyncExec difference.

asyncExec is simply fire and forget. Image doing executing a task that takes 2 seconds to complete and under heavy load 10 seconds. You schedule the execution every 5 seconds and so flooding the queue of the system if it is under heavy load.

The real reason I have to use syncExec is to ensure that the thread / scheduler is done after calling await.

Issue #66, opened 2014-12-15
Reusing existing project requires cumbersome manual selection


dpp-gerrit:
Upon accepting a session invitation, the user has the choice to create a new project (default) or reuse an existing one (for each incoming project). 

If you selected the reuse option in Saros 13.12.6 and a project with the same name as the incoming project already existed locally, that project was selected automatically. In 14.10.31.TESTING, the user has to click on the Browse... button and select the project to be reused. This is especially cumbersome if you share more than one project at once, and the recipient wants to reuse existing projects for all of the incoming ones. (In the User Acceptance Test, we shared the whole Saros repository with 9 projects -- selecting the local counterparts was no fun.)

If I remember correctly, this behavior was changed when the Partial Sharing feature was improved so that newly shared resources of an already partially shared project can no longer be added into a different project than the one they "belong" to. That change was totally fine, but forcing the user to select the project to reuse manually is not.

Reported by: fzieris

Original Ticket: [dpp/bugs/850](https://sourceforge.net/p/dpp/bugs/850)

PR #595, opened 2019-07-26
[FIX][E] #66 Reuse existing project mappings 

srossbach:
This patch introduces a feature that will remember the settings made
when selecting the destination for incoming projects.

The settings will then be applied during a new session and so it is not
necessary to enter the names of the local projects over and over again.

### DIFF
--- a/eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
+++ b/eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
@@ -7,10 +7,13 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.regex.Pattern;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.log4j.Logger;
 import org.eclipse.core.resources.IFile;
@@ -31,6 +34,7 @@ import org.eclipse.jface.dialogs.ErrorDialog;
 import org.eclipse.jface.dialogs.IDialogConstants;
 import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.operation.IRunnableWithProgress;
+import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.jface.wizard.Wizard;
 import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
@@ -79,11 +83,15 @@ public class AddProjectToSessionWizard extends Wizard {
 
   private boolean isExceptionCancel;
 
+  private final ResourceMappingStorage mappingStorage;
+
   @Inject private IChecksumCache checksumCache;
 
   @Inject private IConnectionManager connectionManager;
 
   @Inject private Preferences preferences;
+
+  @Inject private IPreferenceStore preferenceStore;
 
   @Inject private ISarosSessionManager sessionManager;
 
@@ -124,6 +132,8 @@ public class AddProjectToSessionWizard extends Wizard {
 
     SarosPluginContext.initComponent(this);
 
+    this.mappingStorage = new ResourceMappingStorage(preferenceStore);
+
     this.negotiation = negotiation;
     this.peer = negotiation.getPeer();
 
@@ -152,9 +162,16 @@ public class AddProjectToSessionWizard extends Wizard {
 
     if (session == null) return;
 
+    final Map<String, String> lastProjectNameMapping = mappingStorage.getMapping(peer);
+
     namePage =
         new EnterProjectNamePage(
-            session, connectionManager, preferences, peer, negotiation.getProjectNegotiationData());
+            session,
+            connectionManager,
+            preferences,
+            peer,
+            negotiation.getProjectNegotiationData(),
+            lastProjectNameMapping);
 
     addPage(namePage);
   }
@@ -215,6 +232,8 @@ public class AddProjectToSessionWizard extends Wizard {
     }
 
     if (!confirmOverwritingResources(modifiedResources)) return false;
+
+    storeCurrentProjectNameMapping(peer);
 
     triggerProjectNegotiation(targetProjectMapping, openEditors);
 
@@ -666,6 +685,110 @@ public class AddProjectToSessionWizard extends Wizard {
 
     if (proceed) {
       for (IEditorPart editor : dirtyEditors) editor.doSave(new NullProgressMonitor());
+    }
+  }
+
+  private void storeCurrentProjectNameMapping(final JID jid) {
+    Map<String, String> currentProjectNameMapping = new HashMap<>();
+
+    for (final ProjectNegotiationData data : negotiation.getProjectNegotiationData()) {
+      final String projectID = data.getProjectID();
+      currentProjectNameMapping.put(
+          data.getProjectName(), namePage.getTargetProjectName(projectID));
+    }
+
+    mappingStorage.updateMapping(jid, currentProjectNameMapping);
+  }
+
+  private static final class ResourceMappingStorage {
+    private static final String RESOURCE_MAPPING_STORE_PREFIX = "resource.mapping.storage.";
+
+    private static final CharSequence DELIMITER = ":";
+
+    private static final CharSequence ESCAPE_CHARACTER = "\\";
+
+    private static final CharSequence ESCAPE_SEQUENCE =
+        ESCAPE_CHARACTER.toString() + DELIMITER.toString();
+
+    private static final Pattern SPLIT_PATTERN =
+        Pattern.compile(
+            "(?<!"
+                + Pattern.quote(ESCAPE_CHARACTER.toString())
+                + ")"
+                + Pattern.quote(DELIMITER.toString()));
+
+    private final IPreferenceStore store;
+
+    public ResourceMappingStorage(final IPreferenceStore store) {
+      this.store = store;
+    }
+
+    public Map<String, String> getMapping(final JID jid) {
+      final Map<String, String> result = new HashMap<>();
+
+      final String mapping = store.getString(RESOURCE_MAPPING_STORE_PREFIX + jid.getBase());
+
+      if (mapping.isEmpty()) return result;
+
+      final String[] names = SPLIT_PATTERN.split(mapping);
+
+      for (int i = 0; i < names.length - 1; i += 2) {
+        if (names[i].isEmpty() || names[i + 1].isEmpty()) continue;
+
+        result.put(unescape(names[i]), unescape(names[i + 1]));
+      }
+
+      return result;
+    }
+
+    public void updateMapping(final JID jid, final Map<String, String> resourceNameMapping) {
+
+      final Map<String, String> lastResourceNameMapping = getMapping(jid);
+
+      final Set<String> resourceNameDestinations = new HashSet<>(resourceNameMapping.values());
+
+      /*
+       * Remove existing mappings to avoid following problem:
+       * 1. We add A->A, B->B, C->C
+       * 2. Afterwards  A->A, B->C
+       * 3. When we now lookup A,B,C we will get A->A, B->C, C->C which is an invalid mapping
+       *
+       * In other words, ensure that the mapping is injective.
+       */
+
+      final Iterator<Entry<String, String>> it = lastResourceNameMapping.entrySet().iterator();
+
+      while (it.hasNext()) {
+
+        final Entry<String, String> entry = it.next();
+
+        if (resourceNameDestinations.contains(entry.getValue())) it.remove();
+      }
+
+      lastResourceNameMapping.putAll(resourceNameMapping);
+
+      if (lastResourceNameMapping.isEmpty()) return;
+
+      final StringBuilder builder = new StringBuilder();
+
+      for (final Entry<String, String> entry : lastResourceNameMapping.entrySet()) {
+        builder.append(escape(entry.getKey()));
+        builder.append(DELIMITER);
+        builder.append(escape(entry.getValue()));
+        builder.append(DELIMITER);
+      }
+
+      builder.setLength(builder.length() - DELIMITER.length());
+
+      store.setValue(RESOURCE_MAPPING_STORE_PREFIX + jid.getBase(), builder.toString());
+    }
+
+    private static String escape(final String value) {
+      return value.replace(DELIMITER, ESCAPE_SEQUENCE);
+    }
+
+    private static String unescape(final String value) {
+      return value.replace(ESCAPE_SEQUENCE, DELIMITER);
     }
   }
 }
--- a/eclipse/src/saros/ui/wizards/pages/EnterProjectNamePage.java
+++ b/eclipse/src/saros/ui/wizards/pages/EnterProjectNamePage.java
@@ -5,6 +5,7 @@ import java.nio.charset.Charset;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -12,7 +13,6 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.log4j.Logger;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.ResourcesPlugin;
@@ -23,8 +23,8 @@ import org.eclipse.jface.resource.ImageDescriptor;
 import org.eclipse.jface.wizard.WizardPage;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.Image;
-import org.eclipse.swt.layout.GridData;
-import org.eclipse.swt.layout.GridLayout;
+import org.eclipse.swt.graphics.Point;
+import org.eclipse.swt.layout.FillLayout;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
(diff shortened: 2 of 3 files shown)


eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -666,4 +680,79 @@ private void displaySaveDirtyEditorsDialog(final List<IEditorPart> dirtyEditors)
...
+    setNewProjectNameMapping(jid, currentProjectNameMapping);
+  }
+
+  private Map<String, String> getLastProjectNameMapping(final JID jid) {
+    Map<String, String> result = new HashMap<>();
+
+    final String mapping = preferenceStore.getString(projectMappingStorePrefix + jid.getBase());
+
+    if (mapping == null || mapping.isEmpty()) return result;
+
```
m273d15: the following code is very complex. Why not:
```java
// split only at unescaped colons
String[] splitted = res.split("(?<!\\\\):");
for (int i = 0; i < splitted.length; i += 2 )
{
  String key = splitted[i].replace("\\:", ":");
  String value = splitted[i + 1].replace("\\:", ":");
  result.put(key, value);  
}
```

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -666,4 +680,79 @@ private void displaySaveDirtyEditorsDialog(final List<IEditorPart> dirtyEditors)
...
+    return result;
+  }
+
+  private void setNewProjectNameMapping(
+      final JID jid, final Map<String, String> projectNameMapping) {
+
+    final Map<String, String> lastProjectNameMapping = getLastProjectNameMapping(jid);
+
+    for (final Entry<String, String> entry : projectNameMapping.entrySet())
```
m273d15: Why not `lastProjectNameMapping.putAll(projectNameMapping)` (without loop)

eclipse/src/saros/ui/wizards/pages/EnterProjectNamePage.java
```diff
@@ -371,11 +382,27 @@ private void preselectProjectNames() {
 
       if (project != null) continue;
 
-      String projectNameProposal =
-          findProjectNameProposal(
-              remoteProjectMapping.get(projectID), reservedProjectNames.toArray(new String[0]));
+      final String remoteProjectName = remoteProjectIdToNameMapping.get(projectID);
```
m273d15: Why not separating the into two private methods (corresponding to the two steps in the javadoc) in order to improve the readability?

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
srossbach: Thanks ! It the split really Java Syntax ? Isn't it (?!<[^\\\\]):

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
srossbach: Because it was late when I wrote this  ....

eclipse/src/saros/ui/wizards/pages/EnterProjectNamePage.java
srossbach: I guess that might require a little bit of more work. I will look into it if it is possible to split this without adding more checks and corrections.

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
m273d15: Yes this is Java Syntax. I tested the code with a simple example (executed with jdk8):
```java
import java.util.List;
import java.util.Arrays;
import java.util.Map;
import java.util.HashMap;

public class Test {
  public static void main(String[] args) {
    String res = "\\:remote\\:1:\\:loca\\:l1:remote2:local2\\:";
    Map<String, String> result = new HashMap<>();

    String[] splitted = res.split("(?<!\\\\):");
    for (int i = 0; i < splitted.length; i+=2 ) {
      String key = splitted[i].replace("\\:", ":");
      String value = splitted[i + 1].replace("\\:", ":");
      result.put(key, value);  
    }

    System.out.println(result.get(":remote:1"));
    System.out.println(result.get("remote2"));
  }
}

```

eclipse/src/saros/ui/wizards/pages/EnterProjectNamePage.java
srossbach: So after thinking about it for a while. Does it really make sense to split this ? I know that methods should not be that long or complex but the problem I have all the time is how to split methods ... The problem is in general the mutating state of such methods. If I split method A into multiple sub methods I have to ensure that each method called in another context does always produce the same result. In praxis (as I have seen it) you are ending up with a class consisting of so much methods that over time nobody has a clue what each method really does without reading the code.

The problem on this method is the second part. If I split this I am losing the project names that are already assigned through the first part. To get around of this issues I would have to query the enabled state which can also be mutated by any other method or logic. 

This class is going to be deleted / replaced anyway. I am still waiting to see the HTML equivalent of this class. Not one student have every tried to implement this wizard yet in the HTML GUI.

srossbach:
@m273d15 does the logic introduced by this patch is actually solving #66 in a good manner ?

m273d15:
I think #66 should be fixed, but I still think about the location of the logic. In principle the logic of remembering a previously used project name is not IDE specific. What do you think about moving it out of the ui?

srossbach:
I would put at least the storing and retrieving logic to the shared ui package, However I do not find a good package name and currently the whole ui package only contains brower stuff beside the util package.

m273d15:
I am not sure whether the ui package is intended to contain "real logic". As far as I know/understand `ui` currently contains only classes that allow the connection of the (HTML-)GUI with the core functionality.
Why is storing/retrieving logic something ui related? Why not moving it into core?

I think this is an important question and will also come up in the refactorings of @stefaus  who also has to separate business logic and ui code.

srossbach:
It is a storage "logic" that is only used by the UI. Do you think that this will benefit the server project ? If so I could put into the core.

As for the refactoring part. I do not know if there is much left beside a few classes that really does not contain any UI calls.

srossbach:
The problem with guessing the project mapping is:

1. If you share multiple projects you really can get lost because on how the current UI is written. You have to go through all tabs and afterwards you may get the nice warning dialog which can be quite big regarding the changes. Pre-selecting projects based on the name can cause unwanted side effects if not noticed.

2. Second scenario. For what reason even if do not want project A map to project A but to project B. With the former approach it will be mapped to A again every time for a new session.

I guess the real fix would be completely redesign this Wizard / Dialog but currently no one came up with a good solution yet.

srossbach:
Maybe it is helpful to manipulate the tabs in some way (adding an icon) to indicate that a project is preselected ?

tobous:
> Maybe it is helpful to manipulate the tabs in some way (adding an icon) to indicate that a project is preselected ?

Yeah, that would probably be a good idea. Because the argument that users could be surprised by the pre-selection also applies to using the held mapping. Even if the user wanted to share a specific mapping today, they might not want to share it tomorrow.

But I think this is a problem that we won't find a perfect solution for.

As already stated, I am not against the solution presented here. If you think it provides useful functionality, that's good enough for me. But I still think that using a guessed mapping if no data is present is also a good idea. This can still be used in combination with this approach (i.e. guess if we don't have any mappings from previous sessions).

But, I would agree with Kelvin that, if you were to add this functionality, it would be nice to move it to the Saros core package as it could also prove useful to other Saros implementations.
But I am not sure how much more work it would be to make the approach more generic (e.g. for Saros/I, a mapping of just the module names would not be sufficient; for the local representation of the shared resource, you would need more information as module names are not unique when working with multiple open projects). Just talking about it makes me question whether this is really worth the effort :sweat_smile:.

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -666,4 +680,79 @@ private void displaySaveDirtyEditorsDialog(final List<IEditorPart> dirtyEditors)
...
+      lastProjectNameMapping.put(entry.getKey(), entry.getValue());
```
tobous: Just adding the new mapping to the existing mapping could lead to some weird states.
If I were to
- accept a project now and set up the mapping A->A, B->B, and C->C
- afterwards, accept a new session (with the same host) and set up the mapping B->C
- afterwards, accept a new session with the initial three projects

This would lead to the automatic setup suggesting to map two resources to C (A->A, B->C, C->C).

In general, as this call potentially overwrites existing mappings, it could lead to mixing the mappings of different sessions. A solution would be to hold mappings for project combinations and always trying to apply the largest held mappings (you could still also try to combine non-overlapping mappings). But this would make the logic more complicated and would increase the saved data size.

As already stated, I find it hard to construct a behavior with this logic that covers the most important use-cases without leading to too much confusion for the users in the other (corner) cases.

eclipse/src/saros/ui/wizards/pages/EnterProjectNamePage.java
```diff
@@ -46,45 +47,57 @@
...
       JID peer,
-      List<ProjectNegotiationData> projectNegotiationData) {
+      List<ProjectNegotiationData> projectNegotiationData,
+      Map<String, String> desiredRemoteToLocalProjectNameMapping) {
 
     super(Messages.EnterProjectNamePage_title);
     this.session = session;
     this.connectionManager = connectionManager;
     this.preferences = preferences;
+    this.desiredRemoteToLocalProjectNameMapping =
```
tobous: Not strictly necessary as, currently, the passed mapping can't be `null` (`getLastProjectNameMapping(peer)` returns an empty hash map if no mapping is found).

But you can still keep it to protect against future 'misuse' if you would prefer. But I would move this to the end of the initialization block (behind `this.peer = peer;`) or avoid the inline-nullcheck to make this easier to read.

eclipse/src/saros/ui/wizards/pages/EnterProjectNamePage.java
```diff
@@ -371,11 +382,27 @@ private void preselectProjectNames() {
...
+
+      final String desiredLocalProjectName =
+          desiredRemoteToLocalProjectNameMapping.get(remoteProjectName);
+
+      boolean existingProject = false;
+
+      String projectNameProposal = null;
+
+      if (desiredLocalProjectName != null
+          && !reservedProjectNames.contains(desiredLocalProjectName)) {
```
tobous: I am a bit confused by the usage of `projectOptionComposites`. `reservedProjectNames` are filled using `projectOptionComposites`, but you also use it to iterate over the projects to suggest proposals for.

What does the map `projectOptionComposites` contain? The mapping of already shared projects? I.e. are the names 'reserved' as there already is a mapping for the project in the current session?

srossbach: https://help.eclipse.org/kepler/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fswt%2Fwidgets%2FComposite.html


as Swing https://docs.oracle.com/javase/7/docs/api/javax/swing/JComponent.html

Actually this is the content of a tab.

I know this class is a mess but it were much more worse years ago.

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
srossbach: I see three options here.
1. Store set of sets (as you already mentioned an overkill)
2. Refuse to update the specific entry if it already exists
3. Replace the mapping and remove affected stuff, i.e if I map B->C then I would remove the C->C mapping.

If would prefer option 3.

srossbach:
The question is if this is really "core" aka business logic or just simple UI logic. The first one goes into the core, the other one should go (IMO) into the shared ui bundle (which currently only helds HTML stuff)

eclipse/src/saros/ui/wizards/pages/EnterProjectNamePage.java
tobous: Ok, I did not see that the two loops run on different continue conditions. Thought that both were checking `project==null`.

How about adding a short comment before each loop stating the purpose as this is really hard to understand otherwise? This would be a compromise to moving them to different method (even though I also prefer to split this into different methods, but I get your argument against it; as a side note: data re-creation could be avoided by returning the reserved names from the first method and passing it to the second).

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
tobous: Option three would probably be the best compromise. 

But it could still lead to unexpected situations. When you replace the second step in the example with mapping B->D, the mapping in step three would be legal but still not what the user expects (B->D instead of B->B).
Either way, there won't be a perfect solution, so we have to somehow get the user to check the tabs again. But this also somewhat reduces the usefulness of the feature if the user still has to check everything (at least when there were other sessions in-between using different mappings).

How about holding whether a mapping was overwritten and only then notifying the user/marking these tabs? This could minimize the needed interaction by the user while reducing the probability of unexpected outcomes.

eclipse/src/saros/ui/wizards/pages/EnterProjectNamePage.java
srossbach: Moved to the end to improve readability.

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -668,4 +687,105 @@ private void displaySaveDirtyEditorsDialog(final List<IEditorPart> dirtyEditors)
...
+
+    public void updateMapping(final JID jid, final Map<String, String> resourceNameMapping) {
+
+      final Map<String, String> lastResourceNameMapping = getMapping(jid);
+
+      final Set<String> resourceNameDestinations = new HashSet<>(resourceNameMapping.values());
+
+      /* Remove existing mapping to avoid following problem.
+       * 1. We add A->A, B->B, C->C
+       * 2. Afterwards  A->A, B->B
```
srossbach: almost ...

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -668,4 +687,106 @@ private void displaySaveDirtyEditorsDialog(final List<IEditorPart> dirtyEditors)
...
+
+    private static final CharSequence DELIMITER = ":";
+
+    private static final CharSequence ESCAPE_CHARACTER = "\\";
+
+    private static final CharSequence ESCAPE_SEQUENCE =
+        ESCAPE_CHARACTER.toString() + DELIMITER.toString();
+
+    private static final Pattern SPLIT_PATTERN =
+        Pattern.compile(
```
tobous: I am not very well versed when it comes to regular expressions. Did you test that it actually does what it is supposed to?

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -668,4 +687,106 @@ private void displaySaveDirtyEditorsDialog(final List<IEditorPart> dirtyEditors)
...
+    }
+
+    public Map<String, String> getMapping(final JID jid) {
+      final Map<String, String> result = new HashMap<>();
+
+      final String mapping = store.getString(RESOURCE_MAPPING_STORE_PREFIX + jid.getBase());
+
+      if (mapping.isEmpty()) return result;
+
+      final String[] names = SPLIT_PATTERN.split(mapping);
```
tobous: Check whether array length is even to avoid index out of bound exception in case of data corruption? But I am not sure whether this is really necessary.

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -668,4 +687,106 @@ private void displaySaveDirtyEditorsDialog(final List<IEditorPart> dirtyEditors)
...
+
+      for (int i = 0; i < names.length; i += 2) {
+        if (names[i].isEmpty() || names[i + 1].isEmpty()) continue;
```
tobous: Can this happen (i.e. is an empty mapping possible), or is this just a safeguard against data corruption or something similar?

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -668,4 +687,106 @@ private void displaySaveDirtyEditorsDialog(final List<IEditorPart> dirtyEditors)
...
+
+      /*
+       * Remove existing mappings to avoid following problem:
+       * 1. We add A->A, B->B, C->C
+       * 2. Afterwards  A->A, B->C
+       * 3. When we now lookup A,B,C we will get A->A, B->C, C->C which is an invalid mapping
+       *
+       * In other words, ensure that the mapping is injective.
+       */
+      for (final Iterator<Entry<String, String>> it = lastResourceNameMapping.entrySet().iterator();
```
tobous: I would prefer using a while loop defining the iterator beforehand as it is easier to read / not as unexpected.
My assumption is that you used a for loop to help the garbage collector (by binding the variable to the loop scope). But this is a bit confusing to read.

eclipse/src/saros/ui/wizards/pages/EnterProjectNamePage.java
```diff
@@ -110,35 +121,34 @@ public void performHelp() {
...
   public void createControl(Composite parent) {
-    GridData gridData;
 
-    Composite composite = new Composite(parent, SWT.NONE);
-    setControl(composite);
+    Composite composite =
+        new Composite(parent, SWT.NONE) {
+          // dirty hack - if someone knows how to do it right with layout please change this
+          @Override
+          public Point computeSize(int wHint, int hHint, boolean changed) {
```
tobous: Was this change/workaround included in this patch for convenience's sake or to counteract unwanted layout changes caused by the rest of this patch?

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
srossbach: Should be 

for (int i = 0; i < names.length -1; i +=2)

instead of 

for (int i = 0; i < names.length; i +=2)

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
srossbach: Just a safeguard. It not that hard to manipulate the storage by hand.

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
srossbach: Just a personal choice. I do not think is any great benefit either using for or while. I do not even know when the stack is corrected, but I guess "it" is popped of after leaving the for loop.

eclipse/src/saros/ui/wizards/pages/EnterProjectNamePage.java
srossbach: It addresses the FIXME. TBH this class will be deleted somewhere in the near future. I do not see any reason to make small commits here and this change is described in the commit message. Of course it could be split.

eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
srossbach: Well it actually worked even without escaping at all. I have to test this in code as I cannot create directories containing ":" under Windows.

srossbach: Before store: {Hello=World, 1:2:3:4:5:6:7:8:9:0=::::::::::::::::::::::::::::::, Foo:Bar=Foo:Foo}
Stored string: Foo\:Bar:Foo\:Foo:Hello:World:1\:2\:3\:4\:5\:6\:7\:8\:9\:0:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:\:
Read from store: {Hello=World, 1:2:3:4:5:6:7:8:9:0=::::::::::::::::::::::::::::::, Foo:Bar=Foo:Foo}
Write == Read ? true

Issue #32, opened 2012-07-15
Contribution annotation does not disappear


dpp-gerrit:
What happened:

- Session with Julia, Franz, Stephan
- Julia opens a file (MUCManager.java from Saros sourcecode), selects and copies whole content and pastes it multipe times
- On Franz' side: Whole content is colored in Julia's color (blue)
- On Franz' side: Own changes don't have any color at all.
- Stephan joins, and makes a few changes
- On Franz' side: First, Stephan's changes are green, but after a few keystrokes they don't lose their color, instead they get blue, as if Julia's copy-paste actions have made blue the "Default Color".

See attached screenshot for Franz' view (was taken 12:45, thus the corresponding changes happend a few minutes before). 

Reported by: fzieris

Original Ticket: [dpp/bugs/757](https://sourceforge.net/p/dpp/bugs/757)

srossbach:
I was able to reproduce this always.

First defect, the annotation is always inserted twice by the EditorManager.
The EditorManager assumes that there is no anntotation at the given point which can be wrong.

Easy to reproduce

Mark some character (more than 1, eg. 10). Copy and paste them into the same document.
Go somewhere in the middle between the copied text. Start to insert characters.

Result: Eclipse will auto expand the annotation however the ContributionAnnotationManager will not update the history nor insert the annotation because there is already a given annotation at the current position.

srossbach:
![image](https://user-images.githubusercontent.com/41400679/63793820-c1caff00-c900-11e9-9971-8901800856c1.png)

PR #596, opened 2019-07-28
[FIX][E] #32 Splitting annotations do not honor the history

srossbach:
This patch fixes the last outstanding issues regarding #32.

Annotations that were split are now correctly inserted to the history
and so correctly removed.

This patch introduces a new method to the annotation helper as well as
it reduces LOC in the ContributionAnnotationManager.

### DIFF
--- a/eclipse/src/saros/editor/internal/AnnotationModelHelper.java
+++ b/eclipse/src/saros/editor/internal/AnnotationModelHelper.java
@@ -1,6 +1,7 @@
 package saros.editor.internal;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.Map;
@@ -24,7 +25,6 @@ public class AnnotationModelHelper {
   private static Iterable<Annotation> toIterable(final IAnnotationModel model) {
     return new Iterable<Annotation>() {
       @Override
-      @SuppressWarnings("unchecked")
       public Iterator<Annotation> iterator() {
         return model.getAnnotationIterator();
       }
@@ -56,6 +56,37 @@ public class AnnotationModelHelper {
   }
 
   /**
+   * Removes annotations and replaces them in one step.
+   *
+   * @param model The {@link IAnnotationModel} that should be cleaned.
+   * @param annotationsToRemove The annotations to remove.
+   * @param annotationsToAdd The annotations to add.
+   */
+  public void replaceAnnotationsInModel(
+      IAnnotationModel model,
+      Collection<? extends Annotation> annotationsToRemove,
+      Map<? extends Annotation, Position> annotationsToAdd) {
+
+    if (model instanceof IAnnotationModelExtension) {
+      IAnnotationModelExtension extension = (IAnnotationModelExtension) model;
+      extension.replaceAnnotations(
+          annotationsToRemove.toArray(new Annotation[annotationsToRemove.size()]),
+          annotationsToAdd);
+    } else {
+      if (LOG.isTraceEnabled())
+        LOG.trace("AnnotationModel " + model + " does not support IAnnotationModelExtension");
+
+      for (Annotation annotation : annotationsToRemove) {
+        model.removeAnnotation(annotation);
+      }
+
+      for (Entry<? extends Annotation, Position> entry : annotationsToAdd.entrySet()) {
+        model.addAnnotation(entry.getKey(), entry.getValue());
+      }
+    }
+  }
+
+  /**
    * Removes annotations that match a given predicate and replaces them in one step.
    *
    * @param model The {@link IAnnotationModel} that should be cleaned.
@@ -66,7 +97,6 @@ public class AnnotationModelHelper {
       Predicate<Annotation> predicate,
       Map<Annotation, Position> replacement) {
 
-    // Collect annotations.
     ArrayList<Annotation> annotationsToRemove = new ArrayList<Annotation>(128);
 
     for (Annotation annotation : AnnotationModelHelper.toIterable(model)) {
@@ -75,22 +105,7 @@ public class AnnotationModelHelper {
       }
     }
 
-    // Remove collected annotations.
-    if (model instanceof IAnnotationModelExtension) {
-      IAnnotationModelExtension extension = (IAnnotationModelExtension) model;
-      extension.replaceAnnotations(
-          annotationsToRemove.toArray(new Annotation[annotationsToRemove.size()]), replacement);
-    } else {
-      LOG.trace("AnnotationModel does not " + "support IAnnotationModelExtension: " + model);
-
-      for (Annotation annotation : annotationsToRemove) {
-        model.removeAnnotation(annotation);
-      }
-
-      for (Entry<Annotation, Position> entry : replacement.entrySet()) {
-        model.addAnnotation(entry.getKey(), entry.getValue());
-      }
-    }
+    replaceAnnotationsInModel(model, annotationsToRemove, replacement);
   }
 
   public IAnnotationModel retrieveAnnotationModel(IEditorPart editorPart) {
--- a/eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
+++ b/eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
@@ -1,23 +1,23 @@
 package saros.editor.internal;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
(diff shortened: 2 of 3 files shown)


eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -168,33 +172,54 @@ public void splitAnnotation(final IAnnotationModel model, final int offset) {
...
-        annotationsToAdd.put(new ContributionAnnotation(source, model), after);
-      }
-    }
+        annotationsToAdd.put(beforeAnnotation, before);
+        annotationsToAdd.put(afterAnnotation, after);
+
+        final History history = getHistory(source);
+
+        for (final ListIterator<HistoryEntry> lit = history.entries.listIterator();
+            lit.hasNext(); ) {
```
Drakulix: This could be a for-each loop instead of manually advancing the iterator.

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -228,88 +253,112 @@ public void refreshAnnotations(IAnnotationModel model) {
...
+
+    /* the logic assumes that the entry order does not change during lifetime
+     * i.e if we have a history of size 4 the list must look like this regarding the insert stamps
+     * 0 0 0 1 2 3 3 3 0 1 1 1 2 2 2 2 3 0 1 1 1 1 1 2 3 ... and so on
+     *
+     */
+
+    final Iterator<HistoryEntry> it = history.entries.iterator();
+
+    if (!it.hasNext()) return removedEntries;
```
Drakulix: So `if history.entries.isEmpty()`? Would be more readable in my opinion. (Maybe even add an `isEmpty` function to the history class.)

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -228,88 +253,112 @@ public void refreshAnnotations(IAnnotationModel model) {
...
+
+    while (it.hasNext()) {
+      final HistoryEntry entry = it.next();
```
Drakulix: Again. Why is this no for-each loop? You can break on those as well, if I am not mistaken.

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -228,88 +253,112 @@ public void refreshAnnotations(IAnnotationModel model) {
...
       log.warn("a annotation history for user " + oldAnnotation.getSource() + " does not exists");
 
       return;
     }
 
-    for (ListIterator<ContributionAnnotation> it = list.listIterator(); it.hasNext(); ) {
-      ContributionAnnotation annotation = it.next();
-      if (annotation.equals(oldAnnotation)) {
-        it.set(newAnnotation);
+    for (final ListIterator<HistoryEntry> it = history.entries.listIterator(); it.hasNext(); ) {
```
Drakulix: Okay, am I missing something at this point? What is the matter with all those manual iterator loops?

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -321,8 +370,20 @@ private void replaceInHistory(
...
+   * all users.
+   */
   private void removeAllAnnotations() {
-    for (Queue<ContributionAnnotation> queue : sourceToHistory.values())
-      while (!queue.isEmpty()) removeFromHistory(queue.peek());
+
+    final Set<IAnnotationModel> annotationModels = new HashSet<>();
+
+    for (final History history : sourceToHistory.values())
+      while (!history.entries.isEmpty())
```
m273d15: Is this an eternal loop? The def. of `peek` says `Retrieves, but does not remove, the head (first element) of this list.`.
Why not simply `for (HistoryEntry entry : history.entries)` ?

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
srossbach: Please look carefully at the code again, especially the code starting at line 203. This is not a simple iteration.

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
srossbach: THX !!!

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
srossbach: Well in this case I could you for:each.

I guess my intention was to replace the HistoryEntry but it seems I just replaced the annotation in the entry as this works as well.

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
Drakulix: Okay right, then please add some comments to the code.
It took me way too long to grasp, what this is doing (especially because I have no idea about annotations) and I think this could be explained. (I also feel like this could be way easier using functional-programming, but that's my personal opinion.)

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
Drakulix: maybe `for (HistoryEntry entry : history.entries.sublist(0, history.entries.size())` if your intention is to clear `history.entries` to avoid that for loop altogether.

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
srossbach: Functional Programming is normally harder to read.

removeAllAnnotation ended up with:

    sourceToHistory
        .values()
        .stream()
        .flatMap(h -> h.entries.stream())
        .collect(
            HashSet<IAnnotationModel>::new,
            (s, e) -> s.add(e.annotation.getModel()),
            HashSet::addAll)
        .forEach(
            m ->
                annotationModelHelper.removeAnnotationsFromModel(
                    m, (a) -> a instanceof ContributionAnnotation));


On the other hand optimization is done by manipulating the existing data structure.  That is something that you do not do in functional programming, e.g you have a big list where some entries should be removed. In a functional approach you create a new list containing all the entries without the entries that should be removed. While this may be fine for an ArrayList in terms of performance it is a huge performance loss for a LinkedList.

Drakulix: > Functional Programming is normally harder to read.

I would in general not agree with that statement. One could even argue your suggested functional code is more readable, because it does the same, while being shorter.

But I get, that a stream of streams is generally not what you want. (Which is why I just requested some comments on your old approach.)

> While this may be fine for an ArrayList in terms of performance it is a huge performance loss for a LinkedList.

While this is obviously true, this, in my opinion, an optimization the compiler should be able to infer.

eclipse/src/saros/editor/internal/AnnotationModelHelper.java
```diff
@@ -56,43 +56,59 @@ public void removeAnnotationsFromModel(IAnnotationModel model, Predicate<Annotat
...
+   * @param model The {@link IAnnotationModel} that should be cleaned.
+   * @param predicate The filter to use for cleaning.
+   */
+  public void replaceAnnotationsInModel(
+      IAnnotationModel model,
+      Predicate<Annotation> predicate,
+      Map<Annotation, Position> replacement) {
+
+    // Collect annotations.
+    ArrayList<Annotation> annotationsToRemove = new ArrayList<Annotation>(128);
```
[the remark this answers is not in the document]
srossbach: The default size of an array list is like 8 or something. This is simply a tiny performance gain (several nanoseconds) as you can expect to be at least more than 8 of or annotations in an editor / annotation model.

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
srossbach: The Java compiler does not really optimize anything. Even the JIT does not optimize much stuff. You are simply expecting too much from a compiler.

Drakulix: Not from a compiler, just from Java/JVM...

srossbach: Read again. Just In Time (compiler) ... it is part of the JVM.

Really the can not optimize every thing.

for (int i = 0; i < x; i++)
{
   Collection c = xyz.calculate(...) // will run in O(n^10) and always return the same result
   ... do something with c
}

I have seen such cases where the JIT is unable to move such calls out of the loop because the compiler fails to determine if calculate is idempotent in the current context and so you end up with a total running time of O(x * n^10) instead of O(x + n^10).

PR #597, opened 2019-07-28
[INTERNAL][E] Adds comments regarding Preferences

srossbach:
Preferences in Eclipse are quite complex. Without going too deep in
detail we have an issue regarding the preference scope in Eclipse. It is
currently scoped to the workspace and not the application.

This commit also fixes an issue with test cases that will fail if
assertions are enabled.

### DIFF
--- a/eclipse/src/saros/SarosEclipseContextFactory.java
+++ b/eclipse/src/saros/SarosEclipseContextFactory.java
@@ -64,7 +64,6 @@ public class SarosEclipseContextFactory extends AbstractContextFactory {
     return new Component[] {
       // Core Managers
       Component.create(IEditorManager.class, EditorManager.class),
-      Component.create(saros.preferences.Preferences.class, EclipsePreferences.class),
       Component.create(SessionViewOpener.class),
       Component.create(UndoManager.class),
       Component.create(ISarosSessionContextFactory.class, SarosEclipseSessionContextFactory.class),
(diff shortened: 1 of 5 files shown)


eclipse/src/saros/SarosEclipseContextFactory.java
```diff
@@ -131,15 +130,30 @@ public void createComponents(MutablePicoContainer container) {
...
 
-    // for core logic and extended Eclipse session components
+    final org.eclipse.jface.preference.IPreferenceStore instanceScopePreferenceStore =
+        saros.getPreferenceStore();
+
+    container.addComponent(
+        IPreferenceStore.class, new EclipsePreferenceStoreAdapter(instanceScopePreferenceStore));
+
+    /* FIXME the class states that it should be the global preferences, however it does not have method to write preferences and so it is fine for now to use the instance scope */
+    //    container.addComponent(
```
tobous: What is this commented out code doing here? Is this the logic that is supposed to be added once global preferences are possible?

srossbach: It would change the preference store to a global one, i.e shared across all Eclipse Workspaces.

I did not changed it yet because we should look what settings should be really global. And I do not really find it a good idea to talking about global stuff in the core Preferences. It should be up to the caller do decide what scope these preferences have.

srossbach:
The global Eclipse preferences are only used by the Feedback feature which is disabled. This is why I stopped here. I am currently does not see any reason to put any effort into offering 2 preference scopes (as I would have to change the core adding keybindings as already done with the transports).

I.e I would have to throw 2 PreferenceStores into the PICO container and this would require a lot of work unless there is a method where I can PICO which implementation is the default one when injecting dependencies.

Issue #497, opened 2019-04-12
Eclipse test SarosSessionTest.testCreateSarosSession runs into assertion error


tobous:
The test `testCreateSarosSession()` in eclipse/test/junit/saros/session/internal/SarosSessionTest.java runs into an assertion error. Weirdly, this does not cause the test case to fail. 

Stacktrace:
```
Internal Fatal Error:
java.lang.AssertionError: 
  Unexpected method call EditorManager.getOpenEditors():
	at org.easymock.internal.MockInvocationHandler.invoke(MockInvocationHandler.java:44)
	at org.easymock.internal.ObjectMethodsFilter.invoke(ObjectMethodsFilter.java:85)
	at org.easymock.internal.ClassProxyFactory$MockMethodInterceptor.intercept(ClassProxyFactory.java:94)
	at saros.editor.EditorManager$$EnhancerByCGLIB$$5a0a8f13.getOpenEditors(<generated>)
	at saros.concurrent.watchdog.ConsistencyWatchdogServer.calculateChecksums(ConsistencyWatchdogServer.java:205)
	at saros.concurrent.watchdog.ConsistencyWatchdogServer.access$200(ConsistencyWatchdogServer.java:38)
	at saros.concurrent.watchdog.ConsistencyWatchdogServer$1$1.run(ConsistencyWatchdogServer.java:77)
	at saros.util.ThreadUtils$1.run(ThreadUtils.java:35)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Original caller:
StackTrace:
	at saros.util.ThreadUtils.wrapSafe(ThreadUtils.java:29)
	at saros.concurrent.watchdog.ConsistencyWatchdogServer$1.run(ConsistencyWatchdogServer.java:71)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
[...]

PR #598, opened 2019-07-28
[FIX][E] Fixes #469 by reintroduces missing context menu

### LINKED ISSUE #469: The "Work Together On.." dropdown list contains only the first project

srossbach:
This time without the plugin.xml magic. The context menu was only shown
when a JID was selected which is only the case when you select something
in the Saros View, to be more specific in the roster.

Restores the original behavior, the only thing that is missing is the
tooltip for the menu entry.

### DIFF
--- a/eclipse/plugin.xml
+++ b/eclipse/plugin.xml
@@ -1002,6 +1002,7 @@
       <menuContribution
             allPopups="false"
             locationURI="popup:org.eclipse.ui.popup.any?before=menustart">
+         <!-- See SarosView, the context menu is created manually 
          <command
                commandId="saros.ui.commands.ShareResource"
                disabledIcon="icons/dlcl16/spacer.png"
@@ -1023,6 +1024,7 @@
                </and>
             </visibleWhen>
          </command>
+         -->
          <command
                commandId="saros.ui.commands.SessionAddSelectedContacts"
                disabledIcon="icons/dlcl16/session_add_contacts_tsk.png"
--- a/eclipse/src/saros/ui/ImageManager.java
+++ b/eclipse/src/saros/ui/ImageManager.java
@@ -78,8 +78,13 @@ public class ImageManager {
       getImage("icons/elcl16/contact_add_tsk.png"); // $NON-NLS-1$
   public static final Image DLCL_CONTACT_ADD =
       getImage("icons/dlcl16/contact_add_tsk.png"); // $NON-NLS-1$
+
   public static final Image ELCL_SESSION = getImage("icons/elcl16/session_tsk.png"); // $NON-NLS-1$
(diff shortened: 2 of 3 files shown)


eclipse/plugin.xml
```diff
@@ -1002,6 +1002,7 @@
       <menuContribution
             allPopups="false"
             locationURI="popup:org.eclipse.ui.popup.any?before=menustart">
+         <!-- See SarosView, the context menu is created manually
```
m273d15: Why are you using a comment instead of simply removing the stuff and adding the comment?

eclipse/src/saros/ui/ImageManager.java
```diff
@@ -78,8 +78,13 @@
       getImage("icons/elcl16/contact_add_tsk.png"); // $NON-NLS-1$
   public static final Image DLCL_CONTACT_ADD =
       getImage("icons/dlcl16/contact_add_tsk.png"); // $NON-NLS-1$
+
   public static final Image ELCL_SESSION = getImage("icons/elcl16/session_tsk.png"); // $NON-NLS-1$
+  public static final ImageDescriptor ELCL_SESSION__DESCRIPTOR =
```
m273d15: Just a question: Do you know the meaning of the abbreviations (ELCL, DLCL)?

m273d15:
Can we close #497 after merging the PR?

eclipse/plugin.xml
srossbach: Well I can removed it. I just did not removed it in case of E4 migration but I guess this will never happen.

I tried to find some guides (especially mixing E3 and E4 stuff) and I failed badly. Most guides are for RCP applications and I do not know if this also applies to plugins.

eclipse/src/saros/ui/ImageManager.java
srossbach: See https://wiki.eclipse.org/User_Interface_Guidelines Section Folder Structure

eclipse/plugin.xml
m273d15: Yes, I have a similar experience. I also tried [this](https://jaxenter.de/tutorial-so-portieren-sie-eclipse-3-anwendungen-auf-eclipse-4-39492) german guide. I was able to start the SarosView as a e4view, but the Toolbar was still missing. I am still thinking about the relation of `application.e4xmi`, `fragment.e4xmi` and the `plugin.xml`. I will try it again later.

PR #600, opened 2019-07-29
[INTERNAL][HTML] Subtitute JQueryBrowser with IBrowser

m273d15:
* Substitute JQueryBrowser (of swt-browser-improved) with IBrowser.
* Added missing Selector
* Removed dependency to swt-browser-improved


m273d15:
Yes see [here](https://github.com/saros-project/saros/blob/a6a3e9268881e2273d51e221a23350967c40f019/ui.frontend/html/package.json#L34). I think we will find a better solution in the future. Bundling jquery into our application only for testing purposes is not an ideal solution.

ui/src/saros/stf/server/bot/jquery/ISelector.java
```diff
@@ -0,0 +1,114 @@
+package saros.stf.server.bot.jquery;
+
+import org.apache.commons.lang3.StringUtils;
+
+public interface ISelector {
```
tobous: My guess would be that this was copied from the swt-browser-improved. If this is the case, this should be mentioned in the class javadoc with a reference to the repo in order to provide context to other people looking at the class. If this is true for the whole package, you could also add a short note as package-info (and to the other class).

Subsequently, you could remove the `@author` tags from the class as they are not generally used in this project.

m273d15: I will add the information to `ISelector` `AbstractJavaScriptFunction` and `NoCheckedExceptionCallable` and remove `@author`

PR #602, opened 2019-07-29
[INTERNAL][E] Enables SkypeManager

srossbach:
This patch enables the Skype feature again. Reworked most of the
SkypeManager logic by just throwing away code.

### DIFF
--- a/eclipse/src/saros/communication/SkypeManager.java
+++ b/eclipse/src/saros/communication/SkypeManager.java
@@ -2,6 +2,8 @@ package saros.communication;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
 import java.util.Iterator;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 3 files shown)


m273d15:
Independent of this specific PR: Thank you for all these changes in recent days :+1:

Would it be better to just support the transfer of "other account data" (skype, teamspeak, ...) in general instead of limiting the function to skype.
Generating the additional skype uris would be just an additional feature on top.
Furthermore, is it necessary to communicate the additional user information explicit via smack? It just seems to be very low-level...

srossbach:
Well you could use the IReceiver and ITransmitter for doing that. However you still need the roster for tracking the people and so in the end even if you user IReceiver and ITransmitter you basically ending up with the some LOCs. (try it out by using the RosterTracker and IReceiver and ITransmitter instead).

As you mentioned TS3. I did not even know that you can also start Teamspeak via an URL. I tried it out with Discord however I can only start the application (and if the application is started the command stalls ...)

So what do you suggest, allowing multiple "URLs" to be broadcasted or add more mediums, e.g

In the context menu we then have something like:

TBD -> Skype -> ..
        |
        -> TS3 -> ...

m273d15:
The idea was just to allow the user to choose between:
1. broadcast a skype name
2. broadcast an arbitrary service accout name

I think providing the uri in order to start a skype session is a nice shiny feature, but broadcasting the account name is the most basic feature we should provide for other services. 

**Case 2:**
Broadcasting a tuple (service name, service account). The other users just see the additional account name with the information which service is used. A contact entry could have the format  `(<jid>|<nickname>) - <service account> (<service name>)` (e.g. `Kelvin - m273d15 (TeamSpeak)`)
No further function of generating url or something like this.

**Case 1.:**
Should use the logic of **Case 2**. We should be able to detect that the context menu on a contact is clicked that provides a skype account. Based on this information we provide the function to generate the urls.

srossbach:
I do not get your example in case 2. Can you please be more specific, e.g Kelvin - m273d15 (TeamSpeak) ? Is this the available data ? Then how should I setup a TS session without an INET_Address ? Furthermore passing just some URI and "executing" makes the system vulnerable.

m273d15:
I just mean that we broadcast two Strings. One that identifies an arbitrary service and an account identifier. There is no URI involved.
The service name and the account identifier are shown in the contact list. With `Kelvin - m273d15 (TeamSpeak)` I shown an example how an entry in the contact list could look like.

srossbach:
@m273d15 Can we agree that your suggestions are basically a new feature which require a bunch of changes to the current UI (which should be replaced *soon* by the HTML UI).

And what should be displayed if I offer multiple ways on contacting me, e.g Skype and Teamspeak. Please be aware that the we always have the issue that we are very limited when it comes to space for displaying stuff (at least in Eclipse).

m273d15:
> Can we agree that your suggestions are basically a new feature

Yes, we should discuss this independent from this PR.

srossbach:
Hmm somehow my modifications got lost as I accidentally pulled the remote branch. Will add them in a dedicated commit.

srossbach:
Only added URL encoding for the skype name.

PR #606, opened 2019-07-30
[INTERNAL][E] Adds central connection access point for the UI

srossbach:
Adds a central class that is now responsible for connection attempts.

This should help to prevent certain misuse, e.g disconnecting an active
session without any warning.

In addition this patch removes the inheritance between the
XMPPAddAccountWizard and the SarosConfigurationWizard.


srossbach:
I need to merge this now as the AccountStore needs an API and the other way around it is more complicated. This patch should not alter existing behavior aside adding XMPPAccounts to the store now will never trigger a connection attempt regardless of the auto-connect preference.

m273d15:
I don't really understand why you merged the PR without any review ...

stefaus:
@srossbach Seriously?

PR #608, opened 2019-07-31
[INTERNAL][CORE] Amends XMPPAccountStore API

srossbach:
Introduces get/set defaultAccount methods which should be a replacement
for set/get activeAccount in the future. 

They allow null values making the call to isEmpty obsolete. And now it
is also possible to completely remove all account data.

Add getAccount methods as there was no way to retrieve specific account
unless you called getAllAccounts and perform the search by yourself.


core/src/saros/account/XMPPAccountStore.java
```diff
@@ -355,32 +350,30 @@ public void setAccountActive(XMPPAccount account) {
...
     }
 
     notifyAccountStoreListeners();
+    if (defaultAccount == null) notifyActiveAccountListeners();
   }
 
   /**
-   * Creates an account. The account will automatically become active if the account store is empty.
+   * Creates an account. The account will automatically become the default if the account store is
+   * empty. If the account already exists the existing account is returned.
```
srossbach: Wrong

core/src/saros/account/XMPPAccountStore.java
```diff
@@ -318,35 +293,55 @@ public int compare(XMPPAccount a, XMPPAccount b) {
...
+
+    notifyActiveAccountListeners();
   }
 
   /**
    * Makes the given account active.
    *
    * @param account the account to activate
    * @throws IllegalArgumentException if the account is not found in the store
+   * @deprecated Will be removed soon, do not use
```
tobous: Refer to `setDefaultAccount(XMPPAccount)` as an alternative? (hinges on my general question on the meaning of "default" vs "active")

Also, this seems to do exactly the same as `setDefaultAccount(XMPPAccount)` minus allowing `null` values. Why not just replace the old calls with the new method directly and remove this method (again, in the separate refactoring commit)?

core/src/saros/account/XMPPAccountStore.java
```diff
@@ -489,16 +482,34 @@ public void changeAccountData(
...
+  public synchronized XMPPAccount getDefaultAccount() {
+    return defaultAccount;
+  }
+
   /**
    * Returns the current active account.
    *
    * @return the active account
    * @throws IllegalStateException if the account store is empty
+   * @deprecated Will be removed soon, do not use
```
tobous: Refer to `getDefaultAccount()` as the new alternative?
```suggestion
   * @deprecated Will be removed soon, use {@link getDefaultAccount} instead
```

core/src/saros/account/XMPPAccountStore.java
```diff
@@ -543,7 +585,9 @@ public synchronized boolean exists(String username, String domain, String server
    * @param jidString the jid of the user as string
    * @return the matching XMPP account or null in case of no match
    * @throws NullPointerException if jidString is null
+   * @deprecated Will be removed soon, do not use
```
tobous: Refer to alternative methods?

core/src/saros/account/XMPPAccountStore.java
srossbach: Because somehow Gradle fucked up my Eclipse Build Path for IntelliJ, I am currently unable to compile it because of every library entry exists twice.

To be more specific: soon = not the normal student soon aka never but right after the commit that follows this commit.

core/src/saros/account/XMPPAccountStore.java
srossbach: Just a riddle for you. However it is not complicated, I offered two get methods. One should solve your needs.

core/src/saros/account/XMPPAccountStore.java
```diff
@@ -166,7 +164,10 @@ private synchronized void loadAccounts() {
                   new ByteArrayInputStream(Crypto.decrypt(encryptedAccountData, secretKey)));
 
       accounts = new HashSet<>(accountData.configuredAccounts);
-      activeAccount = accountData.configuredAccounts.get(accountData.activeAccountIndex);
+      defaultAccount = null;
+
+      if (accountData.activeAccountIndex != -1)
```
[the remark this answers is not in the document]
srossbach: Really ? The methods that access this variable and do not match the naming are already deprecated.

srossbach:
Well, basically you have the option to either connect with a default / active account or another. The current logic in Eclipse just uses the default/active one all time. If you select another account (just for testing purposes or whatever) this account will be the default one. I do not know why this was implemented in that way. Maybe I have to dig down through my rewrite of the XMPPAccountStore. But my guess is that: activeAccount was always present and my assumption was that it a default account.

My big guess is that someone amended the connect action and run into problem when not marking the account to connect as active. Currently the only way to make an account active through the GUI is hidden in the preferences.

core/src/saros/account/XMPPAccountStore.java
tobous: > Because somehow Gradle fucked up my Eclipse Build Path for IntelliJ, I am currently unable to compile it because of every library entry exists twice.

Do you mean your build path for the IntelliJ project or your build path/config in IntelliJ (i.e. for building Saros in IntelliJ IDEA)?

srossbach: image

tobous: Hmm. Delete the directory containing the gradle cache and build again? Also, did this affect other parts of the config (tracked by git)?

core/src/saros/account/XMPPAccountStore.java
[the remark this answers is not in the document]
tobous: I don't have the complete overview of what the next steps are and how long parts this code will remain unchanged.

If this refactoring is made unnecessary by an upcoming change in the near future, you can just leave it like this (this also goes for the other refactoring requests in this PR). I was looking at this from the perspective that the code would remain like this for a while.

core/src/saros/account/XMPPAccountStore.java
srossbach: I delete the .project file and called gradle again, same issue. I manually edited the .project and now thinks works. TBH I do not get those mass amounts of libraries. IIRC you only need about 20 libraries to compile IntelliJ in the Eclipse IDE.

core/src/saros/account/XMPPAccountStore.java
```diff
@@ -355,32 +350,30 @@ public void setAccountActive(XMPPAccount account) {
...
+      if (!accounts.remove(account))
         throw new IllegalArgumentException(
             "account '" + account + "' is not in the current account store");
 
-      if (this.activeAccount == account)
-        throw new IllegalStateException(
-            "account '" + account + "' is active and cannot be deleted");
-
-      accounts.remove(account);
+      if (defaultAccount == account) defaultAccount = null;
```
tobous: Is there a specific reason to compare the object references instead of using `equals()`? (See [Codacy issues](https://app.codacy.com/app/Saros/saros/pullRequest?prid=3938137&bid=13677390)) Same goes for l.417.
(If its just to avoid the nullcheck, you can use `Objects.equals(Object, Object)` instead.)

srossbach: Because in this scope using == or equals will result in the same behavior. This will of course break if you make the XMPPAccount CTOR public.

tobous: I would still prefer to use `equals(Object)` or `Objects.equals(Object, Object)` to be on the safe side. The performance loss should not be of concern.

PR #609, opened 2019-07-31
[INTERNAL][UI] Use new XMPPStore API

srossbach:
This patch introduces the usage of the new XMPPStore API.

Renaming UI buttons is not done yet as changing UI values is likely to
crash the whole STF.

### DIFF
--- a/eclipse/src/saros/ui/actions/ChangeXMPPAccountAction.java
+++ b/eclipse/src/saros/ui/actions/ChangeXMPPAccountAction.java
@@ -4,6 +4,7 @@ import org.apache.log4j.Logger;
 import org.eclipse.jface.action.Action;
 import org.eclipse.jface.action.ActionContributionItem;
 import org.eclipse.jface.action.IMenuCreator;
+import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Menu;
@@ -74,12 +75,23 @@ public class ChangeXMPPAccountAction extends Action implements IMenuCreator, Dis
   @Override
   public void run() {
 
+    final XMPPAccount defaultAccount = accountService.getDefaultAccount();
+    final boolean isEmpty = accountService.isEmpty();
+
+    if (defaultAccount == null || isEmpty) {
+      if (!MessageDialog.openQuestion(
+          SWTUtils.getShell(),
+          "Default account missing",
+          "A default account has not been set yet. Do you want set a default account?")) return;
+
+      SWTUtils.runSafeSWTAsync(LOG, this::openPreferences);
+      return;
+    }
+
     if (connectionHandler.isConnected()) {
       XMPPConnectionSupport.getInstance().disconnect();
     } else {
-      XMPPConnectionSupport.getInstance()
-          .connect(
-              accountService.isEmpty() ? null : accountService.getActiveAccount(), true, false);
+      XMPPConnectionSupport.getInstance().connect(accountService.getDefaultAccount(), true, false);
     }
   }
 
@@ -97,12 +109,16 @@ public class ChangeXMPPAccountAction extends Action implements IMenuCreator, Dis
   public Menu getMenu(Control parent) {
     accountMenu = new Menu(parent);
 
-    XMPPAccount activeAccount = null;
+    XMPPAccount defaultAccount = null;
 
-    if (connectionHandler.isConnected()) activeAccount = accountService.getActiveAccount();
+    /* FIXME obtain the current JID and the discard the entry.
+     * This logic here only works because we set the account that should connect to be the default one.
+     * If the user is interested in such a behavior is another question.
+     */
+    if (connectionHandler.isConnected()) defaultAccount = accountService.getDefaultAccount();
 
     for (XMPPAccount account : accountService.getAllAccounts()) {
-      if (!account.equals(activeAccount)) addMenuItem(account);
+      if (!account.equals(defaultAccount)) addMenuItem(account);
     }
 
     new MenuItem(accountMenu, SWT.SEPARATOR);
--- a/eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
+++ b/eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
@@ -64,15 +64,17 @@ public class ConnectingFailureHandler implements IConnectingFailureCallback {
         return;
       }
 
-      /* FIXME the active/default account is not always the account that is currently used for connecting */
       if (DialogUtils.popUpYesNoQuestion(
           "Connecting Error",
           generateHumanReadableErrorMessage((XMPPException) exception),
           false)) {
 
-        if (WizardUtils.openEditXMPPAccountWizard(accountStore.getActiveAccount()) == null) return;
+        /* FIXME the active/default account might not always be the account that is currently used for connecting */
+        final XMPPAccount accountUsedDuringConnection = accountStore.getDefaultAccount();
 
-        final XMPPAccount account = accountStore.getActiveAccount();
+        if (WizardUtils.openEditXMPPAccountWizard(accountUsedDuringConnection) == null) return;
+
+        final XMPPAccount account = accountStore.getDefaultAccount();
 
         SWTUtils.runSafeSWTAsync(
             LOG, () -> XMPPConnectionSupport.getInstance().connect(account, false));
--- a/eclipse/src/saros/ui/util/XMPPConnectionSupport.java
+++ b/eclipse/src/saros/ui/util/XMPPConnectionSupport.java
@@ -42,7 +42,11 @@ public class XMPPConnectionSupport {
   }
 
   /**
-   * Connects with the current active / default account.
+   * Connects with the default account.
+   *
+   * <p><b>Note:</b> If no default account is available this method silently returns regardless if
+   * <code>
+   * failSilently</code> is set to <code>false</code> !
    *
    * @param failSilently if <code>true</code> suppresses any further error handling
    */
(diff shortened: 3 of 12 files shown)


tobous:
@m273d15 Any idea why the Codacy build fails? I tried restarting it, but it did not help. Otherwise, I will contact the codacy support.

@srossbach Did you change the base of this PR directly after pushing to the PR branch (while the build was still running)? This caused issue for me in the past.

srossbach:
I do not know what cause this. I am already facing the problem that my commit message is not insert here. I will have to look at EGIT again if it is possible to add a reference on a remote branch. Currently I am doing this in Github when creating the pull request. Maybe it is possible that I changed the branch during build.

tobous:
What do you mean by "that your commit message is not inserted here"? Do you mean that the commit message is actually not attached to the commit or that the commit message is not automatically entered in the PR description?

If its the second one, this is due to us now having a PR template that is used instead. So you have to do the same you also have to do when creating PRs containing multiple commits, i.e. copy the message and put it in the PR description by hand. At least that is how I have been doing it. Maybe there is a better way or a plugin that helps you with it, but this would be news to me.

srossbach:
Sometimes my commit message is included in the PR and the template is appended. Sometimes only the template is present.

eclipse/src/saros/ui/util/XMPPConnectionSupport.java
```diff
@@ -129,16 +129,15 @@ public void connect(final XMPPAccount account, boolean setAsDefault, boolean fai
 
     final XMPPAccount accountToConnect;
 
-    if (account == null && !store.isEmpty()) accountToConnect = store.getActiveAccount();
-    else if (account != null) accountToConnect = account;
-    else accountToConnect = null;
+    if (account == null) accountToConnect = store.getDefaultAccount();
```
tobous: Causes NPE in line 151 (`connectionHandler.connect(accountToConnect, failSilently);`) when `account` is null and default account is not set? (NPE in [ConnectionHandler.connectXMPPInternal](https://github.com/saros-project/saros/blob/master/core/src/saros/communication/connection/ConnectionHandler.java#L190))

Also NPE in [l.117](https://github.com/saros-project/saros/blob/627d04768438cc842167e91fd39dcba16e249c59/eclipse/src/saros/ui/util/XMPPConnectionSupport.java#L117) if `account` is null (if I am not mistaken).

eclipse/src/saros/ui/actions/ChangeXMPPAccountAction.java
```diff
@@ -74,12 +75,23 @@ public ChangeXMPPAccountAction() {
...
+    final XMPPAccount defaultAccount = accountService.getDefaultAccount();
+    final boolean isEmpty = accountService.isEmpty();
+
+    if (defaultAccount == null || isEmpty) {
+      if (!MessageDialog.openQuestion(
+          SWTUtils.getShell(),
+          "Default account missing",
+          "A default account has not been set yet. Do you want set a default account?")) return;
+
+      SWTUtils.runSafeSWTAsync(LOG, this::openPreferences);
```
tobous: Does this force the user to set a default account? Otherwise, there could still not be a default account requesting it in l. 94, leading to a NPE in XMPPConnectionSupport (see comment in file).

eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
```diff
@@ -64,15 +64,17 @@ private void handleConnectionFailed(Exception exception) {
...
           "Connecting Error",
           generateHumanReadableErrorMessage((XMPPException) exception),
           false)) {
 
-        if (WizardUtils.openEditXMPPAccountWizard(accountStore.getActiveAccount()) == null) return;
+        /* FIXME the active/default account might not always be the account that is currently used for connecting */
+        final XMPPAccount accountUsedDuringConnection = accountStore.getDefaultAccount();
 
-        final XMPPAccount account = accountStore.getActiveAccount();
+        if (WizardUtils.openEditXMPPAccountWizard(accountUsedDuringConnection) == null) return;
```
tobous: Throws assertion error when `accountUsedDuringConnection` is `null`: See [EditXMPPAccountWizard l.27](https://github.com/saros-project/saros/blob/master/eclipse/src/saros/ui/wizards/EditXMPPAccountWizard.java#L27)

eclipse/src/saros/ui/wizards/ConfigurationWizard.java
```diff
@@ -72,19 +72,18 @@ public void addPages() {
...
+     */
+
+    accountStore.setDefaultAccount(accountToConnect);
 
     if (preferences.getBoolean(PreferenceConstants.AUTO_CONNECT)) {
-      getShell()
-          .getDisplay()
-          .asyncExec(
-              () -> XMPPConnectionSupport.getInstance().connect(accountToConnect, true, false));
+      getShell().getDisplay().asyncExec(() -> XMPPConnectionSupport.getInstance().connect(false));
```
tobous: Why not pass the determined `accountToConnect` when it is set as the default account, meaning it will be used anyways?

srossbach: Hmm ? See logic above, always make the account from the ConfigurationWizard the default one.

Please read: AUTO_CONNECT = AUTO_CONNECT_ON_STARTUP 

Normally I should present an option on the last page with buttons Finish&Connect Now and Finish (but I guess this is against the Wizard guidelines). So I should add an button for that.

eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
srossbach: Was misguided by comment on line 43. Yeah I will add null check in the XMPPConnectionSupport and abort the connection attempt.

eclipse/src/saros/ui/actions/ChangeXMPPAccountAction.java
srossbach: You cannot reach line 94 when you enter the if block on line 81. Guess I should make a guard that this is called from SWT anyway so I can remove the asyncCall.

tobous: Wupsi. My bad. Apparently I just missed that return statement. :wink:

tobous: Wupsi. My bad. Apparently I just missed that return statement. :wink:

eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
srossbach: This is actually already fixed in a follow up.

intellij/src/saros/intellij/ui/actions/ConnectServerAction.java
```diff
@@ -34,15 +35,22 @@ public String getActionName() {
...
   /** Connects with the given user. */
   public void executeWithUser(String user) {
-    XMPPAccount account = accountStore.findAccount(user);
-    accountStore.setAccountActive(account);
+    JID jid = new JID(user);
+    XMPPAccount account = accountStore.getAccount(jid.getName(), jid.getDomain());
+
+    if (account == null) return;
+
+    accountStore.setDefaultAccount(account);
```
srossbach: I do not know when this called ? Is the "user" fetched from the store or how is the data obtained. If it from the store you just mimic the Eclipse behavior aka if you connect with another account this one will become the default account.

srossbach: Ok found it. Same behavior as Eclipse. Just copied or intended ?

eclipse/src/saros/ui/util/XMPPConnectionSupport.java
tobous: > Also NPE in [l.117](https://github.com/saros-project/saros/blob/627d04768438cc842167e91fd39dcba16e249c59/eclipse/src/saros/ui/util/XMPPConnectionSupport.java#L117) if `account` is null (if I am not mistaken).

What about this? Or did I miss something in the control flow (it's not that easy to read)? Or do you consider this out of scope for this patch?

srossbach: Line 135 ?

tobous: Ok, I did not see that comment.

How about moving it to the javadoc (or some other place more visible) instead of removing it if it is still valid? Because, to me, this seems like a very weird/unexpected control flow that should probably be mentioned somewhere.

srossbach: No its not valid. The problem is I wrote this comment in the ConnectingFailureHandler so I am pretty sure it worked once this way. However it is no longer the case and after regarding this issue it maybe not the best spot to open the ConfigurationWizard in such a place. This should be done elsewhere in already the case in the startup. You will not see it because if you run with STF it is ommited.

srossbach:
I just noticed I missed the server. Will add it in the next commit.

eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
tobous: For future occurrences, it would be useful to also provide a link to the followup PR.

srossbach: This is another wizard. Seems we have to many wizards ...

The ConfigurationWizard is the initial wizard where you can add/create an XMPPAccount, then choose your session color, etc.

This here is the EditXMPPAccountWizard and this have to be opened.

srossbach:
Should be easily revealed when rebasing #615 .

eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
srossbach: My bad, I am totally tired. I had a hard time to understand what you want me to do ... sorry. Yeah I will add a link next time to make reviewing easier.

tobous:
@srossbach From my experience, the Travis build for the Pull Request will never come in. This is an issue that seems to occur when you (force-) push to a branch and then immediately change the base branch while it is still building.

This can only be solved by pushing something new to the branch (i.e. amending a commit and force-push again). But, since the branch build succeeded, this can probably be ignored. It should still be avoided if possible (e.g. by changing the base branch before pushing/updating the PR).

srossbach:
@tobous thank you for the information. I am still used to Gerrit which updated the branches automatically.

PR #613, opened 2019-08-02
[INTERNAL][E] Addresses FIXME in ConnectingFailureHandler

srossbach:
[INTERNAL][E] Addresses FIXME in ConnectinFailureHandler

Added an option to retrieve the current account that was used for
connecting. Another option would have been to modify the ConnectHandler.

As IntelliJ currently have a very minimalistic set of connection option
the necessary logic was put into the XMPPConnectionSupport instead.

### DIFF
--- a/eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
+++ b/eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
@@ -2,13 +2,12 @@ package saros.ui.eventhandler;
 
 import java.text.MessageFormat;
 import org.apache.log4j.Logger;
+import org.eclipse.jface.dialogs.MessageDialog;
 import org.jivesoftware.smack.XMPPException;
 import org.jivesoftware.smack.packet.XMPPError;
 import saros.account.XMPPAccount;
-import saros.account.XMPPAccountStore;
 import saros.communication.connection.ConnectionHandler;
 import saros.communication.connection.IConnectingFailureCallback;
-import saros.ui.util.DialogUtils;
 import saros.ui.util.SWTUtils;
 import saros.ui.util.WizardUtils;
 import saros.ui.util.XMPPConnectionSupport;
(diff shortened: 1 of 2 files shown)


eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
```diff
@@ -19,66 +18,56 @@
...
-
-        DialogUtils.popUpFailureMessage(
+        MessageDialog.openError(
+            SWTUtils.getShell(),
             "Connecting Error",
             MessageFormat.format(
-                "Could not connect to XMPP server. Unexpected error: {0}", exception.getMessage()),
-            false);
-
+                "Could not connect to XMPP server. Unexpected error: {0}", exception.getMessage()));
```
tobous: This should be moved to a messages.properties file as it is text shown to the user.

eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
```diff
@@ -19,66 +18,56 @@
...
-          "Connecting Error",
-          generateHumanReadableErrorMessage((XMPPException) exception),
-          false)) {
+      final XMPPAccount accountUsedForConnecting =
+          XMPPConnectionSupport.getInstance().getCurrentXMPPAccount();
 
-        /* FIXME the active/default account might not always be the account that is currently used for connecting */
-        final XMPPAccount accountUsedDuringConnection = accountStore.getDefaultAccount();
+      // should not happen
+      if (accountUsedForConnecting == null) return;
```
tobous: Log a warning to show that an unexpected state occurred in order to allow us to noticed/investigated this behavior?

eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
```diff
@@ -19,66 +18,56 @@
...
 
-        if (WizardUtils.openEditXMPPAccountWizard(accountUsedDuringConnection) == null) return;
+      final String errorMessesage = generateHumanReadableErrorMessage((XMPPException) exception);
 
-        final XMPPAccount account = accountStore.getDefaultAccount();
+      final boolean editAccountAndConnectAgain =
+          MessageDialog.openQuestion(SWTUtils.getShell(), "Connecting Error", errorMessesage);
```
tobous: Move to a messages.properties file as it is also text shown to the user.

eclipse/src/saros/ui/eventhandler/ConnectingFailureHandler.java
srossbach: Can we agree that moving strings into property files can be done just before the next release or at least later ? There is automated tooling provided by Eclipse for this which also inserts a bunch of NLS.

tobous: Ok. How about opening an issue for this? It would also be helpful to link to some documentation on how this can be done automatically as I had no idea this was an option in Eclipse. ☺️

srossbach: The problem is the Google Formatter.

E.g you have "foo0" + "foo1" + .... + "fooN"

When say you do not want to externalize them Eclipse will add something like this // $NON-NLS-1$, $NON-NLS-2$, ... behind this line.

I have not really tested it what will happen if the comment will be split to several lines afterwards.

srossbach: Wow nice job. https://github.com/google/google-java-format/issues/361

Currently there is now way to tell if the string already was externalized or not.

PR #614, opened 2019-08-02
[UI][E] Moves information from EnterXMPPAccountWizardPage

srossbach:
[UI][E] Moves information from ENterXMPPAccountWizardPage

This information is only needed for the configuration wizard. It is very
confusing if you open the page in stand-alone mode and reading "If you
already have a XMPP account ..." 

I am guessing I would have used Create-Account instead of Add-Account if
I do not own a XMPP account.

### DIFF
--- a/eclipse/src/saros/ui/wizards/ConfigurationWizard.java
+++ b/eclipse/src/saros/ui/wizards/ConfigurationWizard.java
@@ -1,9 +1,14 @@
 package saros.ui.wizards;
 
+import java.util.Arrays;
 import org.bitlet.weupnp.GatewayDevice;
 import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.jface.wizard.Wizard;
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.layout.GridData;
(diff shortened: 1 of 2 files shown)


eclipse/src/saros/ui/wizards/ConfigurationWizard.java
```diff
@@ -30,7 +36,30 @@
 public class ConfigurationWizard extends Wizard {
 
   private final EnterXMPPAccountWizardPage enterXMPPAccountWizardPage =
-      new EnterXMPPAccountWizardPage();
+      new EnterXMPPAccountWizardPage() {
+        @Override
+        public void createControl(final Composite parent) {
```
tobous: Add a javadoc/comment addressing what this does and why it is necessary?

srossbach: The magic happens on the last line of this method. We are "injecting" something into this composite. Guess I will pull the comment on line 44 up here.

tobous: Yeah I read the comment but I couldn't really tell what exactly this hack does and why we need it. The commit message hints at it but I still think it wouldn't hurt to spell it out somewhere in the code.

PR #616, opened 2019-08-05
[FIX] #98 UPnPTest#testDispose() Fails if Assertions are Enabled

### LINKED ISSUE #98: Core Test UPnPTest#testDispose() Fails if Assertions are Enabled

srossbach:
This fixes #98 by implementing the missing functionality to handle
multiple IGD that support lease durations.

This patch also fixes and issue I have found using a FritzBox. It seems
that the Fritzbox is renaming the MappingDescription and so we would
face problems if want to apply the mapping again without removing it
first (e.g due program crash).

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -49,9 +49,8 @@ jar {
 }
 
 
-// Avoid issue GH issue #98
 tasks.withType(Test) {
-  enableAssertions = false
+  enableAssertions = true
 }
 
 task testJar(type: Jar) {
--- a/core/src/saros/net/upnp/internal/UPnPServiceImpl.java
+++ b/core/src/saros/net/upnp/internal/UPnPServiceImpl.java
@@ -10,64 +10,43 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
-import java.util.Timer;
-import java.util.TimerTask;
+import java.util.concurrent.ExecutionException;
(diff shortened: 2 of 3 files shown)


tobous:
I have not looked at the code itself yet. But, since the fixed issue was the only thing that forced us to run tests without assertions enabled for the core, they should be re-enabled as part of this patch (which also automatically tests the patch with the travis builds). This can be done by adjusting the [`build.gradle`](https://github.com/saros-project/saros/blob/master/core/build.gradle#L53) of the core by removing lines 53-55. The same should also be done for the eclipse build ([lines 69-71](https://github.com/saros-project/saros/blob/master/eclipse/build.gradle#L69)) as it can already run with assertions enabled with the current master.

srossbach:
Does not break any existing functionality.

Issue #28, opened 2012-04-05
Make Balloons not overlap editorparts


dpp-gerrit:
It would be nice if users would not have to complain about the balloon notifications overlapping their editor viewpart, which they find too obstrusive.

Woudl be good if the ballooons would hook somewhere else, for example to the bottom of the saros session. Or even better: if  a user is affected, let the balloon point to the user of the session, or the buddy in the buddylist.

Reported by: netcorps

Original Ticket: [dpp/bugs/729](https://sourceforge.net/p/dpp/bugs/729)

PR #619, opened 2019-08-05
[FIX][E] #28 Make Balloons not overlap editorparts

srossbach:
This fixes issue #28.

The popups are now displayed at the bottom of the view. As there was not
functionality to specify the "hooking point" an invisble widget was
added to the bottom of the view which is used as reference point.

### DIFF
--- a/eclipse/src/saros/ui/views/SarosView.java
+++ b/eclipse/src/saros/ui/views/SarosView.java
@@ -24,6 +24,8 @@ import org.eclipse.swt.events.MouseAdapter;
 import org.eclipse.swt.events.MouseEvent;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.layout.FillLayout;
+import org.eclipse.swt.layout.GridData;
+import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Display;
@@ -238,6 +240,8 @@ public class SarosView extends ViewPart {
 
   private static volatile boolean showBalloonNotifications;
 
+  private Composite notificationAnchor;
+
   /**
    * Stores actions by their {@link IAction#getId() ID}, so they can (1) be {@linkplain
    * #getAction(String) retrieved} and (2) {@linkplain Disposable#dispose() disposed} when and if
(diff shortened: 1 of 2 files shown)


eclipse/src/saros/ui/views/SarosView.java
```diff
@@ -592,26 +614,29 @@ public static void showNotification(
           @Override
           public void run() {
 
-            if (control != null) {
-              BalloonNotification.showNotification(control, title, text);
+            Control attachToControl = control;
```
tobous: In my opinion, just using `control` directly here and creating the variable (previously name `sarosViewControl`) below is easier to read. It avoids re-assigning the value for the variable and should not make any functional difference as `control` is a final parameter. (Also, it would allow the call in l.620 to fit into a single line, but that is not really a valid argument.)

But I don't feel very strongly on this, so you can keep it this way if you prefer.

eclipse/src/saros/ui/views/SarosView.java
```diff
@@ -276,10 +280,20 @@ public SarosView() {
...
+    final GridLayout layout = new GridLayout(1, false);
+
+    layout.horizontalSpacing = 0;
+    layout.verticalSpacing = 0;
+
+    parent.setLayout(layout);
 
     final SashForm baseSashForm = new SashForm(parent, SWT.SMOOTH);
 
+    gridData = new GridData(SWT.FILL, SWT.FILL, true, true);
```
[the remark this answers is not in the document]
srossbach: It is declared on top of the method. It is used twice. Introducing a new variable name for every widget would be on overkill.

eclipse/src/saros/ui/views/SarosView.java
srossbach: The control is declared as final (I do not know if this still a must when using lambdas) so I cannot reassign any value to the control variable.

eclipse/src/saros/ui/views/SarosView.java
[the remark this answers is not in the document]
tobous: Ok, so you declared it explicitly in the beginning to show that it is used multiple times. Makes sense.

eclipse/src/saros/ui/views/SarosView.java
tobous: Yeah, what I meant was to just use the `control` variable here and declare the variable `attachToControl` (or whatever you want to call it) later (e.g. in line 631).

PR #622, opened 2019-08-05
[INTERNAL] Addresses issues introduced with PR #608

srossbach:
Includes several fixes for defects introduced by PR #608 

This PR also updates test cases that where failing because of the general changes made by PR #609


stf/test/saros/stf/test/stf/menubar/SarosPreferencesTest.java
```diff
@@ -124,14 +135,15 @@ public void editAccount() throws RemoteException {
         ALICE.superBot().menuBar().saros().preferences().isAccountActive(Constants.JID_TO_CHANGE));
   }
 
-  @Test(expected = RuntimeException.class)
   public void deleteActiveAccount() throws RemoteException {
```
m273d15: You removed the `@Test` and the method has not the prefix `test` and the method is not called. Do you want to remove the test?

srossbach: Seems I removed a little bit to much.

PR #623, opened 2019-08-06
[INTERNAL][E] Allows Saros to properly shutdown

srossbach:
When the stop method of the AbstractUIPlugin it is almost always the case that the UI thread is "not available" anymore because stop is called because the IDE shuts down.

This commit monitors the workbench and triggers a proper shutdown while the UI thread is still available.

In general this commit prevents mass warning logging when the user
simply clicks the close button of the Eclipse IDE.

This patch also adds some safeguards to AbstractContextLifecyle.

### DIFF
--- a/eclipse/src/saros/Saros.java
+++ b/eclipse/src/saros/Saros.java
@@ -3,12 +3,17 @@ package saros;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Properties;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.commons.io.IOUtils;
 import org.apache.log4j.Logger;
 import org.apache.log4j.PropertyConfigurator;
 import org.apache.log4j.helpers.LogLog;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.preferences.ConfigurationScope;
+import org.eclipse.swt.SWTException;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.ui.IWorkbench;
+import org.eclipse.ui.IWorkbenchListener;
 import org.eclipse.ui.plugin.AbstractUIPlugin;
 import org.osgi.framework.BundleContext;
 import org.osgi.service.prefs.BackingStoreException;
@@ -41,9 +46,24 @@ public class Saros extends AbstractUIPlugin {
 
   private static final String VERSION_COMPATIBILITY_PROPERTY_FILE = "version.comp"; // $NON-NLS-1$
 
+  private final IWorkbenchListener workbenchShutdownListener =
+      new IWorkbenchListener() {
+
+        @Override
+        public void postShutdown(final IWorkbench workbench) {
+          stopLifeCycle();
+        }
+
+        @Override
+        public boolean preShutdown(final IWorkbench workbench, final boolean forced) {
+          return true;
+        }
+      };
+
   private static boolean isInitialized;
 
   private EclipsePluginLifecycle lifecycle;
+  private boolean isLifecycleStarted = false;
 
   /**
    * @JTourBusStop 2, Some Basics:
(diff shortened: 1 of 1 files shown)


eclipse/src/saros/Saros.java
```diff
@@ -200,6 +211,59 @@ public static boolean useHtmlGui() {
     return Platform.getBundle("saros.ui") != null && Boolean.getBoolean("saros.swtbrowser");
   }
 
+  private void shutdownPlugin(final boolean readAndDispatch) {
```
m273d15: Could you provide a javadoc and describe the effect of `readAndDispatch`?

eclipse/src/saros/Saros.java
```diff
@@ -200,6 +211,59 @@ public static boolean useHtmlGui() {
...
+            log,
+            () -> {
+              try {
+                lifecycle.stop();
+              } finally {
+                isShutdownComplete.set(true);
+
+                if (display != null) {
+                  try {
+                    display.wake();
```
m273d15: Is this `display.wake()` required in the `if` case?
The `readAndDispatch` loop is never entered and we don't have to wake the display? Am I wrong?

eclipse/src/saros/Saros.java
srossbach: I guess I will remove it and do a check in the method instead if the display is available.

eclipse/src/saros/Saros.java
```diff
@@ -41,9 +46,24 @@
...
+        @Override
+        public boolean preShutdown(final IWorkbench workbench, final boolean forced) {
+          return true;
+        }
+      };
+
   private static boolean isInitialized;
 
   private EclipsePluginLifecycle lifecycle;
+  private boolean isLifecycleStarted = false;
```
srossbach: Is only written but never read.

eclipse/src/saros/Saros.java
```diff
@@ -200,6 +212,67 @@ public static boolean useHtmlGui() {
...
+    if (display != null && !display.isDisposed()) {
+
+      display.timerExec(
+          threadTimeout,
+          () -> {
+            isTimeout.set(true);
+            display.wake();
+          });
+
+      while (isLifeCycleStopped.get() == false && isTimeout.get() == false) {
```
tobous: `AtomicBoolean.get()` returns a `boolean`, you don't have to compare against `false`. Or was this done to improve readability?
```suggestion
      while (!isLifeCycleStopped.get() && !isTimeout.get()) {
```

eclipse/src/saros/Saros.java
```diff
@@ -41,9 +46,24 @@
 
   private static final String VERSION_COMPATIBILITY_PROPERTY_FILE = "version.comp"; // $NON-NLS-1$
 
+  private final IWorkbenchListener workbenchShutdownListener =
```
tobous: I am not completely clear on the concepts in Eclipse. Is the "workbench" the loaded workspace for the current project? What is the difference between this stop notification and the method `stop(BundleContext)` below starting at line 152?

My guess would be that this is specific to the workbench and the method below is specific to the application. This would also mean that you can change the workbench without ending the application, which leads to my next question:

Is the plugin bound to the workbench? Otherwise, why do we want to tear down the whole context when the workbench is closed?

srossbach: There is only one workbench. Stop is called when the plugin should be stopped, e.g the plugin should be reloaded or is deinstalled, or the IDE shuts down.

However stop is normally called while the Display is already disposed. As we do not use Display#getDefault (which would create a new one) the result was a bunch of warnings in the log stating the xyz could not be executed.

tobous: Ok, but do we then really need both methods dealing with the teardown of the plugin?
Could it have negative side-effects calling the teardown method `stopLifeCycle()` multiple times?

srossbach: stopLifeCycle can now be called multiple times, will only shutdown once.

PR #624, opened 2019-08-06
End session when project of shared module is closed

### LINKED ISSUE #582: Saros/I does not end the running session when the shared project is closed
Saros/I does not end the running session when the shared project is closed. This leads to issues as it then tries to apply activities to the disposed project/its modules.

This issue can be solved by implementing a project-closed listener which ends the session if the shared project is closed. The only open issue with this approach is how to synchronously end the session or filter remaining activities to guarantee that no activities are applied to disposed resources. This was further discussed in #539 and #568.

tobous:
#### [FEATURE][CORE] Add logic to report projects as disposed

Adds the logic to report a project as disposed to the session. This
causes all activities for such projects to be dropped.

This functionality is useful to prevent the session from trying to apply
activities to disposed (closed) projects.

As this logic is currently only used by Saros/I, it is not set up to
handle disposed partial-sharing resources. If this would be come useful
at some point, the logic should be adjusted to also hold a mapping for
disposed partially-shared resources.

#### [INTERNAL][I] Explicitly catch and log EDT interruption exceptions

Adjusts IntelliJSynchronizer to catch and log ProcessCanceledExceptions
which are thrown when a synchronous execution on the EDT is interrupted.

#### [FIX][I] #582 End session when project of shared module is closed

Adds a listener to end the session when the project containing the
shared module is closed. The listener also reports the modules of the
project as disposed to the session to prevent it from trying to apply
any already queued activities for the disposed modules.

Fixes #582.

### DIFF
--- a/core/src/saros/session/ISarosSession.java
+++ b/core/src/saros/session/ISarosSession.java
@@ -152,6 +152,23 @@ public interface ISarosSession {
   public void removeListener(ISessionListener listener);
 
   /**
+   * Enables or disables the execution of received activities for the given project. If the
+   * execution is disabled, activities for resources of the given projects will be dropped without
+   * being applied.
+   *
+   * <p>This method can be used to disable the execution of received activities for projects that
+   * are no longer available (i.e. are no longer part of the session or are no longer
+   * present/accessible locally).
+   *
+   * @param project the shared project to enable or disable the activity execution for
+   * @param enabled <code>true</code> to enable or <code>false</code> to disable the activity
+   *     execution
+   */
+  default void setActivityExecution(IProject project, boolean enabled) {
+    // NOP
+  }
+
+  /**
    * @return the shared projects associated with this session, never <code>null</code> but may be
    *     empty
    */
--- a/core/src/saros/session/internal/SarosSession.java
+++ b/core/src/saros/session/internal/SarosSession.java
@@ -29,6 +29,7 @@ import java.util.Set;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.CopyOnWriteArraySet;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.log4j.Logger;
 import saros.activities.FileActivity;
@@ -38,6 +39,7 @@ import saros.activities.IActivity;
 import saros.activities.IFileSystemModificationActivity;
 import saros.activities.IResourceActivity;
 import saros.activities.NOPActivity;
+import saros.activities.SPath;
 import saros.communication.extensions.KickUserExtension;
 import saros.communication.extensions.LeaveSessionExtension;
 import saros.concurrent.management.ConcurrentDocumentClient;
@@ -98,6 +100,8 @@ public final class SarosSession implements ISarosSession {
 
   private final CopyOnWriteArrayList<IActivityProducer> activityProducers =
       new CopyOnWriteArrayList<IActivityProducer>();
+
+  private final Set<IProject> filteredProjects = new CopyOnWriteArraySet<>();
 
   private final List<IActivityConsumer> activeActivityConsumers =
       new CopyOnWriteArrayList<IActivityConsumer>();
@@ -164,6 +168,17 @@ public final class SarosSession implements ISarosSession {
 
         @Override
         public void execute(IActivity activity) {
+          // Filters out resource activities for projects whose activity execution is disabled
+          if (activity instanceof IResourceActivity) {
+            SPath path = ((IResourceActivity) activity).getPath();
+
+            if (path != null && filteredProjects.contains(path.getProject())) {
+              log.debug("Dropped activity for resource of filtered project: " + activity);
+
+              return;
+            }
+          }
+
           /**
            * @JTourBusStop 10, Activity sending, Local Execution, first dispatch:
            *
@@ -557,6 +572,17 @@ public final class SarosSession implements ISarosSession {
   @Override
   public void removeListener(ISessionListener listener) {
     listenerDispatch.remove(listener);
+  }
+
+  @Override
+  public void setActivityExecution(IProject project, boolean enabled) {
+    if (project != null) {
+      if (!enabled) {
+        filteredProjects.add(project);
+      } else {
+        filteredProjects.remove(project);
+      }
+    }
   }
 
   @Override
--- a/intellij/src/saros/intellij/runtime/IntelliJSynchronizer.java
+++ b/intellij/src/saros/intellij/runtime/IntelliJSynchronizer.java
@@ -3,6 +3,8 @@ package saros.intellij.runtime;
 import com.intellij.openapi.application.Application;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.ModalityState;
+import com.intellij.openapi.progress.ProcessCanceledException;
+import org.apache.log4j.Logger;
 import saros.synchronize.UISynchronizer;
 
 /**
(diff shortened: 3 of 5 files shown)


core/src/saros/session/ISarosSession.java
```diff
@@ -151,6 +151,16 @@ public void changePermission(User user, Permission permission)
    */
   public void removeListener(ISessionListener listener);
 
+  /**
+   * Notifies the session that a shared project is now disposed or soon will be disposed. This can
+   * be used to tell the session to drop activities for no longer open projects.
+   *
+   * @param project the (disposed) shared project
+   */
+  default void disposeProject(IProject project) {
```
srossbach: What happens if you just remove the project from the sharing ? I guess there will be inconsistency warnings and I think there is also no broadcast.

srossbach: On the other hand as you are going to shutdown the session anyway, why do you not simply enable queuing ?

intellij/src/saros/intellij/runtime/IntelliJSynchronizer.java
```diff
@@ -32,7 +36,11 @@ private void exec(Runnable runnable, boolean async) {
     if (async) {
       application.invokeLater(runnable, ModalityState.defaultModalityState());
     } else {
-      application.invokeAndWait(runnable, ModalityState.defaultModalityState());
+      try {
+        application.invokeAndWait(runnable, ModalityState.defaultModalityState());
+      } catch (ProcessCanceledException e) {
+        log.error("Synchronous execution on EDT interrupted - " + runnable, e);
```
srossbach: What is interrupted ? The waiting for the execution to complete or the execution itself ?

core/src/saros/session/ISarosSession.java
srossbach: Ok nevermind, at least queuing will not work as it does not affect the queue in the ActivityHandler.

I would still prefer to change the name of this method. In general you do not want to dispose projects but initiate a half ordered shutdown (discarding pending activities). 

Do you have a good name proposal ?

intellij/src/saros/intellij/runtime/IntelliJSynchronizer.java
tobous: The [documentation of the `incokeAndWait(...)`](https://github.com/JetBrains/intellij-community/blob/master/platform/core-api/src/com/intellij/openapi/application/Application.java#L277) call says the following:
```Java
@throws ProcessCanceledException when the current thread is interrupted
```
So I...I guess both? Its not really specific...

core/src/saros/session/ISarosSession.java
tobous: How about something along the lines of `notifyProjectDisposal` or `addActivityFilterForProject`?

As you have already mentioned, this does not have to mean that the session ends. Closing projects could could also be handled by removing the closed projects from the session.
But this should be handled by other parts of the implementation. This is only a workaround to stop the session from trying to apply activities to resources that are no longer available.

tobous: Changed the wording to reflect the filter nature.

core/src/saros/session/internal/SarosSession.java
```diff
@@ -164,6 +167,19 @@ public void send(List<User> recipients, IActivity activity) {
 
         @Override
         public void execute(IActivity activity) {
+          /*
+           * Filters out resource activities for disposed projects
```
srossbach: What are disposed projects when translating this statement to Core context ?

core/src/saros/session/internal/SarosSession.java
```diff
@@ -99,6 +100,8 @@
   private final CopyOnWriteArrayList<IActivityProducer> activityProducers =
       new CopyOnWriteArrayList<IActivityProducer>();
 
+  private final Set<IProject> filteredProjects = new HashSet<>();
```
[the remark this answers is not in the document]
tobous: Done.

core/src/saros/session/internal/SarosSession.java
tobous: My bad. Missed that when removing the "disposed" references from the core stuff.

core/src/saros/session/ISarosSession.java
```diff
@@ -151,6 +151,20 @@ public void changePermission(User user, Permission permission)
...
+   * Adds the given project to the list of held projects to filter activities for. This causes the
+   * session to drop activities for such projects and their resources instead of trying to apply
+   * them locally.
+   *
+   * <p>This method should be used to add filters for projects that are no longer available (i.e.
+   * are no longer part of the session or are no longer present/accessible locally).
+   *
+   * @param project the shared project to add a filter for
+   */
+  default void addActivityFilterForProject(IProject project) {
```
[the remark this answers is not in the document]
tobous: Done. Changed to `setActivityExecution(IProject project, boolean enabled)`.

tobous:
Addressed review comments. And yay, I amended the wrong commit. Will fix this and push again.

Also, this PR currently does not address the teardown issues when an activity execution is already queued (throwing a `ProcessCanceledException). I still don't really know what to do about this. But as this is not bound to this PR, I would like to deal with it separately.

srossbach:
As I told you. Use a POISON pill. This should work after this patch.

tobous:
> As I told you. Use a POISON pill. This should work after this patch.

Yes, I got that (and the basic concept explained in the linked article, but I still don't quite know what exactly to do. But I will have a more thorough look at it and will get back to you if I still don't get it. :relaxed:

srossbach:
I am almost done, will push it in a few minutes.

tobous:
That works as well. Thanks. :relaxed:

tobous:
Moved fixes for review comments to correct commit.

core/src/saros/session/ISarosSession.java
```diff
@@ -151,6 +151,21 @@ public void changePermission(User user, Permission permission)
    */
   public void removeListener(ISessionListener listener);
 
+  /**
+   * Enables or disables the activity execution for the given project. If the execution is disabled,
```
srossbach: A bit misleading as one might assume it works for activities applied to this project and generated by this project. Should be fine though for the current issue.

core/src/saros/session/internal/SarosSession.java
```diff
@@ -164,6 +168,17 @@ public void send(List<User> recipients, IActivity activity) {
 
         @Override
         public void execute(IActivity activity) {
+          // Filters out resource activities for projects whose activity execution is disabled
+          if (activity instanceof IResourceActivity) {
+            SPath path = ((IResourceActivity) activity).getPath();
+
+            if (path != null && filteredProjects.contains(path.getProject())) {
```
srossbach: As far as I know there are EditorActivities with path = null to indicate an unknown editor.

tobous: I only know of EditorActivities with `path=null` to indicate that no shared editor is selected/open.

But either way, if the path is `null`, we have no way of telling which project it belongs to, meaning we also can't apply it to a project. This would only be a problem if we have some activities that are applied to some "default project" such a case.

 But I will look through the `IResourceActivity` implementations to make sure.

tobous: *Currently*, the only other resource activity besides `EditorActivity` that allows a `null` path is `JupiterActivity`, so it should not be a problem.

I am considering adding a comment to `AbstractResourceActivity` to avoid this being overlooked when adding new resource activities. But I am not sure whether it is really worth is.

core/src/saros/session/ISarosSession.java
tobous: Adjusted to better reflect that this only applies to received activities.

tobous:
Rebased onto current master and amended javadoc for `setActivityExecution`.

PR #625, opened 2019-08-06
[JUNIT][E] Remove Smack Roster from Test

stefaus:
To start removing Smack from UI packages, remove it from test classes.


srossbach:
And how do you want to remove the roster from the RosterEntryElement class ? I guess for this test case it is ok to use mocks.

stefaus:
> And how do you want to remove the roster from the RosterEntryElement class ? I guess for this test case it is ok to use mocks.

This will come in another bigger PR... I prepared something ;)

PR #626, opened 2019-08-06
[CORE] Add Subscription Canceled Listener

stefaus:
Add the possibility to listen to subscription canceled events.

### DIFF
--- a/core/src/saros/net/xmpp/subscription/SubscriptionListener.java
+++ b/core/src/saros/net/xmpp/subscription/SubscriptionListener.java
@@ -9,5 +9,16 @@ public interface SubscriptionListener {
    *
    * @param jid the {@linkplain JID} of the contact who requested subscription
    */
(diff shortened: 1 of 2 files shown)


core/src/saros/net/xmpp/subscription/SubscriptionListener.java
```diff
@@ -9,5 +9,16 @@
...
+  public default void subscriptionRequestReceived(JID jid) {
+    // NOP
+  }
+
+  /**
+   * Gets called whenever an subscription cancel was received.
+   *
+   * @param jid the {@linkplain JID} of the contact who canceled the subscription
+   */
+  public default void subscribtionCanceled(JID jid) {
```
srossbach: P not B

stefaus: changed :+1:

PR #629, opened 2019-08-06
[WIP][E] Clean up code of Actions

srossbach:
The code for some actions is very old and can be cleaned up to be more
understandable.


stefaus:
@srossbach would you like to resolve the conflicts and merge it?

PR #632, opened 2019-08-08
Finish color page patch proposed in #482

tobous:
Continuation of #482. The original PR contributed by @davidschreiber was an external PR (i.e. based on a fork). The commits of the initial PR were rebased and squashed and are contained in the first commit of this PR.

#### [FEATURE][I] Add customizable user highlight colors

This commit replaces the previously hard-coded highlight colors inside
ColorManager with user-customizable TextAttributes. It adds IntelliJ
TextColorAttributes for every of the customizable colors which are
'selection' and 'contribution' colors for users 0-4 and a separate set
of highlight colors for the 'default' user (i.e. if no user could be
matched).

This commit also adds a color preference screen to the IDE, so that the
color attributes can be modified inside  Preferences -> Editor -> Color
Scheme -> Saros. This commit adds two default color schemes
configurations for the IntelliJ "Default" and "Darcula" schemes.

#### [REFACTOR][I] Fix order of color entries

Adjust order of color entries in config file to match default Saros
color layout.

#### [INTERNAL][I] Move demo text generation to separate method in color page

Creates a separate method creating the demo text displayed in the Saros
color settings page for the given color id. This was done to avoid code
duplication and make it easier to add color settings for other
annotations types.

Subsequently splits the messages.properties entry for the demo text into
separate values for different annotation types.

Also adjusts the demo text to state which annotation type is shown for
which user.

#### [REFACTOR][I] Move color key names to constants

Moves the color key (prefix) values used to retrieve the colors set in
the settings file to static constants. Also adds javadoc to the
constants mentioning that the keys need to match the option names used
in the config files.

#### [REFACTOR][I] Refactor color manager and color selection page

Reorders the static variables in ColorManager so that the static
variable COLOR_KEYS is declared last so that the static block
initializing the variable can directly follow the declaration.

Completes the javadoc of DEFAULT_COLOR_KEYS in ColorManager.

Moves the static initialization of ADDITIONAL_ATTRIBUTE_DESCRIPTORS in
SarosColorPage to a separate static block in order to match the
initialization of the other static variables. Otherwise, it is kind of
confusing that one of the variables is initialized inline while the
surrounding variables use a separate static block.

Adds a javadoc to HIGHLIGHT_MAP in SarosColorPages.

#### [REFACTOR][I] Replace use of google immutable collections in color logic

Replaces the usage of the google immutable collection classes in
ColorManager and SarosColorsPage with the usage of unmodifiable
collections. As we don't hold any references to the internal
collections, unmodifiable should be the same as immutable in this case.

This was done to avoid the unnecessary use of a new collections library.
 
#### [REFACTOR][I] Rename SarosColorsPage to LocalSarosAnnotationColorsPage


intellij/src/saros/intellij/editor/colorstorage/ColorManager.java
```diff
@@ -1,44 +1,168 @@
...
+  /** Color keys for the default colors. Used if no color keys are available for a given id. */
+  public static final ColorKeys DEFAULT_COLOR_KEYS = new DefaultColorKeys();
+  /**
+   * List of keys for supported colors. A user can select one of these colors by referencing the
+   * {@link IdentifiableColorKeys#getId()}.
+   */
+  public static final List<IdentifiableColorKeys> COLOR_KEYS;
+
+  static {
+    final ImmutableList.Builder<IdentifiableColorKeys> builder = ImmutableList.builder();
```
srossbach: Do you use Google Collection at all ? I personally do not find it a good behavior to use multiple collection frameworks.

intellij/src/saros/intellij/preferences/colors/SarosColorsPage.java
```diff
@@ -0,0 +1,173 @@
...
+import javax.swing.Icon;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import saros.intellij.editor.colorstorage.ColorManager;
+import saros.intellij.editor.colorstorage.ColorManager.IdentifiableColorKeys;
+import saros.intellij.ui.Messages;
+import saros.intellij.ui.util.IconManager;
+
+/** IntelliJ color scheme preferences for Saros specific colors. */
+public class SarosColorsPage implements ColorSettingsPage {
```
srossbach: I just want to remember you that there is an option to select your color for a session. Maybe use a better fitting name.

intellij/src/saros/intellij/editor/colorstorage/ColorManager.java
tobous: I did not write this part of the code. It is part of the code contributed in #482. But, it did not bother me enough to change it.

tobous:
@m273d15 Do you understand why Codacy insists on every class being abstract (e.g. see [codacy build of this PR](https://app.codacy.com/app/Saros/saros/pullRequest?prid=3979882&bid=13792314))? This seems to be a recent change in behavior, but it is really weird.

intellij/src/saros/intellij/preferences/colors/SarosColorsPage.java
tobous: How about `LocalSarosAnnotationColorsPage`?

I am assuming that you are hinting at the future need for a settings page where you can set your preferred color (that other see your annotations in), which could be confused with this page. If this is not the case, please explain what you mean.

srossbach: You guessed it right.

tobous: Good. Any issues with the suggested name?

intellij/src/saros/intellij/editor/colorstorage/ColorManager.java
tobous: I could replace this by just using normal collections (lists, maps) and then using `Collections.unmodifiableList(...)` or `Collections.unmodifiableMap(...)` to create the same outcome. (Even though unmodifiable is not the same as immutable in general, it should hold in this case as there is no reference to the internal representation of the unmodifiable collection, which *should* mean that it can't be modified by anything.)

@srossbach Would this be a preferable solution?

intellij/src/saros/intellij/preferences/colors/SarosColorsPage.java
srossbach: Nope. Can you please post a picture what this page looks like ? Maybe we can introduce that too in Saros/E.

intellij/src/saros/intellij/editor/colorstorage/ColorManager.java
srossbach: If it does not consume too much time to change it, why not. Have seen the usage also in another class.

tobous: Yes, in `SarosColorsPage`. Should only take ~5 minutes. Will do.

intellij/src/saros/intellij/preferences/colors/SarosColorsPage.java
tobous: There already is a picture of this page in the initial PR #482 (I also linked it below). The page did not change much except that I changed the text to state which annotation for which "user" (color id) is shown.

Preferences

We could add a separate page for this, but these settings are already available in Saros/E under "General" > "Editors" > "Text Editors" > "Annotations", which should have multiple entries starting with "DPP".

I just noticed that we also have to rename these. I will open an issue for this.

srossbach: Nice. The Eclipse stuff is pretty poor and cumbersome to use and currently changing the values is not the best idea as we are replacing them on every IDE start with the default values. 😈 

image

m273d15:
@tobous No, It seems like they updated their checkstyle version [a few days ago](https://github.com/codacy/codacy-checkstyle/commits/master), but I don't know whether this is related. The UI still shows version 8.22. I will try to find the reason and otherwise disable the rule (which is currently our only enabled rule.) as temporary solution.

tobous:
@m273d15 I disabled the rule for now. It is really weird that it is the only enabled rule for checkstyle.

intellij/src/saros/intellij/editor/colorstorage/ColorManager.java
tobous: Done.

tobous:

PR #633, opened 2019-08-08
Reload annotations on color scheme change

tobous:
#### [REFACTOR][I] Adjust method to add local annotation representation

Adjusts AnnotationManager.addLocalRepresentationToAnnotations to accept
only a single annotations, thereby moving the process of iterating over
the list of annotations to the caller. This was done to make the method
compatible with an upcoming use case where it is needed for single
annotations, avoiding the need to wrap single elements in a list.

#### [INTERNAL][I] Reload annotations when IDE color scheme changes

Adds AnnotationReloader to reload all held annotations when the IDE
color scheme changes. This is done to ensure that all currently
displayed Saros annotations use the correct colors for the newly
selected IDE color scheme.


intellij/src/saros/intellij/eventhandler/colorscheme/AnnotationReloader.java
```diff
@@ -0,0 +1,81 @@
...
+    this.enabled = false;
+    this.disposed = false;
+  }
+
+  /**
+   * Reloads all current Saros annotations.
+   *
+   * @see AnnotationManager#reloadAnnotations()
+   */
+  private void reloadAnnotations() {
```
srossbach: Where is this called ? Cannot find any reference in this class.

tobous: In line 23. Its a bit hidden with the lambda stuff.

intellij/src/saros/intellij/eventhandler/colorscheme/AnnotationReloader.java
```diff
@@ -0,0 +1,81 @@
...
+    setEnabled(false);
+  }
+
+  @Override
+  public void setEnabled(boolean enabled) {
+    assert !disposed || !enabled : "disposed listeners must not be enabled";
+
+    if (!this.enabled && enabled) {
+      messageBusConnection = project.getMessageBus().connect();
+      messageBusConnection.subscribe(EditorColorsManager.TOPIC, editorColorsListener);
```
srossbach: Ok, I thought the interface that this class implements is IntelliJ specific and that you subscribe to the bus with this instance and so was looking for the callback.

PR #635, opened 2019-08-09
[INTERNAL][CORE] Stop ActivityHandler more gracefully

srossbach:
It is now stopped with a poison pill to avoid some issues in some
UISynchronizer implementations.

### DIFF
--- a/core/src/saros/session/internal/ActivityHandler.java
+++ b/core/src/saros/session/internal/ActivityHandler.java
@@ -99,7 +99,7 @@ public final class ActivityHandler implements Startable {
             dispatchQueue.drainTo(pendingActivities);
 
             for (final List<IActivity> activities : pendingActivities) {
-              if (activities == POISON_PILL) {
+              if (activities == POISON_PILL) { // NOPMD - object reference comparison needed
                 isPoisoned = true;
                 break;
               }
--- a/core/src/saros/session/internal/ActivityHandler.java
+++ b/core/src/saros/session/internal/ActivityHandler.java
@@ -34,6 +34,8 @@ public final class ActivityHandler implements Startable {
 
   private static final Logger LOG = Logger.getLogger(ActivityHandler.class);
 
+  private static final List<IActivity> POISON_PILL = new ArrayList<>();
+
   /** join timeout when stopping this component */
   private static final long TIMEOUT = 10000;
 
@@ -82,7 +84,9 @@ public final class ActivityHandler implements Startable {
         public void run() {
           LOG.debug("activity dispatcher started");
 
-          while (!Thread.currentThread().isInterrupted()) {
+          boolean isPoisoned = false;
+
+          while (!Thread.currentThread().isInterrupted() && !isPoisoned) {
             pendingActivities.clear();
             activitiesToExecute.clear();
 
@@ -94,8 +98,14 @@ public final class ActivityHandler implements Startable {
 
             dispatchQueue.drainTo(pendingActivities);
 
-            for (final List<IActivity> activities : pendingActivities)
+            for (final List<IActivity> activities : pendingActivities) {
+              if (activities == POISON_PILL) {
+                isPoisoned = true;
+                break;
+              }
+
               activitiesToExecute.addAll(activities);
+            }
 
             dispatchAndExecuteActivities(activitiesToExecute);
           }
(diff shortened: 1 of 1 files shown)


core/src/saros/session/internal/ActivityHandler.java
```diff
@@ -94,8 +98,14 @@ public void run() {
 
             dispatchQueue.drainTo(pendingActivities);
 
-            for (final List<IActivity> activities : pendingActivities)
+            for (final List<IActivity> activities : pendingActivities) {
+              if (activities == POISON_PILL) {
```
tobous: As you actually want/need to use the object reference comparison here, how about adding a comment to disable the PMD ([see PMD documentation](https://pmd.github.io/latest/pmd_userdocs_suppressing_warnings.html#nopmd-comment)) check? This would avoid Codacy complaining and would serve as a comment for others reading the code.

```suggestion
              if (activities == POISON_PILL) { // NOPMD - object reference comparison needed
```

tobous:
@srossbach Ok, I am still a bit confused by the initial issue I found that led to this patch.

I now get why you used the poison pill approach. It allows for a more graceful teardown as it wakes up the executor thread waiting on `dispatchQueue.take()` and tells it to end the run-loop.

But why did this issue only occur in my testing when I just created an activity locally (i.e. typed something in a shared activity and then immediately closed the project; this lead to an issue in the Saros instance where I closed the project)?
To me, this sounded more like an issue with the execution of the local callback being interrupted, which this should not have any influence on, shouldn't it? Or am I missing something?

srossbach:
Didnt you get the problem when call stop from EDT, you can try it  out again, you will now get a timeout exception instead.

tobous:
Yes, that was the initial issue. But that also got solved by spawning a new thread to call stop (as the listener reacting to the project close is notified on the EDT).

But, even with this change I was still able to reproduce the issue mentioned in https://github.com/saros-project/saros/issues/582#issuecomment-515048625. So far, I have not been able to reproduce the issue with this patch (in combination with the PR to end the session). But my test method is not consistent, so I can't be completely sure that it isn't still happening (at least not from my testing).

srossbach:
When you close the project, will a modal dialog open up ? If yes every pending activity cannot be executed, when you call stop it is likely that interrupt is called while this dialog is still shown / EDT execution is still busy, causing them pending syncExec call to gets interrupted.

tobous:
> When you close the project, will a modal dialog open up ? If yes every pending activity cannot be executed, when you call stop it is likely that interrupt is called while this dialog is still shown / EDT execution is still busy, causing them pending syncExec call to gets interrupted.

Nope. The project just closes. IntelliJ just asks for confirmation if you close the last project of the application as this would also shut down the whole application (but this dialog can also be disabled).
Furthermore, I am not shutting down the IDE through any options but by using the "kill window" shortcut of my window manager. I have no idea how this is handled internally/by IntelliJ.

Also, I can't be sure that the issue is really caused by me closing the project after creating local activities. It could also be caused by unfortunately timed ChecksumActivities received from the host (as I am trying this on the client side). This is something I haven't considered so far. All in all, my testing methodology is not good enough to reliably pinpoint a reason for the issue.

tobous:
Ok, from some more testing, it seems like this is caused by incoming activities and not local actions. (Without this patch) I can force the issue by just setting the checksum timeout to something ridiculously small like 10, which leads to the issue consistently occuring when just starting a session and then closing the shared project on the client side while having other projects open (i.e. not also ending the application).

But, the same setup no longer causes the issue with this patch.

This is also not solely caused by the mass of activities as I was also able to recreate it (again, without this patch) while having the checksum activities "disabled" (interval set to max long, meaning a period of ~292 million years). I had the debugger running and the activity handler was executing a single viewport activity from the host.

I will do some more testing.

tobous:
Was not able to recreate it with this patch, independently of the checksum activity period, whether I was quitting as the client or the host, or what I did before quitting.

I am giving up for now but will have another look in the debugger at the failure state (without the patch) in the next couple of days. Either way, I think this PR can proceed with this PR independently.

Other than the suggested comment, I have no issues with this PR.

srossbach:
Debugging a multithreaded application with a debugger ... have not that since years. Adding some Logging statements is most of the time more useful. Will add the PMD command.

srossbach:
@tobous do you know how the commit message looks like when I use Squash and Merge ? Will it be generated from the description of this PR ?

tobous:
> Debugging a multithreaded application with a debugger ... have not that since years. Adding some Logging statements is most of the time more useful. Will add the PMD command.

I still like to use it if I am not completely certain what I am looking for. It allows me to have a look around in the internal state.

tobous:
> @tobous do you know how the commit message looks like when I use Squash and Merge ? Will it be generated from the description of this PR ?

No. The UI will open a text-box that contains all commit messages (not the PR text). You can then still modify them.

Issue #634, opened 2019-08-08
Rename Saros/E annotation color settings


tobous:
Saros/E offers settings to customize the colors that annotations are shown in.
The settings are located in the settings under "General" > "Editors" > "Text Editors" > "Annotations".

The names of the different entries are outdated, i.e. they still start with "DPP". This should be changed to "Saros" to avoid confusion as the shorthand DPP is no longer used by the project.

srossbach:
The settings are automatically provided by Eclipse. We do not need to write any line of code. Since the settings you made there will be overwritten by us every time on plugin start I recommend to not showing them at all (which can be easily done be settings a key for every annotation in the plugin.xml to false).

Any reasons for showing them ?

tobous:
Well it depends on whether we replace them or not. I still think it is beneficial to allow the users to customize the used colors. I was not aware that the settings made there are dropped every time the IDE is restarted.

If the settings are replaced by an alternative settings page (which you already suggested in #632), these options can probably be just removed if they are completely obsoleted. If, however, there are still some options in the menu that we won't include in our settings page, it could still be useful to offer such options, even if just in a temporary fashion.

srossbach:
You got me wrong. As our plugin resets the values (not Eclipse) it is currently the best option to hide them. We also have annotations that will be never shown there because this annotations are managed by us (to be more specific we draw them on the widgets). 

https://github.com/saros-project/saros/blob/master/eclipse/src/saros/editor/annotations/SelectionFillUpStrategy.java

tobous:
Hmmm. That is indeed a problem.

But I am still not blown away by removing a customization option (even if it is not *that* usable) without offering a replacement.

srossbach:
![image](https://user-images.githubusercontent.com/41400679/62782618-e9c20200-baba-11e9-8b06-f61fd1d2d087.png)

Using this mechanism is not great for picking colors. Especially setting 15 of them.

srossbach:
And IIRC there was once an option to set colors (however it was coded so badly that it was also setting the color on the remote side). It got removed, using only a default color set for everyone.

tobous:
Using default sets would certainly help.

But, with eclipse also supporting (custom) color themes, it would be nigh impossible to provide an option that is usable with every color scheme. So I still think that allowing the user to customize the colors is important.

srossbach:
[...]

PR #636, opened 2019-08-12
[UI][E] Custom colors for annotations.

srossbach:
This patch allows customization of the annotation colors. Colors or now fetched up from the store and the stored values are no longer overwritten.


eclipse/src/saros/editor/annotations/SarosAnnotation.java
```diff
@@ -73,31 +78,46 @@ public static Color getUserColor(int colorID) {
   }
 
   /**
-   * Returns the color that corresponds to a user.
+   * Returns the color for the given annotation type and user color id.
    *
    * <p><b>Important notice:</b> Every returned color instance allocates OS resources that need to
    * be disposed with {@link Color#dispose()}!
    *
    * @param type
```
tobous: Complete javadoc for params now that method is public?

eclipse/src/saros/editor/annotations/SarosAnnotation.java
```diff
@@ -73,31 +78,46 @@ public static Color getUserColor(int colorID) {
...
+    if (rgb == null)
+      throw new IllegalArgumentException(
+          "annotation '" + typeToLookup + "' does not have a default color");
+
+    return new Color(Display.getDefault(), rgb);
   }
 
-  /** Loads the colors from the plugin.xml and overwrites possible errors in the PreferenceStore */
+  /**
+   * Loads the colors from the plugin.xml and overwrites possible errors in the EditorsUI preference
```
tobous:
```suggestion
   * Loads the default Saros colors from the plugin.xml and overwrites possible errors in the EditorsUI preference
```

eclipse/src/saros/editor/annotations/SarosAnnotation.java
```diff
@@ -73,31 +78,46 @@ public static Color getUserColor(int colorID) {
...
+   * store.
+   */
   public static void resetColors() {
```
tobous: IntelliJ displays that this is now no longer used. If so, can this be removed or will it be used again in an upcoming patch?

eclipse/src/saros/editor/annotations/SarosAnnotation.java
```diff
@@ -73,31 +78,46 @@ public static Color getUserColor(int colorID) {
...
+        PreferenceConverter.getColor(EditorsUI.getPreferenceStore(), ap.getColorPreferenceKey());
 
-    return new Color(Display.getDefault(), ap.getColorPreferenceValue());
+    if (rgb == PreferenceConverter.COLOR_DEFAULT_DEFAULT) { // NOPMD
+      rgb = ap.getColorPreferenceValue();
+    }
+
+    if (rgb == null)
+      throw new IllegalArgumentException(
+          "annotation '" + typeToLookup + "' does not have a default color");
```
tobous: How about stating that we encountered an unknown annotation color type (under the assumption that every valid type has a default value)?

srossbach:
I do not plan to implement a page for customizing. If someone has to much time to spend, feel free to do so. 

As for the naming, it all can be found in the mighty plugin.xml as usual. So how should we name them ? They will be sorted and so the question is.

Should this look like:

User 1 Contr
User 1 High
User 1 View

User 2 Contr
User 2 High
User 2 View

or

User 1 Contr
User 2 Contr
User 3 Contr
......
User 1 High
User 2 High
User 3 High

eclipse/src/saros/editor/annotations/SarosAnnotation.java
srossbach: TBH adding Javadoc will not help you. The type numbers are hardcoded in the plugin.xml file and also hardcoded into the code.

eclipse/src/saros/editor/annotations/SarosAnnotation.java
srossbach: if you call this with ("foo.bar", 4711) it will be translated to foo.bar.Default

I could also throw "you fucked up the plugin.xml". I think this is more helpful.

It is just a look up of

      <specification
            annotationType="saros.annotations.viewport.default"
            colorPreferenceKey="saros.annotations.viewport.color.default"
            colorPreferenceValue="128,128,128"
            contributesToHeader="true"
            highlightPreferenceKey="saros.annotations.viewport.highlight.default"
            highlightPreferenceValue="false"
            includeOnPreferencePage="true"
            label="DPP Default Viewport"
            overviewRulerPreferenceKey="saros.annotations.viewport.overview.default"
            overviewRulerPreferenceValue="true"
            presentationLayer="1"
            textPreferenceKey="saros.annotations.viewport.text.default"
            textPreferenceValue="false"
            textStylePreferenceValue="NONE"
            verticalRulerPreferenceKey="saros.annotations.viewport.ruler.default"
            verticalRulerPreferenceValue="false"/>
      <specification

if you do add a ColorPreferenceValue this exception will be thrown

eclipse/src/saros/editor/annotations/SarosAnnotation.java
tobous: Yes, but you could point out that `Annotation.getType()` (of the appropriate annotation implementation) should be used to obtain the type as it will give you the only valid options.

eclipse/src/saros/editor/annotations/SarosAnnotation.java
srossbach: Smurfing code with Saros is not always the best option. It should be clear that this is the plugin.xml of the Saros plugin. The class is already smurfed with Saros.

tobous:
I would sort them by user id as I would assume you would want to change all colors for one id. Having the options grouped together makes it easier to switch between different annotation types to ensure that the colors match/go together.

(I don't know if you just dropped it for conveniences sake in the listing, but every entry should start with "Saros", i.e. "Saros User 1 Contribution Color".)

eclipse/src/saros/editor/annotations/SarosAnnotation.java
srossbach: Logic could be useful with a little change to apply themes.

srossbach:
image

Do you think adding Color as suffix looks professional ?

eclipse/plugin.xml
```diff
@@ -138,7 +139,8 @@
             highlightPreferenceKey="saros.annotations.viewport.highlight.default"
             highlightPreferenceValue="false"
             includeOnPreferencePage="true"
-            label="DPP Default Viewport"
+            label="Saros Default User Viewport"
+            icon="icons/obj16/contact_saros_obj.png"
```
tobous: Is there a reason to use this icon? How about using the Saros icon instead to make the options easier to find?

srossbach: image

It is this icon, do you prefer the Saros Icon instead of the Saros User Icon ?

tobous: For me, the Saros User icon just reminds me of the old MSN contact icon. I don't particularly associate it with Saros.

I would use the Saros Icon, but if you think that the Sarus User Icon works as well/better, I don't mind using it instead.

srossbach: Ok changed to Saros Icon.

srossbach:
Just changed the icon.

srossbach:
#634

PR #637, opened 2019-08-12
[FIX][E] #19 Add Contact to session is enabled for offline contacts

### LINKED ISSUE #19: Add Buddies Context Menu Item is enabled for Offline Buddies
See screenshot. The Context Menu item is enabled for offline buddies or buddies without Saros support. If you invite them, you will get a dialog concerning that you cannot invite that buddy to a session.

Reported by: kargor

Original Ticket: [dpp/bugs/641](https://sourceforge.net/p/dpp/bugs/641)

srossbach:
This patch fixes #19. The option(s) to invite an offline contact is no
longer possible.

### DIFF
--- a/eclipse/plugin.xml
+++ b/eclipse/plugin.xml
@@ -1040,10 +1040,13 @@
                         definitionId="saros.ui.definitions.isSarosSessionRunning">
                   </reference>
                   <reference
+                        definitionId="saros.ui.definitions.isSarosSessionHost">
+                  </reference>
+                  <reference
                         definitionId="saros.ui.definitions.contactsSelected">
                   </reference>
                   <reference
-                        definitionId="saros.ui.definitions.isSarosSessionHost">
+                        definitionId="saros.ui.definitions.contactsOnline">
                   </reference>
                </and>
             </visibleWhen>
@@ -1188,6 +1191,28 @@
                type="saros.net.xmpp.JID">
          </adapt>
       </definition>
+      <definition
+      		id="saros.ui.definitions.isContactOnline">
+         <adapt
+               type="saros.net.xmpp.JID">
+            <test
+                  property="saros.isOnline">
+            </test>
+         </adapt>
+      </definition>
+            <definition
+            id="saros.ui.definitions.contactsOnline">
+         <with
+               variable="selection">
+            <iterate
+                  ifEmpty="false"
+                  operator="and">
+               <reference
+                     definitionId="saros.ui.definitions.isContactOnline">
+               </reference>
+            </iterate>
+         </with>
+      </definition>      
       <definition
             id="saros.ui.definitions.contactsSelected">
          <with
@@ -1380,7 +1405,7 @@
             class="saros.ui.expressions.ContactPropertyTester"
             id="saros.ui.expressions.ContactPropertyTester"
             namespace="saros"
-            properties="isInSarosSession"
+            properties="isInSarosSession,isOnline"
             type="saros.net.xmpp.JID">
       </propertyTester>
       <propertyTester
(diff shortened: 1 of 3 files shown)


eclipse/src/saros/ui/views/SarosView.java
```diff
@@ -456,7 +457,13 @@ public void menuAboutToShow(final IMenuManager manager) {
                 SelectionRetrieverFactory.getSelectionRetriever(JID.class).getSelection();
             if (contacts.size() == 0) return;
 
-            if (sarosSessionManager.getSession() == null) {
+            final JID jid = contacts.get(0);
+
+            // FIXME dirty hack
+            final ContactPropertyTester tester = new ContactPropertyTester();
```
tobous: I am not completely sure on how the tester should be used properly, but, as also indicated by the "dirty hack" fixme, this does not seem to be the correct way.
Are there any restrictions keeping us from using the tester correctly? (And how are they supposed to be used anyways.? The javadoc of `PropertyTester` doesn't really seem helpful to me...)

eclipse/plugin.xml
```diff
@@ -1039,11 +1039,14 @@
...
                   </reference>
+                  <reference
+                        definitionId="saros.ui.definitions.isSarosSessionHost">
+                  </reference>
                   <reference
                         definitionId="saros.ui.definitions.contactsSelected">
                   </reference>
                   <reference
-                        definitionId="saros.ui.definitions.isSarosSessionHost">
+                        definitionId="saros.ui.definitions.contactsOnline">
```
tobous: I am really not familiar with the Eclipse plugin.xml config stuff. Could you explain what these changes do?

You add all of this config stuff which makes it look like the online check should be done by eclipse, but then call the tester manually (see comment in `SarosView`). If it wasn't added for that purpose, why was it added?

eclipse/src/saros/ui/views/SarosView.java
srossbach: We got an issue with Eclipse 4. Maybe you look at my former commits. The entry that is displayed here was handle be the magic XML stuff in the plugin.xml. With Eclipse 4 it broke and so we have to do it manually.

eclipse/plugin.xml
srossbach: Seems the author of this incredible stuff has left the building long ago. Maybe you talk to @m273d15 . I can tell you his answer. It should be the same as mine: WTF is this ******* ....

It is really complicated, it is a waste of time, coding this stuff takes minutes, define it as XML takes hours with frustrating trail and errors.

eclipse/src/saros/ui/views/SarosView.java
srossbach: Better explanation. Suppose you have a JTree. If you right - click on on entry context menu will appear, basically the logic here is building the context menu. But with Eclipse it (was) is possible to even append this menu, e.g when you open the context menu for a tree item your own logic ads A,B,C and on top of that the logic defined in the plugin.xml will add D,E, F.

tobous: Ok, so the whole config stuff in the plugin.xml was also added in case this check ever were to function again in the future? Otherwise, if the logic is broken and unnecessarily complicated to use, why add the config for it?

srossbach: Are you aware that the logic in the plugin.xml adds the context menu item "Add contact to session" and the logic in the SarosView adds "Share Projects..." ?

eclipse/plugin.xml
srossbach: Basically it is: Add the command (menu item) "Add contacts to session" to the any context menu if the following conditions holds true.

You are connected, a session is running, you are the host, the current selection(s) are contacts and these contacts are currently online.

eclipse/src/saros/ui/views/SarosView.java
tobous: No. From your explanations, I just assumed that everything dealing with the plugin.xml menu stuff was broken. But if some stuff is still working, the question basically answers itself.

How about amending the FIXME addressing how/when to fix/replace the added hack?

srossbach:
I guess that can be replaced when the new XMPP stuff from Stefan is introduced.

Issue #61, opened 2014-10-15
Project names with trailing numbers are truncated in Wizard


dpp-gerrit:
* Share a project with a trailing number in its name, such as "ALP4".
* In the "Add Projects" dialog on the recipients side, the text box under the option "Create new project" is pre-filled with the project name without the number, i.e. "ALP" in the above mentioned example.

This should probably by fixed in <code>EnterProjectNamePage.findProjectNameProposal()</code>.
Better proposals would be "ALP4_1", "ALP4_2", ...

Reported by: fzieris

Original Ticket: [dpp/bugs/837](https://sourceforge.net/p/dpp/bugs/837)

PR #638, opened 2019-08-12
[FIX][E] #61 Project names with trailing numbers are truncated in Wizard

srossbach:
This patch fixes issue #61.

Names are now calculated as follow:

A
A (2)
A (3)

and so on.

In addition if you switch from an existing project to a non existing
project and the name is not set yet the logic insert on. 

Eg: Remote Project Name is A, mapping is B (existing). If you now select Create New Project the text field will be filled with a valid non existing project name like A or A (2)


eclipse/src/saros/ui/wizards/pages/EnterProjectNamePage.java
```diff
@@ -171,8 +175,32 @@ private void attachListeners() {
...
+
+              final List<String> reservedNames = new ArrayList<>();
+
+              for (ProjectOptionComposite c : projectOptionComposites.values()) {
+                if (c.getProjectName().isEmpty()) continue;
+
+                reservedNames.add(c.getProjectName());
+              }
+
+              composite.getRemoteProjectID();
```
srossbach: remove

srossbach: Replaced

Issue #357, opened 2018-12-14
Move FollowingActivitiesManager from Eclipse part into the Core once it is needed in IntelliJ part


tobous:
Move [`FollowingActivitiesManager`](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/project/internal/FollowingActivitiesManager.java) into the core once it is needed by the IntelliJ part. See #350. As mentioned by the comment in `FollowingActivitiesManager`, this might require introducing an abstraction from the actual class handling the follow modes.

Related classes that also have to be looked at during this migration are [IFollowModeChangesListener](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/project/internal/IFollowModeChangesListener.java) and [AwarenessInformationCollector](https://github.com/saros-project/saros/blob/master/de.fu_berlin.inf.dpp/src/de/fu_berlin/inf/dpp/awareness/AwarenessInformationCollector.java).

PR #639, opened 2019-08-12
[INTERNAL][CORE] Merges FollowingActivitiesManager and FollowModeManager

srossbach:
This patch addresses issue #357, merging the remaining logic into the
core.


core/src/saros/editor/IFollowModeListener.java
```diff
@@ -27,7 +28,23 @@
...
+  }
+
+  /** Gets called when the <b>local</b> user stops following another session user. */
+  public default void stoppedFollowing(Reason reason) {
+    // NOP;
+  }
 
-  public void startedFollowing(User target);
+  /** Gets called when a <b>remote</b> user starts following another session user. */
+  public default void startedFollowing(final User follower, final User followee) {
```
tobous: It seems a bit weird that, with the current usage, this information (like who is now following whom) is never used. Instead, the whole state is just queried and the entire information is reloaded. But I guess that goes back to the same reasoning you mentioned in the review today, it being to cumbersome to update specific elements (and just reloading everything instead does not create that much of an overhead).

core/src/saros/editor/FollowModeManager.java
```diff
@@ -212,22 +249,29 @@ public void stop() {
...
-    User previouslyFollowedUser = currentlyFollowedUser;
+    final User previouslyFollowedUser = currentlyFollowedUser;
 
     currentlyFollowedUser = newFollowedUser;
 
     if (newFollowedUser != null && !newFollowedUser.equals(previouslyFollowedUser)) {
       editorManager.jumpToUser(newFollowedUser);
 
       if (previouslyFollowedUser != null) {
+        followModeStates.remove(localUser);
```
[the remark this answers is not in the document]
srossbach: It is basically to setup the state correctly when calling l.267

core/src/saros/editor/IFollowModeListener.java
srossbach: The whole Eclipse IDE can be created on the fly and disposed on the fly. So there is no chance to maintain the "state machine" somewhere in the GUI logic. Of course I only could load a snapshot on then do my updates, but TBH why should a developer does that when such "wonderful" methods are provided unless they would run in O(n^x)  with x >= 2 ?

Issue #392, opened 2019-01-23
EditorManager may fire duplicate partActivated events


tobous:
When the follower closes the active editor, they correctly leave the follow mode. But the follow mode button does not seem to be updated correctly. If the button is clicked, the editor currently focused by the followee is also focused by the follower and the followee displays locally that the follow mode has started. But the follower does not scroll or switch editors when the followee does, meaning they are not in the follow mode. If the follower presses the follow mode button again in this state, nothing happens. The follow mode display for the followee is also not updated when the follower takes an action that would otherwise end the follow mode (like manually switching the editor).

This issue can be avoided by explicitly selecting "Follow user XYZ" from the drop down menu of the follow mode button instead of just clicking the follow mode button itself on the followee side to start the follow mode.

This issue does not always occur. Sometimes the button does seem to work in this situation. From my testing however, it seems that more often then not, the issue occurs.

This issue was found and first discussed as part of #390.
The logic behind the button is located in [FollowModeAction](https://github.com/saros-project/saros/blob/master/eclipse/src/saros/ui/actions/FollowModeAction.java), but the actual problem might be located elsewhere.

#### Steps to Reproduce
1. Alice and Bob start a session containing the shared files A and B.
2. Alice and Bob both open an editor for A and B.
3. Alice selects the file A.
4. Bob follows Alice.
  Bob now also has the file A selected.
5. Bob closes the editor for A.
  Bob has now left the follow mode. This is also correctly displayed by Alice.
6. Bob uses the follow mode button (not the drop-down menu, the button itself) to again start the follow mode.
  This causes Bob to re-open the editor selected by Alice. Alice now sees Bob as following here.
7. Alice scrolls the editor or changes the selected editor.
  This action will not be transmitted to Bob.

srossbach:
After analyzing this someone this has nothing to do with the action. It is simply broken.

When you start follow mode you call EditorManager#jumpToUser which will trigger a listener forcing the follow mode to terminate.

So when you call FollowManager#follow you will generate a StopFollowingActitivty followed by a StartFollowingActivity. When the method returns you broadcasted a follow mode state however the internal state is non following.

srossbach:
As it turns out we hook into partBroughtOnTop events. What is going on is something like:

EditorManager#jumpToUser

[...]

PR #640, opened 2019-08-13
[FIX][E] #392 EditorManager may fire duplicate partActivated events

srossbach:
This patch fixes issue #392.

As it turns out we hook into partBroughtOnTop events. What is going on
is something like:

EditorManager#jumpToUser

For what reason ever when opening / switching to the new editor a
partBroughtOnTop event may be fired containing the current focused
editor. The logic transforms this into an editorActivated event which
gets passed back to the FollowModeManager before the
FollowModeManager#follow method is even finished. The result is that the
FollowModeManager state is changed to not following and after that it
will report that it has start following.

This patch will drop editorActivated events for editors that are
currently the active editor.

### DIFF
--- a/eclipse/src/saros/editor/EditorManager.java
+++ b/eclipse/src/saros/editor/EditorManager.java
@@ -839,6 +839,15 @@ public class EditorManager extends AbstractActivityProducer implements IEditorMa
     }
 
     SPath editorPath = EditorAPI.getEditorPath(editorPart);
+
+    if (editorPath.equals(locallyActiveEditor)) {
+      LOG.debug(
+          "ignoring partActivated event for editor path "
+              + editorPath
+              + " because it is already active");
+      return;
+    }
+
     TextSelection selection = EditorAPI.getSelection(editorPart);
 
     // Set (and thus send) in this order:


srossbach:
@m273d15 maybe creating a branch from the current release and applying this as hot fix ? My basic understanding is that when you activate follow mode and you are currently not on the same editor it will kick you out of the follow mode.

tobous:
> @m273d15 maybe creating a branch from the current release and applying this as hot fix ? My basic understanding is that when you activate follow mode and you are currently not on the same editor it will kick you out of the follow mode.

From just trying it out on the master without this patch, this does not seem to be true (starting the follow mode when on different editors works, but not consistently; in my limited testing, it still worked most of the time).

Either way, I don't think creating a hotfix branch for this is not necessary.
@srossbach @m273d15 How about just creating a real minor release? There already were a lot of usability improvements since the last release, so why not ship them as well?

srossbach:
I am not done yet with the stuff for the next Saros/E release. Maybe we can just add some API to the Core too as we has to increase the version number anyway because of the account store changes.

PR #646, opened 2019-08-13
[INTERNAL][E] Disable Concurrent Undo Feature

srossbach:
This feature was disable since Saros Version 14. However it seems that
the patch was not merged back after the release so this patch will
disabled it again so it is not accidently shipped.

Clarification: This feature is BROKEN as HELL, if you turn it on it is likely to freeze the IDE sooner or later.

### DIFF
--- a/eclipse/src/saros/SarosEclipseContextFactory.java
+++ b/eclipse/src/saros/SarosEclipseContextFactory.java
@@ -10,7 +10,6 @@ import saros.communication.SkypeManager;
 import saros.communication.chat.muc.negotiation.MUCNegotiationManager;
 import saros.communication.connection.IProxyResolver;
 import saros.communication.connection.Socks5ProxyResolver;
-import saros.concurrent.undo.UndoManager;
 import saros.context.AbstractContextFactory;
 import saros.context.IContextKeyBindings;
 import saros.editor.EditorManager;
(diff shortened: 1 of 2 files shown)


tobous:
What exactly is the concurrent undo feature?
Is it the restriction to only allow the user to undo local changes?

srossbach:
Alice, Bob

Alice adds Hello (as one operation)
Bob adds World (as one operation)

Without this feature if Alice invokes (IDE) undo it would remove World.
With this (broken) feature enabled Alice (IDE) undo would remove Hello.

srossbach:
Do not get me wrong. I strongly prefer that this feature has to be enabled all the times, however it does not work and I am not happy that it has to be disabled.

tobous:
I remember having a discussion with Franz about this topic when looking at it for Saros/I.
I am still not sure how to implement this properly. This feature gets significantly more complicated when the remote and local changes overlap. How should we determine the correct state after such an undo? (i.e. Alice changes "Hello" to "Hallo", and then Bob replaces it with "World"; what should Alice's undo do?)
How was this handled by the now disabled logic?
(Also, I am not sure whether this is even possible in IntelliJ as every document modification has to be wrapped as a "command", which automatically adds it to the undo queue.)

srossbach:
Just a looking at the logic is calculates the inverse of a JupterActivity that is applied locally.

The logic is done in UndoManager#calcUndoOperation

TBH, I only now what this component is supposed to do, not how. Currently it does not work well. There are also tests for the UndoManager, maybe there are more helpful. But guessing, based on you concerns, the logic does not cover all cases.

PR #655, opened 2019-08-20
Correctly initialize module as part of project negotiation instead of creating stub module

### LINKED ISSUE #649: HashCode implementation added in #539 breaks session when module is created during project negotiation
The new `hashCode()` implementation for `IntelliJProjectImpl` introduced in #539 breaks the Saros session when the shared module was created on the client side as part of the session negotiation.

This issue is caused by the fact that, in the process of creating a local representation of the shared module, a stub module is created to allow access to the local resources before the content of the shared module is available. When the session starts, the hash value of this stub module is used to initialize the `SharedProjectMapper`. However, this value is not updated when the module is reloaded from the received module file, which is problematic as the reload changes the hash value (since the held `Module` object is replaced).

<details><summary><i>More info on why this approach was chosen</i></summary>

When wanting to create the module as part of the project negotiation, all resources are currently zipped and transferred from the host to the client. These resources are then unpacked and written to the "local workspace". This writing to the "local workspace" (IntelliJ project) is done through the `IResource` interface. However, to create the needed `IProject` object, a local representation of the shared module is necessary. As this module does not exist yet, the logic creates a stub module. This stub module is used to create the needed `IntelliJProjectImpl` instance that is also added to the `SharedProjectMapper`. Later on, the module is reloaded from the transferred module file by the class `ModuleInitialization`. As part of reloading the module, the held `Module` object in the `IntelliJProjectImpl` instance held by the session is replaced with the new, reloaded `Module` reference.

</details>

#### Solution approach

I see two possible solution approaches (see below).

I have already implemented a prototype of solution 1. Despite that, I would still (somewhat strongly) prefer solution 2 as it seems considerably cleaner/less hacky. But solution two is still somewhat theoretical, so other problems may arise while implementing it. Attempting to use this approach is still worth it in my opinion.

##### 1. Update the `SharedProjectMapper` when reloading the module

I could add a method reloading all entries for the module in the `SharedProjectMapper` using the new hash value. This method would have to be available through the `ISarosSession` interface.
This is a **very** hacky solution as I can't use the normal collection methods to remove the entires (since they are based on the hash value, which no longer matches), meaning I have use iterators to find and remove the correct entry and the re-add it.

##### 2. Avoid the module stubbing by passing the module configuration as part of the `ProjectNegotiationData`

Instead of creating the module stub and later-on reloading the module from the transferred module file, I could instead pass the necessary information required to create the module (which currently should only be the module type; but the need for other options could arise in the future) as part of the `ProjectNegotiationData`. This would allow the logic to create the correct module from the get-got, avoiding the reload. As a result, we could remove the class `ModuleInitialization` and some code from `IntelliJProjectImpl` (e.g. the held `Module` object could be final again; the logic allowing the module object to change was iffy to begin with if it were to be called in any other circumstance than reloading the module stub; it's more of a hack to begin with).

This change would also require us to ignore module configuration files as they now no longer have to match for all participants. But, this is a change I was already considering anyways as the module configuration can theoretically also contain user configuration that is not the same for all participants. Furthermore, even though module file changes were technically "shared" with other participants with the current setup, the results of such changes were not applied to the modules of other participants as loading changes from disk to the active module require a complete module reload (as far as I know).

As this additional project data is not always needed and IDE dependent, I was thinking about a generalized approach. My current plan to extend the `ProjectNegotiationData` would be to add a new `Map<String,String>` to that is serialized and transferred as well. To fill this map, components can register properties providers that get passed an IProject object and then return two strings (key and value). These providers would then be called with the corresponding IProject object when creating a `ProjectNegotiationData` data instance to fill the map.

tobous:
Fixes #649 by directly creating the shared module with the correct type and loading the correct configuration instead of using the module file of the host.

#### [FEATURE][CORE] Add additional project data to project negotiation data

Adds a map for additional project data to the project negotiation data.

Subsequently adds a factory that can be used to build this mapping. The
entries for this mapping are provided by ProjectDataProviders that can
be registered with the factory.

This logic was added to allow us to pass IDE specific project
options/parameters as part of the project negotiation data without the
core having any IDE specific knowledge/logic.

Subsequently adjusts the AbstractOutgoingProjectNegotiation to build and
pass the mapping to the project negotiation data.

Base for fix for #649.

#### [INTERNAL][I] Add module configuration data provider

Adds a project data provider adding the local module configuration to
the project negotiation data.

This configuration data is needed on the receiving side when wanting to
create the module as part of the project negotiation.

Adds the data provider to the plugin context. The provider registers
itself automatically with the AdditionalProjectDataFactory.

Supplies the needed module configuration information for fix for #649.

#### [INTERNAL][I] Add module configuration initializer

Adds ModuleConfigurationInitializer to load the transferred module
configuration for modules created as part of the project negotiation.

This commit only adds the class. It will replace ModuleInitialization in
an upcoming change that moves the logic from using the transferred
module file to using the transferred module configuration to initialize
the module completely without the use of the host module file.

Adds logic to apply the configuration for fix for #649.

#### [FIX][I] #649 Use transferred module configuration for module creation

Adjusts the project negotiation logic to use the module configuration
options transferred as part of the project negotiation data. Previously,
this was done by creating a stub module and then reloading it from the
transferred module file. This lead to problems with the new hashCode
implementation for IntelliJProjectImpl, as it relies on the held module
object, which changes when reloading the module.

Subsequently prevents the module file from being shared. This is done by
adding a check to the IntelliJFileImpl.isDerived, leading to such files
being reported as derived.

Adjusts AddProjectToSessionWizard to use the transferred module
configuration data for the initial module creation and register it with
the ModuleConfigurationInitializer.

Adds the ModuleConfigurationInitializer to the session context to ensure
that the module configuration is correctly loaded.

Fixes #649.

Removes the ModuleInitializer from the session context as it is no
longer needed. The class will be removed in an upcoming cleanup commit.

#### [FIX][I] Check whether resource is derived when reporting initial state

Adds a check whether the resource for an open file of the shared module
is derived to EditorManager.AddProjectResources(IProject).

This check was added as trying to obtain an Editor object for a derived
resource leads to a null result which will be passed to NotNull
parameter without prior checking, leading to an assertion error.

This only now became an apparent issue as we previously only reported
resources as derived if they were non-existent, meaning the actual
"excluded" state known to IntelliJ is ignored.

#### [INTERNAL][I] Ensure ModifiableModuleRootModel is always disposed


core/src/saros/negotiation/ProjectNegotiationData.java
```diff
@@ -54,4 +74,13 @@ public String getProjectID() {
   public boolean isPartial() {
     return partial;
   }
+
+  /**
+   * Returns an unmodifiable view of the map of additional project options.
```
srossbach: Or null if a Saros client with a former version sent this packet. So question is: do you want to make the protocol incompatible ?

intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
```diff
@@ -0,0 +1,214 @@
...
+
+    Map<String, String> configurationOptions = queuedModuleOptions.remove(module);
+
+    if (configurationOptions == null) {
+      log.debug("Skipped module root model adjustment as not options were found - " + module);
+
+      return;
+    }
+
+    ModuleRootModificationUtil.updateModel(
```
srossbach: Does this ensure that the update is called in the correct thread context ? As far as I remember projectAdded may be called from any thread.

intellij/src/saros/intellij/negotiation/ModuleConfigurationProvider.java
```diff
@@ -0,0 +1,175 @@
...
+public class ModuleConfigurationProvider implements ProjectDataProvider {
+
+  public static final String MODULE_TYPE_KEY = "MODULE_TYPE";
+  public static final String SDK_KEY = "SKD";
+  public static final String SOURCE_ROOTS_KEY = "SOURCE_ROOTS";
+  public static final String TEST_SOURCE_ROOTS_KEY = "TEST_SOURCE_ROOTS";
+  public static final String RESOURCE_ROOTS_KEY = "RESOURCE_ROOTS";
+  public static final String TEST_RESOURCE_ROOTS_KEY = "TEST_RESOURCE_ROOTS";
+
+  private static final String DELIMITER = File.pathSeparator;
```
srossbach: Very bad idea.

Have you read the Javadoc ?

The **system-dependent path-separator** character, represented as a string for convenience. This string contains a single character, namely pathSeparatorChar.

Should be ";" on Windows and ":" on Unix IIRC.

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -221,6 +228,11 @@ private void doNewModule(
             return;
           }
 
+          sessionManager
+              .getSession()
```
srossbach: NPE number 1

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -221,6 +228,11 @@ private void doNewModule(
...
+              .getComponent(ModuleConfigurationInitializer.class)
```
srossbach: NPE number 2

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -339,25 +351,31 @@ public void run() {
   }
 
   /**
-   * Creates an empty stub module with the given name and given base path in the given project.
+   * Creates an empty base module with the given name, base path, and module type in the given
+   * project.
    *
-   * <p>The created module has the module type {@link IntelliJProjectImpl#RELOAD_STUB_MODULE_TYPE}
-   * which allows us to easily identify it as stub.
+   * <p>The created module does not contain any further configuration. Need configuration options
```
srossbach: Needed ?

core/src/saros/negotiation/AdditionalProjectDataFactory.java
```diff
@@ -0,0 +1,79 @@
...
+ * every time project negotiation data is created for a shared project.
+ *
+ * <p>The registered project data providers should provide a way of identifying their registered
+ * options (e.g. by providing references to the used key values).
+ *
+ * <p><b>NOTE:</b> This factory does not ensure that the keys used by different project data
+ * providers are disjoint. If multiple providers use the same key, the mapping of the provider added
+ * last will be used.
+ */
+public class AdditionalProjectDataFactory {
```
srossbach: I would either throw this into the PicoContainer or at least using the singleton pattern for this class.

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -177,15 +179,20 @@ public void next() {
...
-         * @see AddProjectToSessionWizard#createModuleStub(String, Path, Project)
+         * @see AddProjectToSessionWizard#createBaseModule(String, String, Path, Project)
          */
         private void doNewModule(
             @NotNull Project project, @NotNull String moduleName, @NotNull Path moduleBasePath) {
 
+          Map<String, String> moduleParameters =
+              negotiation.getProjectNegotiationData(remoteProjectID).getAdditionalProjectData();
+
+          String moduleType = moduleParameters.get(ModuleConfigurationProvider.MODULE_TYPE_KEY);
```
srossbach: moduleType could theoretically be null.

core/src/saros/negotiation/ProjectNegotiationData.java
tobous: Good point. I haven't thought of that. Can't think of a solution that is not breaking compatibility off the top of my head.

For Saros/I, this is much of a problem in my opinion. We don't have a lot of users and, more importantly, no version compatibility table, so every new version is seen as incompatible by default (but this should be fixed sooner than later; together with adjusting the existing logic to differentiate IDE implementations). Furthermore, the new module creation logic is dependent on the passed information, so backwards compatibility is not really an option.

But the issue would be Saros/E. I am not sure what would happen if an outdated version is setting up the session. Would the value really just be `null`? In that case, it wouldn't be an issue *yet* as Saros/E is not using the logic. On the other hand, if the usage is going to be added in the foreseeable future, it might be better to make a clean cut and break compatibility now while the plugin is still a bit in "turmoil" (i.e. a lot of stuff is happening/changing internally).

What are your opinions on the matter?

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -404,9 +418,18 @@ public Module compute() throws IOException {
...
+                    ModuleRootManager.getInstance(module).getModifiableModel();
 
-                modifiableRootModel.commit();
+                try {
+                  modifiableRootModel.addContentEntry(moduleRoot);
+                  modifiableRootModel.commit();
+
+                } finally {
+                  if (!modifiableRootModel.isDisposed()) {
+                    modifiableRootModel.isDisposed();
```
srossbach: dispose ?

intellij/src/saros/intellij/negotiation/ModuleConfigurationProvider.java
srossbach: Good luck finding a character that is disallowed in filenames under Unix 😆

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
srossbach: You can easily proof it, just disconnect (pull the plug). It is null afterwards.

https://github.com/saros-project/saros/blob/master/core/src/saros/session/SarosSessionManager.java#L368

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
srossbach: Hmm sorry I meant "Needed" instead of "Need". Maybe my English is bad and "Need" is correct although I hardly remember to even read such a sentence.

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
srossbach: The call says it can return null, mainly during session shutdown.

core/src/saros/negotiation/AdditionalProjectDataFactory.java
srossbach: Basically it just the design pattern. Leave it as it is, however the approach is very bad. We should not start overuse static access to some "components" or otherwise we end up in global state programming.

intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
```diff
@@ -0,0 +1,214 @@
...
+   *
+   * <p>Does nothing if the string representing the relative source root paths is <code>null
+   * </code>.
+   *
+   * @param module the module to add source root entry for
+   * @param contentEntry the content entry to add the roots to
+   * @param contentEntryFile the virtual file representing the content root
+   * @param rootType the type of source root to add the paths as
+   * @param relativeSourceRootPaths the list of relative path to add as source roots
+   * @param <P> the type of the source root
```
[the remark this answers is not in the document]
tobous: Because it was named that way in the IntelliJ code. Can replace it with `T`

intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
```diff
@@ -0,0 +1,214 @@
...
+ * This class ensures that modules which were transferred as part of the project negotiation are
+ * correctly configured. This is done by loading the correct module configuration from the
+ * additional project negotiation data once the module is added to the session.
+ */
+public class ModuleConfigurationInitializer implements Startable {
+  private static final Logger log = Logger.getLogger(ModuleConfigurationInitializer.class);
+
+  private final ISarosSession session;
+
+  private final Map<Module, Map<String, String>> queuedModuleOptions;
```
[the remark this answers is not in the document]
tobous: Isn't it enough that it is a concurrent map? Or do you want to add it for documentation's sake?

core/src/saros/negotiation/ProjectNegotiationData.java
srossbach: I do not see an issue to return an empty map if none was given. You should verify the content anyway.

intellij/src/saros/intellij/negotiation/ModuleConfigurationProvider.java
```diff
@@ -0,0 +1,175 @@
...
+    } else {
+      sdkName = "";
+    }
+    optionsMap.put(SDK_KEY, sdkName);
+
+    ContentEntry[] contentEntries = moduleRootManager.getContentEntries();
+
+    if (contentEntries.length != 1) {
+      throw new IllegalStateException(
+          "Encountered shared module with multiple content roots - " + module);
```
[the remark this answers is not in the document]
tobous: Yeah, I know. It was a lazy solution... Same goes for module creation in `AddProjectToSessionWizard`. Will adjust it.

intellij/src/saros/intellij/negotiation/ModuleConfigurationProvider.java
```diff
@@ -0,0 +1,175 @@
...
+  @Nullable
+  private String flatten(@NotNull String basePath, @NotNull List<SourceFolder> sourceFolders) {
+    if (sourceFolders.isEmpty()) {
+      return null;
+    }
+
+    return sourceFolders
+        .stream()
+        .map(SourceFolder::getFile)
+        .filter(Objects::nonNull)
```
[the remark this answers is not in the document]
tobous: If the root is "invalid" according to the documentation. What I think they mean by this is that it is present in the configuration but does not have a valid representation/counterpart in the filesystem. Such a configuration is possible.

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
tobous: Jop. No idea how to handle this though. Will think about it.

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
tobous: No, "Needed" is indeed correct. Will change it.

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
tobous: Jop. Will fix it.

core/src/saros/negotiation/AdditionalProjectDataFactory.java
tobous: I mean I can change it to a singleton as it doesn't really change anything for accessing the class. As already said, I don't really mind. Would this be preferable to you?

Or are you set on having it be part of the context?

intellij/src/saros/intellij/negotiation/ModuleConfigurationProvider.java
tobous: How about `:;` as the delimiter? As at least one of the two is forbidden in either windows or unix, it should be impossible to create a group of paths (in a single system) that breaks the delimiter scheme.

intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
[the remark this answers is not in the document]
tobous: (Here is the reference to the usage of `P` in [ContentEntry.setSourceFolder(...)](https://github.com/JetBrains/intellij-community/blob/master/platform/projectModel-api/src/com/intellij/openapi/roots/ContentEntry.java#L138)

core/src/saros/negotiation/AdditionalProjectDataFactory.java
srossbach: Having it part of the content would be nice, and I do not think it is that complicated to get it into the IPN / OPN.

intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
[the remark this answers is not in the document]
srossbach: Ah sorry it missed the initialization of this map.

intellij/src/saros/intellij/negotiation/ModuleConfigurationProvider.java
srossbach: Well you could escape like I did with the patch that introduced project name mapping or you dig through some Unix guidelines on how not to name files and compare it with Windows forbidden characters and then pick one that is forbidden under Windows and not recommended to use in filenames under Unix.

intellij/src/saros/intellij/negotiation/ModuleConfigurationProvider.java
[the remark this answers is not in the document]
tobous: After looking at it, `AddProjectToSessionWizard` is not the same. It uses a `ThrowableExecutable` and we catch the exceptions. It's not the prettiest solution, but it works.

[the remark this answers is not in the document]
tobous: Done.

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
tobous: Done. Aborted negotiation if type not provided.

tobous:
Used same escaping method as introduced in #595 for transferred source paths.

tobous:
Rebased onto current master. Added nullability annotations to methods `escape` and `unescape`.

PR #656, opened 2019-08-20
[LOG][CORE] Drop log level of PortUnreachableException

tobous:
The stun discovery often throws a PortUnreachableException that the ICMP
port is unreachable. This does not seem to hinder the functionality, but
is still logged as a warning.

This warning leads to confusion for the users when reporting bugs as
they assume that the log entry (including the stacktrace) is the cause
of whatever issue they are having with Saros.

To avoid this wrong assumption (and users only sending this part of the
stacktrace), the log level is reduced to info and the stacktrace is
dropped.

### DIFF
--- a/core/src/saros/net/stun/internal/StunServiceImpl.java
+++ b/core/src/saros/net/stun/internal/StunServiceImpl.java
@@ -10,6 +10,7 @@ import java.net.DatagramPacket;
 import java.net.DatagramSocket;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
+import java.net.PortUnreachableException;
 import java.net.SocketAddress;
 import java.net.SocketTimeoutException;
 import java.util.ArrayList;
@@ -140,6 +141,9 @@ public final class StunServiceImpl implements IStunService {
                 new InetSocketAddress(stunAddress, stunPort),
                 new InetSocketAddress(localAddress, 0),
                 timeout);
+      } catch (PortUnreachableException e) {
+        log.warn("aborted STUN discovery: " + e.getMessage());
+        return;
       } catch (IOException e) {
         log.error("an error occurred while performing a STUN discovery: " + e.getMessage(), e);
         return;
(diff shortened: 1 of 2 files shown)


core/src/saros/net/stun/internal/StunServiceImpl.java
```diff
@@ -140,6 +141,9 @@ public void run() {
                 new InetSocketAddress(stunAddress, stunPort),
                 new InetSocketAddress(localAddress, 0),
                 timeout);
+      } catch (PortUnreachableException e) {
+        log.info("Aborted stun discovery - ICMP port unreachable: " + e.getMessage());
```
srossbach: This exception is raised when you receive an ICMP packet stating the the port (in this case UDP) is not reachable.

The message reads like the ICMP port is not reachable. Guess you forgot a delimiter after ICMP.

For clarification the code here is executed on every interface, so I would suggest.

1. to log this a warning or debug.
2. the message should contain the interface that was used.

tobous: I set us as 'info' as I am not sure how important this information is. If its just our default stun discovery that always fails, this can be debug. But, if the user sets up a stun server and the connection is still refused, they should at least be able to find something about it in the log files (as far as I know, debug logging is disabled on the shipped Saros versions).

> Guess you forgot a delimiter after ICMP.

What exactly do you mean with this?
I know that the error message is somewhat redundant as it also reads "ICMP Port Unreachable", but I still decided to keep it in, in case there are other cases where the message might be useful. I probably should just drop the hard-coded message part of the log instead.

tobous:
Removed redundant wording, pushed log level back up to warn. Added a commit to remove the default stun server address.

core/src/saros/net/stun/internal/StunServiceImpl.java
```diff
@@ -140,6 +141,9 @@ public void run() {
                 new InetSocketAddress(stunAddress, stunPort),
                 new InetSocketAddress(localAddress, 0),
                 timeout);
+      } catch (PortUnreachableException e) {
+        log.warn("Aborted stun discovery - " + e.getMessage());
```
srossbach:
```suggestion
        log.warn("aborted STUN discovery: " + e.getMessage());
```

tobous: Done.

tobous:

PR #661, opened 2019-08-26
Offer module configuration adjustment for existing modules

tobous:
#### [INTERNAL][I] Add non-blocking yes/no dialog

Adds the option for a non-blocking yes/no dialog that accepts a runnable
that is run if the 'Yes' option is chosen.

#### [INTERNAL][I] Add data holder class for module configurations

Adds the data holder class ModuleConfiguration. The class encapsulates
the parsing of the module configuration option contained in the
additional project options transferred as part of the project
negotiation data.

#### [INTERNAL][I] Update module config of existing modules

Adjusts the ModuleConfigurationInitializer to also offer the user to
load the module config of the host in case an existing module is used
but the config differs from the received one.
This functionality is now useful as we no longer transfer the module
file.

Subsequently adjusts the logic applying the config to newly created
modules to also use the ModuleConfiguration data holder class.

Also adjusts ModuleConfigurationProvider.split(String) to accept (and
subsequently return) null values.

### DIFF
--- a/intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
+++ b/intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
@@ -5,19 +5,28 @@ import com.intellij.openapi.projectRoots.ProjectJdkTable;
 import com.intellij.openapi.projectRoots.Sdk;
 import com.intellij.openapi.roots.ContentEntry;
 import com.intellij.openapi.roots.ModifiableRootModel;
+import com.intellij.openapi.roots.ModuleRootManager;
 import com.intellij.openapi.roots.ModuleRootModificationUtil;
+import com.intellij.openapi.roots.SourceFolder;
 import com.intellij.openapi.vfs.VirtualFile;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
 import org.apache.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.jetbrains.jps.model.JpsElement;
-import org.jetbrains.jps.model.java.JavaResourceRootType;
-import org.jetbrains.jps.model.java.JavaSourceRootType;
 import org.jetbrains.jps.model.module.JpsModuleSourceRootType;
 import saros.filesystem.IProject;
 import saros.intellij.filesystem.IntelliJProjectImpl;
+import saros.intellij.ui.Messages;
+import saros.intellij.ui.util.SafeDialogUtils;
 import saros.repackaged.picocontainer.Startable;
 import saros.session.ISarosSession;
 import saros.session.ISessionListener;
@@ -32,7 +41,7 @@ public class ModuleConfigurationInitializer implements Startable {
 
   private final ISarosSession session;
 
-  private final Map<Module, Map<String, String>> queuedModuleOptions;
+  private final Map<Module, ModuleConfiguration> queuedModuleOptions;
 
   private final ISessionListener sessionListener =
       new ISessionListener() {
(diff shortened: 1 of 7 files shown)


intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
```diff
@@ -81,43 +95,163 @@ public void enqueueModuleConfigurationChange(
...
+      @NotNull ContentEntry contentEntry,
+      @NotNull Map<JpsModuleSourceRootType<? extends JpsElement>, String[]> rootPaths) {
+
+    for (Entry<JpsModuleSourceRootType<? extends JpsElement>, String[]> entry :
+        rootPaths.entrySet()) {
+
+      JpsModuleSourceRootType<? extends JpsElement> type = entry.getKey();
+      String[] paths = entry.getValue();
+
+      if (paths == null) {
```
tobous: Use empty set instead. Might differ even if no received. (case: local has config, host does not)

intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
```diff
@@ -81,43 +95,163 @@ public void enqueueModuleConfigurationChange(
...
+    if (moduleConfiguration == null) {
       log.debug("Skipped module root model adjustment as not options were found - " + module);
 
       return;
     }
 
-    ModuleRootModificationUtil.updateModel(
-        module,
-        modifiableRootModel -> applyModuleConfiguration(modifiableRootModel, configurationOptions));
+    if (!moduleConfiguration.isExistingModule()) {
```
srossbach: code can be changed so it is not nested so deep, see below

intellij/src/saros/intellij/ui/util/SafeDialogUtils.java
```diff
@@ -193,4 +194,34 @@ public static Integer showYesNoDialog(Project project, final String message, fin
...
+   * @param title the text displayed as the title of the dialog
+   * @param runAfter the runnable to execute if the user chooses {@link Messages#YES}
+   */
+  public static void showYesNoDialog(
+      @NotNull Project project,
+      @NotNull String message,
+      @NotNull String title,
+      @NotNull Runnable runAfter) {
+
+    LOG.info("Showing non-blocking yes/no dialog: " + title + " - " + message);
```
srossbach: debug ?

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -310,13 +282,70 @@ private void doExistingModule(@NotNull Module existingModule) {
...
+         *
+         * @param module the module the configuration belongs to
+         * @param moduleConfiguration the module configuration
+         */
+        private void queueModuleConfigurationChange(
+            @NotNull Module module, @NotNull ModuleConfiguration moduleConfiguration) {
+
+          ISarosSession session = sessionManager.getSession();
+
+          if (session == null) {
```
srossbach: The session manager already calls cancel. And so the cancelListener should be notified already.

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -310,13 +282,70 @@ private void doExistingModule(@NotNull Module existingModule) {
...
+                Messages.AddProjectToSessionWizard_no_session_message,
+                Messages.AddProjectToSessionWizard_no_session_title);
+
+            return;
+          }
+
+          ModuleConfigurationInitializer moduleConfigurationInitializer =
+              session.getComponent(ModuleConfigurationInitializer.class);
+
+          if (moduleConfigurationInitializer == null) {
```
srossbach: See above.

intellij/src/saros/intellij/ui/util/SafeDialogUtils.java
tobous: The info log level is currently used for all dialog methods. I am not sure what exactly led me to make this decision. We previously had a guideline when to use the different log-levels, but it got dropped during the website migration. (It was not that helpful anyways.)

I think it was something along the lines that we would like to know what was shown to the user when reading logs of released plugin versions (as released versions only log info and above). They can, for example, be used to get some information about the state of the session and session actions. For this reason, any dialog and notification is currently logged with the log level `info`.

Currently, I am not sure whether the release logging restriction even applies for Saros/I (I don't think so), but it still doesn't hurt to have these as an indication of the plugin state. This is especially useful for me as I use a plugin to fold any debug and trace loggin by default in the IDE, which allows me to only see the important information by default but still leaves me the option to check the debug loggin in the IDE when needed.

Either way, if this should get changed to debug, I would also change it for the other methods and the notification class (and move this change to a separate commit). But currently, I am leaning towards leaving it as `info` logging.

intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
tobous: Done.

intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
tobous: Done.

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
tobous: Done. Removed call to `cancelNegotiation(String)`.

srossbach: Looking at the code, when the wizard is cancelled a popup is already displayed. Of course it does not hurt however not displaying anything would  not hurt either.

PR #663, opened 2019-08-27
[INTERNAL][E] Logs warnings for incompatible editors.

srossbach:
Basically all document providers that do not extend from
TextFileDocumentProvider (like XText) are likely not to work with Saros.

Even retrieving the document provider from the editor will not help us
as we also operating in headless mode. Otherwise we would have to open
editors on the fly and so disturbing the user experience.

As we cannot do much about this issue for now log at least a message.

### DIFF
--- a/eclipse/src/saros/editor/EditorPool.java
+++ b/eclipse/src/saros/editor/EditorPool.java
@@ -15,9 +15,11 @@ import org.eclipse.jface.text.IDocumentListener;
 import org.eclipse.jface.text.ITextViewer;
 import org.eclipse.ui.IEditorInput;
 import org.eclipse.ui.IEditorPart;
+import org.eclipse.ui.editors.text.TextFileDocumentProvider;
 import org.eclipse.ui.ide.ResourceUtil;
 import org.eclipse.ui.texteditor.IDocumentProvider;
 import org.eclipse.ui.texteditor.IElementStateListener;
+import org.eclipse.ui.texteditor.ITextEditor;
 import saros.activities.SPath;
 import saros.editor.internal.EditorAPI;
 import saros.filesystem.ResourceAdapterFactory;
@@ -121,6 +123,8 @@ final class EditorPool {
       LOG.warn("editor part does not use a file storage: " + editorPart);
       return;
     }
+
+    findAndLogDocumentProviderIssues(editorPart);
 
     /*
      * Connecting causes Conversion of Delimiters which trigger Selection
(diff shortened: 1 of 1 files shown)


eclipse/src/saros/editor/EditorPool.java
```diff
@@ -310,4 +314,36 @@ public void setElementStateListenerEnabled(final boolean enabled) {
   public void setDocumentListenerEnabled(final boolean enabled) {
     documentListener.setEnabled(enabled);
   }
+
+  private static void findAndLogDocumentProviderIssues(final IEditorPart editorPart) {
+
+    final IDocumentProvider defaulDocumentProvider =
```
srossbach: missing a t

srossbach:
I would prefer a fix instead of a popup. Obtaining the correct DocumentProvider for an editor is not hard as you can see in the logging. The challenging part (and that has to be verified if it is possible) is to release our "headless" provider, committing all changes before a user opens the document locally. It also gets more complicated if you have the same file opened in multiple editors. I guess this issue was never really analyzed and evaluated as most of Eclipse default editors are using the shared TextFileDocument provider. However it would be interesting what will happen if you edit such a file when it is open multiple times, e.g are activities send for each editor (which would be wrong) ?  ... etc.

I guess you do not have such a hard time in IntelliJ.

tobous:
I can't really estimate how complicated a fix would be, so this is pretty much up to your judgement.
I am fine with merging this PR in its current state. However, as already mentioned, some kind of user warning would be nice if this issue is not fixed in the near future (i.e. before the next release).

> I guess you do not have such a hard time in IntelliJ.

Well it depends on what you mean. We don't have an issue with editor representation (as far as I know), but the issue with having the same file open in different Editors can also be a problem in IntelliJ (e.g. if you share two modules that both contain the same file, actions on that file will most likely be send twice, once for each module/representation; but that's not really important for this issue).

eclipse/src/saros/editor/EditorPool.java
```diff
@@ -310,4 +314,36 @@ public void setElementStateListenerEnabled(final boolean enabled) {
...
+      return;
+    }
+
+    final ITextEditor textEditor = editorPart.getAdapter(ITextEditor.class);
+
+    if (textEditor == null) return;
+
+    final IDocumentProvider editorDocumentProvider = textEditor.getDocumentProvider();
+
+    if (!(defaulDocumentProvider instanceof TextFileDocumentProvider)) {
```
tobous: Seems like you missed one.
```suggestion
    if (!(defaultDocumentProvider instanceof TextFileDocumentProvider)) {
```

srossbach:
This is clearly a bug, this online editor does not support refactoring ...

eclipse/src/saros/editor/EditorPool.java
srossbach: In theory yes, but looking at it it should be the editorDocumentProvider.

tobous:
> This is clearly a bug, this online editor does not support refactoring ...

Interesting. Do you mean it doesn't offer the functionality or that it's broken? I have never used the online editor.

eclipse/src/saros/editor/EditorPool.java
tobous: Makes sense^^

srossbach:
Was just kidding 😃 

You have not experienced the times where after a release test the bug tracker got filled with bugs that were mostly feature requests but not bugs.

PR #664, opened 2019-08-27
[FIX][I] Honor IntelliJ resource exclusion

tobous:
Adjusts the isDerived() implementation for the IntelliJ file and folder
implementation to also consider whether the resource is seen as excluded
for the corresponding project.

Even though this is a general improvement, it brings new possibilities
for issues during a session. When the excluded resource mapping does not
match for all clients, creating a local resource that already exists for
another client as an excluded resource leads to an unsolvable session
desync.

### DIFF
--- a/intellij/src/saros/intellij/editor/ProjectAPI.java
+++ b/intellij/src/saros/intellij/editor/ProjectAPI.java
@@ -5,6 +5,7 @@ import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.fileEditor.FileEditorManager;
 import com.intellij.openapi.fileEditor.OpenFileDescriptor;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.roots.ProjectFileIndex;
 import com.intellij.openapi.vfs.VirtualFile;
 import org.jetbrains.annotations.NotNull;
 import saros.intellij.filesystem.Filesystem;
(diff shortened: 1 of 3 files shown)


srossbach:
I do not know if excluded / ignored in IntelliJ context equals the Eclipse context of isDerived https://help.eclipse.org/kepler/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2FresAdv_derived.htm

I know it is handy to mark them as derived because the FileListFactory will ignore them.

Maybe we should either add filters are generalize the concept of isDerived.

srossbach:
To be more clear, while every derived resource can be excluded != every excluded resource is derived.

tobous:
Yes, not all excluded files are derived and the current usage in Saros/I goes beyond what derived files imply in Eclipse.
But, it is a mechanism to mark files as "not of interest for the IDE", which, in my mind, also means they shouldn't be of interest for the user / resource sharing.

I was also thinking about adding some kind of filter system (as Saros/I already kind of has a need for such a system with the filter for module files (which was already added) and for project configuration folders) but wasn't sure whether there actually was a need for it. But it does sound like a good idea. It would allow us to create IDE-dependent filters (e.g. derived for Eclipse, excluded for IntelliJ) and drop the IDE-independent concept of "derived resources" (which is not clearly defined).

tobous:
Superseded by #690.

PR #666, opened 2019-09-02
Allow different module names

tobous:
#### [INTERNAL][I] Allow different module names

Adjusts the project negotiation dialog to allow the accepting side
choose a module name different from the name on the host side. This is
not possible as we no longer share the module file.

Subsequently adds a check to the option to create a new module to
disallow to continue the dialog when a module with the chosen name
already exists in the chosen project. Previously, such a situation would
lead to the negotiation aborting with an error message.

#### [FIX][I] Only show fields as invalid if they are currently evaluated

Adjusts the dialog to accept a project negotiation to only show fields
as invalid if they are currently being evaluated. Otherwise, fields are
marked as invalid that are not used for the currently chosen mode, which
could lead to confusion.


tobous:

PR #668, opened 2019-09-02
Remove module name reference in IntelliJProjectImpl

tobous:
#### [INTERNAL][I] Remove held reference to module name

Removes the held reference to the module name and replaces it by
dynamically requesting the name from the module object. This change is
needed as module names can change during runtime, but the module is
always represented by the same object. This allows us to better handle
such name changes.

#### [INTERNAL][I] Adjust module comparison in IntelliJProjectImpl

Adjusts the module comparison in
IntelliJProjectImpl.getProjectRelativePath(VirtualFile) to use the
module objects directly. Previously, the module names were compared. As
already mentioned in the changes reworking the equals method, the name
can not be used as a unique identifier when multiple project are
allowed.


tobous:

PR #669, opened 2019-09-03
[TEST][E] Addresses #497

srossbach:
Added a minimal behavior to the editor manager mock.


srossbach:
If I am not mistaken "andStubReturn" cannot have a calling amount behavior. And yes I could reproduce this. It really depends how fast or slow your PC is.

tobous:
> If I am not mistaken "andStubReturn" cannot have a calling amount behavior.

You are right.

Issue #407, opened 2019-01-29
Replace `compile*` configurations in gradle build files


m273d15:
The gradle java plugin marked the dependency handling via `compile` and `testCompile` as deprecated. We should use `implementation`, `testImplementation`, `api` (which is new) and `provided` (instead of `compileOnly`).
See [here](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation) for more information

m273d15:
This issue has to be fixed before updating to Gradle 7 in the future. This is already mentioned by the gradle 6 build:
> Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.1/userguide/command_line_interface.html#sec:command_line_warnings

tobous:
Also, IntelliJ complains about the usage of some deprecated fields and some closure stuff in multiple of the `build.gradle` files. It isn't that important (and I think some of it was already present before the update to Gradle 6), but you could still have a look at it if you are in the process of cleaning up anyways.

m273d15:
I started to fix all these issues but recognized that this goes beyond updating gradle.
I think the tasks to fix this are:
* replace all usages of `version` with `archiveVersion`
* replace all usages of `xyCompile` with `xyImplementation` (which could break some logic and has to be tested with stf)

PR #671, opened 2019-09-05
Update gradle to 5.6.1

m273d15:
* Updated to the latest gradle version. 
The changes are created by the command:
`./gradlew wrapper --gradle-version=5.6.1`
* Update the gradle scan plugin to the latest version


srossbach:
Do you upgrade just because a new version was released or is there something specific you would like to resolve using a newer gradle version ?

m273d15:
I updated just to use the newer version.
Sooner or later I will adopt some new syntax which will replace the old syntax (#407).

gradlew
```diff
@@ -28,7 +44,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS='"-Xmx64m"'
+DEFAULT_JVM_OPTS='"-Xmx64m" "-Xms64m"'
```
srossbach: Just wondering if this is a bit too less memory. Have you checked garbage collector times ?

m273d15: The gradle [docs](https://docs.gradle.org/current/userguide/build_environment.html#sec:configuring_jvm_memory) say:
> The JAVA_OPTS environment variable controls the command line client, which is only used to display console output. It defaults to -Xmx64m

Issue #593, opened 2019-07-26
Improve separation between IDE specific documentation on website


tobous:
Currently, the sections that require IDE specific details (like installation or usage instructions) are split into different pages (e.g. [eclipse installation](https://www.saros-project.org/documentation/installation-eclipse.html) and [intellij installation](https://www.saros-project.org/documentation/installation-intellij.html).

This separation creates a lot of clutter in the navbar and makes it much harder to navigate the site or find the needed page.

A suggested alternative would be to split such pages into different tabs, each containing the detailed information for an implementation. This would also introduce better scalability when adding information for other implementations (like the server).
The biggest issue with this approach is that we need the user to notice the tabs, as they otherwise make the site harder to navigate instead of easier.

m273d15:
First proposal on my gh-page instance [here](https://m273d15.github.io/saros/documentation/getting-started.html) and [here](https://m273d15.github.io/saros/documentation/installation.html).
* Uses tabs in order to differentiate between different saros documentation flavors
* Switches the whole content if one tab is clicked (if we have just small boxes of IDE specific content it is possible to click on "eclipse" and all other boxes also show the "eclipse" content)

Known flaws:
* TOC is broken (might be be hard to fix! The current toc is created statically via native Jekyll commands. The new solution either requires to create two tocs and the js selects the current one or the toc has to be created dynamically)
* Not cite-able (It is currently not possible to reference a documentation flavor via url). Should be fixable with a js workaround.

Is this a usable approach? Any design proposals? @Drakulix already proposed to use an IDE specific border color. I am not sure whether the tabs are to big and should be shrinked. Otherwise:
> The biggest issue with this approach is that we need the user to notice the tabs, as they otherwise make the site harder to navigate instead of easier.

tobous:
I am not sure how this would look, but for readability, I would suggest moving the tabs to the left side. This would allow the reader to naturally come across the tabs when following the website-flow (nav-bar -> headline -> text).

Also as a side note: Currently, the tab that is *not* actually selected is highlighted as such (blue text). This is kind of confusing as it is opposite to the rest of the side (see nav-bar), where blue highlighting implies selection.

I think I would prefer a dynamically changing TOC. But that is just a first thought as this most likely has to be experimented with to figure out whether it is actually usable or just confusing. Can you approximate how much work such a dynamic TOC would take?

As an alternative, could you include the TOC in the tab? This way the tab would act more like a sub-page.

m273d15:
[...]

PR #672, opened 2019-09-05
Add IDE content tabs

m273d15:
This PR adds the function to create tabs for IDE specific content (as discussed in #593 ). These tabs have the following properties:
* The selection of all tabs on one page is synchronized (if you select the "eclipse" content in one tab, the selection of the other tabs are also changed)
* It is possible to reference a page with a specific tab selection via an url parameter `<url>?tab=<ide>` (e.g. `http://0.0.0.0:4000/contribute/documentation.html?tab=eclipse`)

I reworked the table of contents, because the shown content is not static anymore. The new toc is created based on the content which is shown by the selected tabs.

The implemented is split into the following methods and files:
* `ide-tabs.js` - checks whether the used url contains the `tab` parameter and changes the selection to the value. The file provides the methods `loadToc` and `selectTabs`.
* `default.html` - calls `loadToc()` --> initializes the toc with the currently shown content
* `ide-tabs.html` - calls `selectTabs( <ide> )` --> selects all corresponding tabs on the page --> calls `loadToc()` in order to refresh the toc


tobous:
It seems like referencing a tag breaks the automatic md->html resolution. E.g. if I try to link to a page `[stuff](page.md?tab=intellij#headline)`, the link does not get converted to `page.html?tab=intellij#headline` but rather keeps its `.md` ending, leading to an incorrect reference.

tobous:
For an example, you can check out and look at the new links in the commits on the branch [pr/doc/use_tabs](https://github.com/saros-project/saros/tree/pr/doc/use_tabs) which replaces the split documentation pages by single pages using the new tabs.

tobous:
Also, the navigation sidebar on the left gets messed up (is located to hight) when there is no scroll bar. This can be seen on the Screenshots page when on the IntelliJ tab.

docs/contribute/documentation.md
```diff
@@ -14,6 +14,44 @@ please follow these guidelines:
...
+{% include ide-tabs.html eclipse=eclipse intellij=intellij %}
+
+
+**Attention:** The major disadvantage of this solution is that we use [jekyll liquid tags](https://jekyllrb.com/docs/liquid/).
+Therefore the content of the documentation could break the documentation build! This happens if you have typo in your tags or forgot something etc.
+If you want to make sure that the documentation works, compile the documentation (see below for more information).
+
+## Reference IDE specific documentation
+
+You can reference the IDE specific version of a page via an url parameter `tab=<ide>` (e.g. `<url>?tab=intellij` or `<url>?tab=eclipse`).
```
tobous: This works in this example, but, as I have already noted in my other comments, does not specify whether to use the `.md` or `.html` ending for the references.
Normally, GitHub pages would use the `.md` ending everywhere as it allows you to also navigate the references in the source view and they are automatically resolved when building the pages. But, this resolution does not work with tab-specific links. The question is now whether there is some workaround for this (or another logic to replace the current tab specifier in the URL) or whether we want to directly use the `.html` reference for such references. This should then also be mentioned in this section.

docs/_includes/ide-tabs.html
```diff
@@ -0,0 +1,26 @@
...
+            </a>
+        </li>
+        <li class="nav-item">
+            <a class="nav-link intellij" onclick="selectTabs( 'intellij' );" id="intellij-tab" data-toggle="tab" href="#">
+                <img alt="IntelliJ logo" src="{{ '/assets/images/logo/intellij-logo.png' | relative_url }}" /> IntelliJ
+            </a>
+        </li>
+    </ul>
+
+    <div class="tab-content border rounded-bottom border-top-0">
```
tobous: For whatever reason, the vertical content border is missing for the first few pixels if the first content is a headline. This behavior can be observed on the [pr/doc/use_tabs](https://github.com/saros-project/saros/tree/pr/doc/use_tabs) branch on the "Installation" page on the eclipse tab (or see screenshot below).

saros_page_missing_border

m273d15:
btw I would squash the PR.

tobous:
> Also, there are some more pictures with a missing the `img-fluid` tag. But I can adjust them in my follow up PR. Otherwise, they would complicate the rebase anyways (as I copy/move a lot of content).

Ok, this is not as easy as pasting the tag everywhere pictures are linked. Sometimes, it messes up the formatting. @m273d15 Could I move that responsibility back to you with a followup PR after my tab-usage PR is merged?

tobous:
Here is a patch changing the occurrences of other image links that were not "tagged".
[image_mobile_fix.txt](https://github.com/saros-project/saros/files/3591147/image_mobile_fix.txt)
Those are just the `.jpg` and `.png` occurrences. I don't know if we use other extensions or if I missed some.

The patch messes up the formatting of the getting-started-page (host comics and icon placement in "Step 5: Exploring the Saros Toolbar".

PR #678, opened 2019-09-09
Adjusting SPath references in Saros/I with IntelliJReferencePointManager

pdler:
This PR contains adjustings of SPath references in Saros/I, because the SPath class can't only return the IReferencePoint and the relative path to the saros resource, but also the saros resources itself (like IFolder, IFile, ...).
The idea is, that the IntelliJReferencePointManager gets the SPath object and determinates IntelliJ and Saros resources

### DIFF
--- a/intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
+++ b/intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
@@ -32,7 +32,9 @@ import saros.activities.FolderCreatedActivity;
 import saros.activities.FolderDeletedActivity;
 import saros.activities.IActivity;
 import saros.activities.SPath;
+import saros.filesystem.IFile;
 import saros.filesystem.IPath;
+import saros.filesystem.IResource;
 import saros.intellij.editor.DocumentAPI;
 import saros.intellij.editor.EditorManager;
 import saros.intellij.editor.LocalEditorHandler;
@@ -40,6 +42,8 @@ import saros.intellij.editor.ProjectAPI;
 import saros.intellij.editor.annotations.AnnotationManager;
 import saros.intellij.eventhandler.IApplicationEventHandler;
 import saros.intellij.eventhandler.editor.document.LocalDocumentModificationHandler;
+import saros.intellij.filesystem.IntelliJFileImpl;
+import saros.intellij.filesystem.IntelliJReferencePointManager;
 import saros.intellij.filesystem.VirtualFileConverter;
 import saros.intellij.project.filesystem.IntelliJPathImpl;
 import saros.observables.FileReplacementInProgressObservable;
@@ -67,6 +71,7 @@ public class LocalFilesystemModificationHandler extends AbstractActivityProducer
   private final FileReplacementInProgressObservable fileReplacementInProgressObservable;
   private final AnnotationManager annotationManager;
   private final LocalEditorHandler localEditorHandler;
+  private final IntelliJReferencePointManager intelliJReferencePointManager;
 
   private boolean enabled;
   private boolean disposed;
@@ -176,7 +181,8 @@ public class LocalFilesystemModificationHandler extends AbstractActivityProducer
       ISarosSession session,
       FileReplacementInProgressObservable fileReplacementInProgressObservable,
       AnnotationManager annotationManager,
-      LocalEditorHandler localEditorHandler) {
+      LocalEditorHandler localEditorHandler,
+      IntelliJReferencePointManager intelliJReferencePointManager) {
 
     this.project = project;
 
@@ -190,6 +196,7 @@ public class LocalFilesystemModificationHandler extends AbstractActivityProducer
     this.disposed = false;
 
     this.localFileSystem = LocalFileSystem.getInstance();
+    this.intelliJReferencePointManager = intelliJReferencePointManager;
   }
 
   /**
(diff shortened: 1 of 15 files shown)


m273d15:
Is it possible to split this PR into multiple smaller PRs?

pdler:
> Is it possible to split this PR into multiple smaller PRs?

Before I rebased this branch, I saw over 40 commits in this PR. That was wrong, of course. 
Now, this PR has 5 commits, as it was intended, or do you mean, if it is possible to split this PR with 5 commits into multiple smaller PRs?

m273d15:
@pdler I talked about the 40 commits. I am fine with the current size

pdler:
@m273d15 @tobous , could you review this PR please, or should I split this PR into several PRs (because its to big?)

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
```diff
@@ -697,7 +721,11 @@ private void generateFileMove(
 
       editorManager.removeAllEditorsForPath(oldFilePath);
 
-      annotationManager.removeAnnotations(oldFilePath.getFile());
+      IFile file =
```
tobous: Why are you calculating this `IFile` object even though it is you never used?

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
```diff
@@ -617,11 +630,22 @@ private void generateFileMove(
...
+        intelliJReferencePointManager.getSarosResource(
+            oldFilePath.getReferencePoint(), oldFilePath.getProjectRelativePath());
+
+    IFile oldSarosFile = oldResource.adaptTo(IntelliJFileImpl.class);
+
+    IResource newResource =
+        intelliJReferencePointManager.getSarosResource(
+            newParentPath.getReferencePoint(), newParentPath.getProjectRelativePath());
+
+    IFile newSarosParent = newResource.adaptTo(IntelliJFileImpl.class);
```
tobous: This will always be a folder and not a file.

Why do you need to resolve the resource type here anyway? (Same goes for the `oldResource`.)

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
```diff
@@ -662,7 +686,7 @@ private void generateFileMove(
 
       editorManager.replaceAllEditorsForPath(oldFilePath, newFilePath);
 
-      annotationManager.updateAnnotationPath(oldFilePath.getFile(), newFilePath.getFile());
+      annotationManager.updateAnnotationPath(oldSarosFile, newSarosParent);
```
tobous: Needs to be the new file (`newFilePath`, instantiated in l.672) and not the parent folder (which is currently used).

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
pdler: Why should this always be a folder??

The `IntelliJReferencePointManager` only provides resolving the Saros resource for a given IReferencePoint and relative path. So the `IntelliJReferencePointManager` can be extended, so that it also returns a Saros File and/or folder. Or the IntelliJReferencePointManager resolves the Saros resources as `IResource` and converts it to IFile or IFolder, if it is really needed.

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
pdler: Yes right...fixed

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
pdler: Because...I don't know why. I removed this useless calculation...

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
tobous: > Why should this always be a folder??

This is always a folder as we don't have a reference for the new resource yet (as it does not exist). Instead, we use the reference for the *parent* of the new resource, which is why the SPath was called `newParentPath`. As it is a parent resource, it *must* be a folder.

So, in the new implementation, this should also be called `newParentResource`. The actual new resource reference is created in l.687 and l.716.

pdler: Ok, thank you for your explanation. Now I know what you meant. I adjusted the code fragment.

pdler: I would merge this PR, if you agree with the changes

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
```diff
@@ -369,7 +388,13 @@ private void generateResourceDeletionActivity(@NotNull VirtualFileEvent virtualF
 
     SPath path = VirtualFileConverter.convertToSPath(project, deletedVirtualFile);
 
-    if (path == null || !session.isShared(path.getResource())) {
+    IResource resource =
+        intelliJReferencePointManager.getSarosResource(
+            path.getReferencePoint(), path.getProjectRelativePath());
+
+    IFile file = resource.adaptTo(IntelliJFileImpl.class);
```
tobous: This resource does not have to be a file. This casting is unnecessary. Just use the `IResource` reference.

PR #680, opened 2019-09-09
Fix version mismatch notification and notification utils bug

### LINKED ISSUE #674: User not notified about session invitation failure when an error occurs.
Users are not notified that the session negotiation failed because of a version mismatch (the Saros version of the negotiation participants were not compatible). Instead, the negotiation just fails silently.

A better behavior would be to show a notification to the user stating this fact.

tobous:
#### [FIX][I] Fix project selection logic for notifications

Adjusts the notification panel logic to use the session project or, if
no such project is available, use the last focused project. This
guarantees that a project is always specified to guarantee that the
notification is always displayed as a pop-up.

Indirectly fixes a bug where the old logic requesting the Project object
directly from the plugin context was still used in
UIProjectUtils.getProjectFromSession(). This lead to the method always
returning null, even if a shared project was available.

Adds a method to check whether the shared IDE context is completely
initialized. This is needed as it is not guaranteed that the
notification logic is only called after the initialization. This is only
a workaround. The logic should be improved to better handle calls in the
uninitialized state.

Subsequently removes the method UIProjectUtils.getSharedProject() as it
is no longer used.

#### [FIX][I] #674 Show notification if outgoing session negotiation errors

Adjusts the logic in NegotiationHandler to display a user notification
when the outgoing session negotiation exist with an error status. This
is only a temporary fix. In the long-term, the usage of UIMonitoredJob
should be replaced by a generic implementation based on
ProgressMonitorAdater that handles all of these cases for all calls.

Fixes #674.

### DIFF
--- a/intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java
+++ b/intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java
@@ -122,7 +122,7 @@ public class NegotiationHandler implements INegotiationHandler {
     OutgoingInvitationJob(OutgoingSessionNegotiation negotiation) {
       super(
           MessageFormat.format(
-              Messages.NegotiationHandler_inviting_user, getNickname(negotiation.getPeer())));
+              Messages.NegotiationHandler_session_processing, getNickname(negotiation.getPeer())));
       this.negotiation = negotiation;
       peer = negotiation.getPeer().getBase();
     }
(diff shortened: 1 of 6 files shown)


intellij/src/saros/intellij/ui/util/UIProjectUtils.java
```diff
@@ -69,7 +58,13 @@ private Project getProjectFromSession() {
       return null;
     }
 
-    return currentSession.getComponent(Project.class);
+    SharedIDEContext sharedIDEContext = currentSession.getComponent(SharedIDEContext.class);
```
srossbach: Nullcheck missing.

intellij/src/saros/intellij/ui/Messages.java
```diff
@@ -85,10 +85,12 @@
   public static String ConnectButton_connect_to_new_account_title;
   public static String ConnectButton_connect_to_new_account_message;
 
-  public static String NegotiationHandler_canceled_invitation;
-  public static String NegotiationHandler_canceled_invitation_text;
-  public static String NegotiationHandler_error_during_invitation;
-  public static String NegotiationHandler_error_during_invitation_text;
+  public static String NegotiationHandler_session_local_error_title;
```
srossbach: I do not know if it is a good decision to mix session_negotiation and invention. Just stick to one wording (I do not prefer session negotiation and not invitation)

intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java
```diff
@@ -136,33 +136,41 @@ protected IStatus run(IProgressMonitor monitor) {
...
+
             return new Status(
                 IStatus.ERROR, SarosComponent.PLUGIN_ID, negotiation.getErrorMessage());
           case OK:
             break;
           case REMOTE_CANCEL:
             NotificationPanel.showInformation(
-                MessageFormat.format(Messages.NegotiationHandler_canceled_invitation_text, peer),
-                Messages.NegotiationHandler_canceled_invitation);
+                MessageFormat.format(
```
srossbach: Duplicated code. Maybe you can declare: String title, message at the top of the method an then assign the values. This is easier to read because you generate the message e.g 2 times here in this section.

intellij/src/saros/intellij/ui/util/UIProjectUtils.java
tobous: Done.

intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java
tobous: Done.

intellij/src/saros/intellij/ui/Messages.java
tobous: Done.

PR #689, opened 2019-09-11
Refactor IResource.isDerived()

tobous:
#### [REFACTOR][E] Always check ancestors for EclipseResourceImpl.isDerived()

Adjusts EclipseResourceImpl.isDerived() to always check ancestors. This
is done in preparation of the removal of IResource.isDerived(boolean),
which was only used by the Eclipse implementation.

#### [REFACTOR] Remove IResource.isDerived(boolean)

Removes IResource.isDerived(boolean) and its implementations as it is no
longer used.

#### [REFACTOR] Rename IResource.isDerived() to isIgnored()

Renames IResource.isDerived() to isIgnored() to avoid the
Eclipse-specific notion of derived resources.

This was done as there can be other reasons for resources to be ignored
by Saros, e.g. them being part of the IDE configuration.


core/src/saros/filesystem/IResource.java
```diff
@@ -47,7 +47,7 @@
 
   public int getType();
 
-  public boolean isDerived();
+  public boolean isIgnored();
```
srossbach: Documentation.

PR #694, opened 2019-09-19
Add some initial Saros-Server documentation

Drakulix:

Please, check that:
- [x] your code complies with the google java format (see [here](https://www.saros-project.org/contribute/development-environment.html) for details) or ESLint configuration

### Description

Some initial docs for Saros-Server in preparation of the first release.
Feel free to correct any spelling-mistakes / etc and make suggestions.

### DIFF
--- a/docs/contribute/saros-server.md
+++ b/docs/contribute/saros-server.md
@@ -31,6 +31,8 @@ The server will need some kind of permission system to allow a remote user to ma
 
 One server currently represents only one session. To lift this short-coming a Super-Server could be developed, that starts a new session-server on demand. This would also play nicely with the idea of an admin-user, as the user requesting the session could automatically be configured to be the first admin.
 
+This concept could prove to be very memory inefficient, as a new process and JVM would be started for each session. The overhead this brings needs to be evaluated to decide, how this should exactly be implemented. If the additional memory requirements are not overshadowed by Saros overall memory usage, an in-process solution could prove to be more reasonable.
+
 #### Challenges
 
 - It is not clear, how this can be represented in the XMPP-Protocol. The Super-Server would be logged with an XMPP-ID to setup the initial communication, but each session-server would need to be dynamically assigned a new JID or the same JID. New JIDs would need some kind of configuration and cooperation with the XMPP-Server. Re-using the same JID is currently not tested with Saros and might cause other unpredictable issues.
--- a/docs/contribute/saros-server.md
+++ b/docs/contribute/saros-server.md
@@ -1,0 +1,40 @@
+---
+title: Saros Server Development
+---
+
+## What is the Saros Server
+
+Please refer to the [User Documentation](../documentation/saros-server.md) for an explaination and the use-case of the saros server.
+
+Technically it is a headless session-host.
+
+## Code Structure
+
+The Saros Server codebase is comparably slim, as it only requires a [small subset of components](https://github.com/saros-project/saros/blob/master/server/src/saros/server/ServerContextFactory.java) compared to IDE-Plugins. Therefor it may also help to understand Saros overall architecture and as a reference for starting new Saros-Implementations.
+
+Notable exceptions are the [filesystem](https://github.com/saros-project/saros/tree/master/server/src/saros/server/filesystem), [editor](https://github.com/saros-project/saros/tree/master/server/src/saros/server/editor) and [preferences](https://github.com/saros-project/saros/tree/master/server/src/saros/server/preferences) implementations, which are usually adapted from IDE-specific APIs.
+
+## Open Topics
+
+### Permission System
+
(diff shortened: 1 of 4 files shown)


docs/documentation/saros-server.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Use-case
+
+The use-case of the server is to host session independently of any participant. As it takes the role of the session-host any participant may leave or join the session at any time.
+
+## Usage
+
+The Saros-Server is currently available in version **0.1.0** as a standalone jar on [our Github Releases](https://github.com/saros-project/saros/releases).
+
+The server needs it's own Jabber-ID and can then be started via:
```
m273d15: Even if we call it a JID( = Jabber ID) I would prefer to call it an "XMPP account" as in the getting started docs.
```suggestion
The server needs it's own XMPP account and can then be started via:
```

docs/contribute/saros-server.md
```diff
@@ -0,0 +1,40 @@
...
+
+## What is the Saros Server
+
+Please refer to the [User Documentation](../documentation/saros-server.md) for an explaination and the use-case of the saros server.
+
+Technically it is a headless session-host.
+
+## Code Structure
+
+The Saros Server codebase is comparably slim, as it only requires a [small subset of components](https://github.com/saros-project/saros/blob/master/server/src/saros/server/ServerContextFactory.java#L63) compared to IDE-Plugins. Therefor it may also help to understand Saros overall architecture and as a reference for starting new Saros-Implementations.
```
m273d15: Referencing a line has a high probability that the you reference the wrong code in the future.
```suggestion
The Saros Server codebase is comparably slim, as it only requires a [small subset of components](https://github.com/saros-project/saros/blob/master/server/src/saros/server/ServerContextFactory.java) compared to IDE-Plugins. Therefor it may also help to understand Saros overall architecture and as a reference for starting new Saros-Implementations.
```

docs/contribute/saros-server.md
```diff
@@ -0,0 +1,40 @@
...
+The server will need some kind of permission system to allow a remote user to manage these tasks. The simplest (and likely sufficient) variant would move this role to some kind of "Admin"-User, that may be configured for a given server.
+
+#### Challenges
+
+- New Activities will likely need to be introduced, that need to be understood by all existing IDE-Clients and show the appropriate dialogs.
+- There needs to be a system to transfer the admin rights, otherwise the admin-user becomes the new session-host, that will leave behind a broken (rather then disconnected) session, once he leaves.
+
+### Super-Server
+
+One server currently represents only one session. To lift this short-coming a Super-Server could be developed, that starts a new session-server on demand. This would also play nicely with the idea of an admin-user, as the user requesting the session could automatically be configured to be the first admin.
```
srossbach: This does not scale well (consumes to much memory) if you fork a child process for each session. I think 64 MB is the minimum of memory (guess this is still to low) that the process would need, So hosting about 30 session will already consume almost 2 GB.

Drakulix: This idea exists for quite some time already and has some nice properties:

- It is nicely capsulated, which makes this a good issue for a newcomer to the project (e.g. bachelor thesis)
- One session crashing would not crash the entire server
- It would likely simplify the code

Of course, each separate process and JVM would need additional memory, but a connected saros-server only takes about 28mb on my system, which already includes a good chunk of saros objects. So you are overestimating, what resources the JVM needs these days.

My guess is, that most of the memory would be used by the open documents / jupiter-algorithm / etc anyway, as saros was not designed to be particular memory-efficient anyway.

As a result, the server would need quite a lot of memory to be able to handle 30 sessions anyway. I expect an early release of this feature to be used for 5-10 session max. I think the main benefit is the ease of use, that it provides when you do not need to start the server manually.

That said, this is clearly something that needs to be evaluated by whoever want to implement this feature. I will this to the docs.

docs/contribute/saros-server.md
```diff
@@ -31,6 +31,8 @@ The server will need some kind of permission system to allow a remote user to ma
 
 One server currently represents only one session. To lift this short-coming a Super-Server could be developed, that starts a new session-server on demand. This would also play nicely with the idea of an admin-user, as the user requesting the session could automatically be configured to be the first admin.
 
+This concept could prove to be very memory inefficient, as a new process and JVM would be started for each session. The overhead this brings needs to be evaluated to decide, how this should exactly be implemented. If the additional memory requirements are not overshadowed by Saros overall memory usage, an in-process solution could prove to be more reasonable.
+
```
Drakulix: @srossbach I hope this explanation is satisfactory, as it clearly states, that the nature of the actual implementation still remains to be discussed/decided/evaluated.

srossbach:
I think the memory footprint can be even smaller, maybe 16 MB. However we have to major problems.

1. There is no traffic control, i.e you can flood the server with activities until it runs OOM.
2. The famous FileActivities (see various issues in the issue tracker) which produces OOM on either the sending or receiving side.

PR #704, opened 2019-09-25
[FIX][I] #703 Adjust URI parsing to work with Windows paths

### LINKED ISSUE #703: Windows: "Use existing module" fails in Saros/I
**Describe the bug**
Sharing a project with the option "Use existing module" fails with the following message if the client (not host) side is windows:
```
ERROR 13:18:43,982 [ApplicationImpl pooled thread 5] (Negotiation.java:357) IPN [remote side: dev8_alice_stf@saros-con.imp.fu-berlin.de/Saros] unknown error
java.nio.file.InvalidPathException: Illegal char <:> at index 4: file://C:/Users/kelvin/IdeaProjects/untitled1/SharedModule
	at java.base/sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at java.base/sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at java.base/sun.nio.fs.WindowsPath.parse(WindowsPath.java:92)
	at java.base/sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:229)
	at java.base/java.nio.file.Path.of(Path.java:147)
	at java.base/java.nio.file.Paths.get(Paths.java:69)
	at saros.intellij.negotiation.ModuleConfigurationInitializer.relativize(ModuleConfigurationInitializer.java:201)
	at saros.intellij.negotiation.ModuleConfigurationInitializer.lambda$configurationDiffers$3(ModuleConfigurationInitializer.java:178)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.AbstractList$RandomAccessSpliterator.forEachRemaining(AbstractList.java:720)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at saros.intellij.negotiation.ModuleConfigurationInitializer.configurationDiffers(ModuleConfigurationInitializer.java:179)
	at saros.intellij.negotiation.ModuleConfigurationInitializer.applyModuleConfiguration(ModuleConfigurationInitializer.java:126)
	at saros.intellij.negotiation.ModuleConfigurationInitializer.access$000(ModuleConfigurationInitializer.java:39)
	at saros.intellij.negotiation.ModuleConfigurationInitializer$1.resourcesAdded(ModuleConfigurationInitializer.java:51)
	at saros.session.internal.SessionListenerDispatch.resourcesAdded(SessionListenerDispatch.java:56)
	at saros.session.internal.SarosSession.addSharedResources(SarosSession.java:278)
	at saros.negotiation.AbstractIncomingProjectNegotiation.run(AbstractIncomingProjectNegotiation.java:207)
	at saros.intellij.ui.wizards.AddProjectToSessionWizard$7.run(AddProjectToSessionWizard.java:616)
	at com.intellij.openapi.progress.impl.CoreProgressManager$TaskRunnable.run(CoreProgressManager.java:894)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcess$2(CoreProgressManager.java:169)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:591)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:537)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:59)
	at com.intellij.openapi.progress.impl.CoreProgressManager.runProcess(CoreProgressManager.java:156)
	at com.intellij.openapi.progress.impl.CoreProgressManager$4.run(CoreProgressManager.java:408)
	at com.intellij.openapi.application.impl.ApplicationImpl$1.run(ApplicationImpl.java:294)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

**To Reproduce**
1. Start a session with a windows participant
2. the windows user uses the option "use existing module"

tobous:
#### [FIX][I] #703 Adjust URI parsing to work with Windows paths

Adjusts the URI parsing used when reading the source root configuration
of the shared module to also work with Windows paths.

Subsequently adjusted ModuleConfigurationProvider to also use the URI
approach instead of relying on the VirtualFiles of the roots. This was
done to unify the shared root state. When using the VirtualFiles, roots
that are registered with the module but don't have a valid
representation in the file system (and therefore don't have a valid
VirtualFile representation) would be ignored. Such roots are now also
transmitted by the host.

Subsequently moved the logic to parse and relativize root paths to
ModuleUtils so that it can be used by ModuleConfigurationProvider and
ModuleConfigurationInitializer.

Fixes #703.

#### [REFACTORING][I] Reduce method visibility in ModuleConfigurationProvider

### DIFF
--- a/intellij/src/saros/intellij/negotiation/ModuleUtils.java
+++ b/intellij/src/saros/intellij/negotiation/ModuleUtils.java
@@ -1,0 +1,56 @@
+package saros.intellij.negotiation;
+
+import com.intellij.openapi.roots.SourceFolder;
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.apache.log4j.Logger;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/** Provides utility methods to interact with modules. */
+class ModuleUtils {
+  private static final Logger log = Logger.getLogger(ModuleUtils.class);
+
+  /**
+   * Returns the relative path from the given base path to the relative source folder.
+   *
+   * @param basePath the base path to construct the relative path from
+   * @param sourceFolder the source folder to construct the relative path for
+   * @return the relative path from the given base path to the relative source folder or <code>null
+   *     </code> if the relative path could not be constructed
+   * @throws AssertionError when the path of the source folder does not start with the given base
+   *     path
+   */
+  @Nullable
+  static Path getRelativeRootPath(@NotNull Path basePath, @NotNull SourceFolder sourceFolder) {
+    String sourcePath = sourceFolder.getUrl();
+    try {
+      // Workaround to convert the URI to the canonical path needed for Windows paths
+      File childFile = new File(new URI(sourcePath).getPath());
+      Path childPath = Paths.get(childFile.getCanonicalPath());
+
+      assert childPath.startsWith(basePath)
+          : "Encountered path that is not located below the given base directory "
+              + basePath
+              + " - "
+              + childPath;
+
+      return basePath.relativize(childPath);
+
+    } catch (URISyntaxException e) {
+      log.warn("Could not parse source folder url", e);
+
+    } catch (IOException e) {
+      log.warn("Could not make source folder path canonical", e);
+
+    } catch (IllegalArgumentException e) {
+      log.warn("Could not construct relative path for the given source folder", e);
+    }
+
+    return null;
+  }
+}
--- a/docs/releases/saros-i_0.2.1.md
+++ b/docs/releases/saros-i_0.2.1.md
@@ -28,11 +28,12 @@ The current release `0.2.1` is not compatible with the previous Saros/I releases
 
 ## Changes
 
-- Fixed [#696](https://github.com/saros-project/saros/issues/696) - Color settings page not loading for Windows users
 - Allow project module to be shared
   - The project module (i.e. the module that contains the project configuration and is used to load the project) can not be shared through Saros. The project configuration files will be ignored by the session.
 - Honor excluded resources
   - The Saros session now ignores resources marked as 'excluded' in the IDE. As a result, such resources and activities modifying them won't be shared with other participants.
+- Fixed [#696](https://github.com/saros-project/saros/issues/696) - Color settings page not loading on the Windows platform
+- Fixed [#703](https://github.com/saros-project/saros/issues/703) - Fix "Use existing module" option failing on the Windows platform
 
 ## Features
 
--- a/intellij/resources/META-INF/plugin.xml
+++ b/intellij/resources/META-INF/plugin.xml
@@ -11,9 +11,10 @@
     <change-notes>
         <![CDATA[
         <ul>
-            <li>Fixed <a href="https://github.com/saros-project/saros/issues/696">#696</a> - Color settings page not loading for Windows users</li>
             <li>Allow project module to be shared</li>
             <li>Honor excluded resources</li>
+            <li>Fixed <a href="https://github.com/saros-project/saros/issues/696">#696</a> - Color settings page not loading on the Windows platform</li>
+            <li>Fixed <a href="https://github.com/saros-project/saros/issues/703">#703</a> - Fix "Use existing module" option failing on the Windows platform</li>
         </ul>
         <br>
         This list only refers to changes included in the latest release.
--- a/intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
+++ b/intellij/src/saros/intellij/negotiation/ModuleConfigurationInitializer.java
@@ -7,7 +7,6 @@ import com.intellij.openapi.roots.ContentEntry;
 import com.intellij.openapi.roots.ModifiableRootModel;
 import com.intellij.openapi.roots.ModuleRootManager;
 import com.intellij.openapi.roots.ModuleRootModificationUtil;
-import com.intellij.openapi.roots.SourceFolder;
 import com.intellij.openapi.vfs.VirtualFile;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -15,6 +14,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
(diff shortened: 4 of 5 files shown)


intellij/src/saros/intellij/negotiation/ModuleUtils.java
```diff
@@ -0,0 +1,56 @@
...
+   * @throws AssertionError when the path of the source folder does not start with the given base
+   *     path
+   */
+  @Nullable
+  static Path getRelativeRootPath(@NotNull Path basePath, @NotNull SourceFolder sourceFolder) {
+    String sourcePath = sourceFolder.getUrl();
+    try {
+      // Workaround to convert the URI to the canonical path needed for Windows paths
+      File childFile = new File(new URI(sourcePath).getPath());
+      Path childPath = Paths.get(childFile.getCanonicalPath());
```
tobous: This solution isn't pretty (and it took me waaay to long to find it), but it seems to work.

tobous:
I still have to adjust the release notes to mention this fix. Will do so later.

srossbach:
Why am I asked to do a review when you do not need my feedback at all ? I took a look at this patch at 5:00 A.M . However I have no clue what your issue are and why you have to use an Uri.

tobous:
> Why am I asked to do a review when you do not need my feedback at all ? I took a look at this patch at 5:00 A.M . However I have no clue what your issue is and why you have to use an Uri.

My bad. I was a little hasty because I really want to get the bugfix release 0.2.1 out of the door.

I specifically asked for you to have a look at it as you have more experience with the Windows platform.

The issue is the following:

In the ModuleConfigurationProvider/Initializer, I am dealing with the "source roots" of the shared modules, meaning which directories of the content root are marked as "Source", "Test Source", etc (currently excluding "Excluded" resources as they are more complicated).

To read these, I have two options: Request their VirtualFile or their URL. As the configuration doen't necessarily have to be represented by a file in filesystem (which would prevent me from obtaining the VirtualFile), I decided to use the URL of the file. This way, I can ensure that the configuration of the module on host and client side matches (initially; in-session changes will be dealt with later on). And if some configuration for non-existent source roots are transferred, it won't really hurt as they will simply be ignored by the module. (This patch also moves the sending side to using URLs.)

In the beginning, I simply used the URL for the `Paths.get(String)` call to obtain the path for the resource. And, while this works for Unix systems (which I tested on), this does fail for Windows systems as the URL has the form `file://C:/foo/bar/src`. The stacktrace is given in #703.

The issue was then that, even though using `Paths.get(new URI(String))` works, it provides the path in the wrong format. For the base path, I have the path in the format `C:\foo\bar`, but for the child resource, I have the format `//C/foo/bar/src` (or something along the lines). But I need the resources to match, so I had to use the canonical path.

The implementation added in this PR works, but I am not sure whether there is a better way or I am missing some corner cases and stuff like that.

srossbach:
Seeing this ```//C/foo/bar/src``` reminds me of something, but I do not know what exactly.

I guess this is not related to https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN

but some other faulty URI/URL parsing.

srossbach:
Ok seems that at least the File (not Paths) API is still buggy

```
File path: lua5.1.exe
Absolute file path: X:\\lua5.1.exe
Canonical file path: X:\lua5.1.exe
```

And I am just wondering why the canonical path is not Z:\lua\lua5.1.exe because X is a virtual drive.

srossbach:
```file://C:/foo/bar/src```is not a file url for the local system but an UNC path.

See: https://blogs.msdn.microsoft.com/ie/2006/12/06/file-uris-in-windows/

```
	URI u = new URI("file:///C:/Windows");
	
	Path path = Paths.get(u);
	
	Path relativePath = Paths.get("foo");
	Path child = path.resolve(relativePath);
	
	System.err.println(child);
```

Prints: ```C:\Windows\foo```

```
	URI u = new URI("file://C:/Windows");
	
	Path path = Paths.get(u);
	
	Path relativePath = Paths.get("foo");
	Path child = path.resolve(relativePath);
	
	System.err.println(child);
```

Prints: ```\\C\Windows\foo```

: is forbidden in such UNC path, I do not know why the Api is not throwing an exception

tobous:
Yeay, I got another issue related to this problem. The current solution does *not* work correctly, at least not in all cases.

The current parsing drops the drive letter for windows paths (happens [here](https://github.com/saros-project/saros/blob/098052bce4c0ce8d1a90a6be1569919c534b14b1/intellij/src/saros/intellij/negotiation/ModuleUtils.java#L33); it is missing from the result of `new URI(sourcePath).getPath()`).
As a result, when getting the canonical path, the "default" drive is used. (The "default" probably being the drive the IDE runs on or the user directory is contained in? It's not the one Windows is installed on for me.)
This leads to errors when the "default" is detected as a different drive as the project is on (i.e. default is "D:", but project is on "C:").

> `file://C:/foo/bar/srcis` not a file url for the local system but an UNC path.

I had another more look at the documentation you linked to and I don't actually think that the used path is a UNC path but rather that the path provided by IntelliJ is plain wrong. They use the double slash (`file://`), which denotes a reference to a network device, instead of the triple slash (`file:///`), which denotes a reference to a local file.

If I adjust the returned path to instead use the triple slash in the beginning (`file:///C:/path`) as described in https://blogs.msdn.microsoft.com/ie/2006/12/06/file-uris-in-windows/, everything works out great. The path retains the drive letter throughout the transformation, solving the issue described above.

Otherwise, if they wanted to use a UNC path, they would also have to specify the network device and replace ':' with '$' for the drive letter (i.e. `file://localhost/c$/path`). This works as well, but either way, the current formatting just seems wrong.

I will have a closer look at the internal parsing logic. If I can't find any reasoning for this weird URI format (that is not supported by Windows), I will open a bugtracker entry for it on the official IntelliJ tracker.

tobous:
Yay. Ok, the issue is known to the devs ([discussed here](https://youtrack.jetbrains.com/issue/IDEA-146869), but they won't fix it (as it is apparently only meant for "internal"/IDE-specific uses). This is, however, noted in the javadoc of `VirtualFile.getURL()`, which I seem to have missed during my investigation.

The fix is to use a utility class they provide to fix the incorrect usage of URIs.

PR #705, opened 2019-09-25
Wrap calls to FileIndex methods in a read action

tobous:
#### [INTERNAL][I] Remove Nullable annotation from Filesystem

Removes Nullable annotation from FileSystem.runReadAction and
runWriteAction as we can't make any statement about the return value of
the computation. Whether the returned value can be null or not depends
solely on the contents of the computation itself. This makes the
annotation not very helpful.

#### [INTERNAL][I] Wrap calls to FileIndex methods in a read action

Wraps any calls to methods of FileIndex or its implementations in a read
action to comply with the IntelliJ access guidelines and avoid illegal
access assertion errors.

### DIFF
--- a/intellij/src/saros/intellij/ui/menu/ShareWithUserAction.java
+++ b/intellij/src/saros/intellij/ui/menu/ShareWithUserAction.java
@@ -14,6 +14,7 @@ import org.apache.log4j.Logger;
 import saros.core.ui.util.CollaborationUtils;
 import saros.filesystem.IResource;
 import saros.intellij.context.SharedIDEContext;
+import saros.intellij.filesystem.Filesystem;
 import saros.intellij.filesystem.IntelliJProjectImpl;
 import saros.intellij.ui.Messages;
 import saros.intellij.ui.util.IconManager;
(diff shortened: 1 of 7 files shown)


tobous:
I would appreciate a thorough review for this one. Even though there isn't much happening, I managed to include two faults (flipped conditional and moved call before null check). I am *very* tired.

intellij/src/saros/intellij/ui/menu/ShareWithUserAction.java
```diff
@@ -61,21 +63,21 @@ public void actionPerformed(AnActionEvent e) {
...
 
     SharedIDEContext.preregisterProject(project);
     CollaborationUtils.startSession(resources, contacts);
   }
 
-  private IResource getModuleFromVirtFile(VirtualFile virtFile, Project project) {
+  private IResource getModuleForVirtualFile(VirtualFile virtualFile, Project project) {
+    ProjectFileIndex projectFileIndex = ProjectFileIndex.getInstance(project);
 
-    Module module = ProjectFileIndex.SERVICE.getInstance(project).getModuleForFile(virtFile);
```
m273d15: Is there a difference between `ProjectFileIndex.SERVICE.getInstance(project)` and `ProjectFileIndex.getInstance(project)`? You old version used the `SERVICE` version.

tobous: With the current implementation, it makes no difference. `ProjectFileIndex.SERVICE.getInstance(project)` internally just calls `ProjectFileIndex.getInstance(project)` (see [source](https://github.com/JetBrains/intellij-community/blob/master/platform/projectModel-api/src/com/intellij/openapi/roots/ProjectFileIndex.java#L25)).

Maybe this mattered at some point in the past? The usage was added years ago and the IntelliJ API changed a lot since then.

Issue #707, opened 2019-09-25
Line endings are overwritten when starting a session


tobous:
Saros/I overwrites the line endings of all client files with the ones used by the host. I.e. if the host uses Unix (LF) and the client Windows (CRLF), all files will be created as (LF) on the client side when creating the new module. Furthermore, when using an existing module, CRLF files will be detected as changed and will be overwritten by their LF counterpart from the host as part of the negotiation.

This is due to [`IntelliJFileImpl#getContents()`](https://github.com/saros-project/saros/blob/master/intellij/src/saros/intellij/filesystem/IntelliJFileImpl.java#L144) using the VirtualFile content, meaning the content on disk.

#### Solution approach
An alternative to using the VirtualFile content could be using the document representing the file. This would allow us to access the content independently of the line endings.

srossbach:
I do not think this is a bug but a configuration issue. On Eclipse we transmit the meta data files which include the line separator for the project and specific files. I do not see a reason to use Windows Line endings on Windows systems when it is not requested.

tobous:
> I do not see a reason to use Windows Line endings on Windows systems when it is not requested.

The issue is that the local configuration is completely ignored during the negotiation.

At any other point during the session, this is not an issue as the [`Document`](http://www.jetbrains.org/intellij/sdk/docs/basics/architectural_overview/documents.html) content is used for transmitting content and creating checksums for the consistency watchdog. The document content is what is internally managed by IntelliJ. Its consistent across all platforms as it always uses '\n' line endings.

But, during the negotiation (to create the FileList, checksum hashes and archive), `IFile.getContent()` is used. This uses the [`VirtualFile`](http://www.jetbrains.org/intellij/sdk/docs/basics/architectural_overview/virtual_file.html) to just output a stream of the file contents *on disk*, which does contain operating system specific information (like the line endings).

As a result, the client line endings will always be overwritten with the ones on the host side during the negotiation. Afterwards, this is no longer an issue and both client and host can change line endings freely without influencing each other (since the document content is used).

srossbach:
So why do you want to change the IntelliJ implementation of getContent (which will result in transmitting pure binary files as UTF-8 encoded strings) instead of negotiating the line delimiter for the module ?

tobous:
[...]

PR #708, opened 2019-09-25
[I] Add reference to bug #707 to release notes

tobous:


m273d15:
Do add the reference also in plugin.xml?

tobous:
The plugin.xml currently only contains the fixed issues (change log) and not the known bugs.

PR #714, opened 2019-10-02
Fix #223 - Drop vector time when resource is removed

### LINKED ISSUE #223: Vector time does not get reset correctly when deleting a file

### LINKED ISSUE #710: Deleting a shared file cause vector time mismatch errors
Deleting a file during a session causes periodic errors (see stacktrace below) 
 on the host side due to vetor time mismatches.

These errors seem to stem from the [`HeartBeatDispatcher`](https://github.com/saros-project/saros/blob/master/core/src/saros/concurrent/management/HeartbeatDispatcher.java) as they no longer occur when the dispatcher is disabled.

It is not clear to me whether this is caused by the dispatcher directly (e.g. by it trying to create activities for deleted files) or whether the dispatcher is just the visible failure caused by an issue of the internal vector time handling. If it's the latter, it might be related to #223.

The issue does not seem to be critical as I couldn't observe any issues with the session after the issue occurred. Activities are still properly transferred and applied. But maybe there are issues in corner cases like when executing the recovery.

```Java
ERROR 17:23:56,178 [saros-DispatchContext] (ConcurrentDocumentServer.java:137) failed to transform jupiter activity: JupiterActivity(timestamp: [11,5], operation: Noop(0,''), source: intellij_test_1@saros-con.imp.fu-berlin.de/Saros)
saros.concurrent.jupiter.TransformationException: precondition #2 violated (Remote vector time is greater than local vector time).
	at saros.concurrent.jupiter.internal.Jupiter.checkPreconditions(Jupiter.java:265)
	at saros.concurrent.jupiter.internal.Jupiter.receiveJupiterActivity(Jupiter.java:158)
	at saros.concurrent.jupiter.internal.JupiterDocumentServer.transformJupiterActivity(JupiterDocumentServer.java:65)
	at saros.concurrent.management.JupiterServer.transform(JupiterServer.java:104)
	at saros.concurrent.management.ConcurrentDocumentServer.receive(ConcurrentDocumentServer.java:135)
	at saros.concurrent.management.ConcurrentDocumentServer.transformIncoming(ConcurrentDocumentServer.java:111)
	at saros.session.internal.ActivityHandler.directServerActivities(ActivityHandler.java:396)
	at saros.session.internal.ActivityHandler.handleIncomingActivities(ActivityHandler.java:146)
	at saros.session.internal.SarosSession.exec(SarosSession.java:730)
	at saros.session.internal.ActivitySequencer$3.run(ActivitySequencer.java:362)
	at saros.util.ThreadUtils$1.run(ThreadUtils.java:34)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

#### Steps to recreate
The error is somewhat inconsistent when just deleting a file (but it does also occur when just removing a pre-existing shared file). A consistent setup is as follows:

- Start a session on an existing module
- Create a file on the host side
- Delete the created file on host side
- Wait for at least a minute to ensure a heartbeat occurred

Works on Saros/I and Saros/E. The setup also works when creating and deleting the file on the client side.

tobous:
#### [FIX][CORE] #223 Drop vector time when resource is removed

Adjusts ConcurrentDocumentClient and ConcurrentDocumentServer to drop
the vector time for a resource when it is deleted (or moved, thereby
"deleting" the resource with the previous path).

Adds DeletedResourceFilter to encapsulate the logic of detecting and
correctly handling resource deletions.
Furthermore, it provides a way of filtering out subsequent activities
for deleted resource.  This is needed as there can be straggling
activities for deleted resources that were created by remote
participants before they executed the deletion activity for the
resource.  Such activities are now filter out until the corresponding
resource is recreated.  This ensures that activities for the new
resource will still be processed correctly.
Checksum activities confirming the file deletion are explicitly excluded
from this filer (i.e. they will not be dropped for deleted resources) as
they are needed to check the consistency of all clients.

Fixes #223. Also fixes #710.

#### [INTERNAL][CORE] Add deletion acknowledgments

Adds DeletionAcknowledgmentActivity, which can be used to acknowledge
that a file was successfully deleted locally as a reaction to a received
activity. This acknowledgment allows us to drop the activity filter for
the deleted resource once all other session participants have
acknowledged that they received the deletion.

Adds the DeletionAcknowledgmentDispatcher to automatically dispatch an
acknowledgment when an activity is received that contains a file
deletion.

#### [INTERNAL][CORE] Use deletion acknowledgments to drop activity filters

Uses received deletion acknowledgment activities to determine when all
other participants have received the deletion, allowing the local
instance to drop the activity filter for the deleted resource.

#### [REFACTOR][CORE] Adjust wording in activity filter

Adjusts the wording in the activity filter to reflect that it is
specifically dealing with files and not resources in general.

#### [FIX][CORE] Exclude editor closed activities from filter

Adjusts ResourceActivityFilter.isFiltered(IActivity) to also never
filter out editor activities of the type "CLOSED". This is necessary as
such activities are used to clean up the user editor state.

Adjusts the wording of the javadoc for isFiltered(IActivity).

Adjusts code in ResourceActivityFilter to always uses the full qualifier
when using a variable of the type 'FileActivity.Type'. This was done to
make them easier to differentiate from the newly introduced usage of
variables of the type 'EditorActivity.Type'.

#### [DOC][CORE] Add documentation on how to handle deleted child resources

Adds information on how to handle deleted child resources to
FolderDeletedActivity.

Adds notes that activities might be filtered out to the corresponding
methods in ConcurrentDocumentServer and ConcurrentDocumentClient.

### DIFF
--- a/core/src/saros/concurrent/management/ConcurrentDocumentClient.java
+++ b/core/src/saros/concurrent/management/ConcurrentDocumentClient.java
@@ -10,6 +10,7 @@ import saros.activities.SPath;
 import saros.activities.TextEditActivity;
 import saros.concurrent.jupiter.Operation;
 import saros.concurrent.jupiter.TransformationException;
+import saros.repackaged.picocontainer.Startable;
 import saros.session.ISarosSession;
 
 /**
@@ -22,7 +23,7 @@ import saros.session.ISarosSession;
  * <p>When JupiterActivities are received from the server they are transformed by the
  * ConcurrentDocumentClient to TextEditActivities which can then be executed locally.
  */
-public class ConcurrentDocumentClient {
+public class ConcurrentDocumentClient implements Startable {
 
   private static Logger log = Logger.getLogger(ConcurrentDocumentClient.class);
 
@@ -36,7 +37,17 @@ public class ConcurrentDocumentClient {
     this.sarosSession = sarosSession;
     this.jupiterClient = new JupiterClient(sarosSession);
 
-    this.deletedResourceFilter = new DeletedResourceFilter(jupiterClient::reset);
+    this.deletedResourceFilter = new DeletedResourceFilter(sarosSession, this::reset);
+  }
+
+  @Override
+  public void start() {
+    deletedResourceFilter.initialize();
+  }
+
+  @Override
+  public void stop() {
+    deletedResourceFilter.dispose();
   }
 
   /**
@@ -175,7 +186,7 @@ public class ConcurrentDocumentClient {
    *     authoritative one and thus does not need to be reset).
    */
   public synchronized void reset(SPath path) {
-    log.debug("Resetting jupiter client: " + path.toString());
+    log.debug("Resetting jupiter client for " + path);
     jupiterClient.reset(path);
   }
 
--- a/core/src/saros/concurrent/management/ConcurrentDocumentClient.java
+++ b/core/src/saros/concurrent/management/ConcurrentDocumentClient.java
@@ -4,9 +4,7 @@ import java.util.ArrayList;
 import java.util.List;
 import org.apache.log4j.Logger;
 import saros.activities.ChecksumActivity;
-import saros.activities.FileActivity;
 import saros.activities.IActivity;
-import saros.activities.IActivityReceiver;
 import saros.activities.JupiterActivity;
 import saros.activities.SPath;
 import saros.activities.TextEditActivity;
@@ -32,10 +30,13 @@ public class ConcurrentDocumentClient {
 
   private final JupiterClient jupiterClient;
 
-  public ConcurrentDocumentClient(ISarosSession sarosSession) {
+  private final DeletedResourceFilter deletedResourceFilter;
 
+  public ConcurrentDocumentClient(ISarosSession sarosSession) {
     this.sarosSession = sarosSession;
     this.jupiterClient = new JupiterClient(sarosSession);
+
+    this.deletedResourceFilter = new DeletedResourceFilter(jupiterClient::reset);
   }
 
   /**
@@ -68,6 +69,9 @@ public class ConcurrentDocumentClient {
       return jupiterClient.withTimestamp(checksumActivity);
 
     } else {
+      deletedResourceFilter.handleResourceDeletion(activity);
+      deletedResourceFilter.handleResourceCreation(activity);
+
       return activity;
     }
   }
@@ -91,7 +95,13 @@ public class ConcurrentDocumentClient {
     List<IActivity> activities = new ArrayList<IActivity>();
 
     try {
-      activity.dispatch(clientReceiver);
+      deletedResourceFilter.handleResourceCreation(activity);
+
+      if (deletedResourceFilter.isFiltered(activity)) {
+        log.debug("Ignored activity for already deleted resource: " + activity);
+
+        return activities;
+      }
 
       if (activity instanceof JupiterActivity) {
         activities.addAll(receiveActivity((JupiterActivity) activity));
@@ -101,6 +111,8 @@ public class ConcurrentDocumentClient {
       } else {
         activities.add(activity);
       }
+
+      deletedResourceFilter.handleResourceDeletion(activity);
 
     } catch (Exception e) {
       log.error("Error while transforming activity: " + activity, e);
@@ -122,17 +134,6 @@ public class ConcurrentDocumentClient {
     }
     return activity;
   }
-
-  /** Used to remove JupiterClientDocuments for deleted files */
-  private final IActivityReceiver clientReceiver =
-      new IActivityReceiver() {
-        @Override
-        public void receive(FileActivity fileActivity) {
-          if (fileActivity.getType() == FileActivity.Type.REMOVED) {
-            jupiterClient.reset(fileActivity.getPath());
-          }
-        }
-      };
 
   /**
    * Transforms the JupiterActivity back into textEditActivities.
--- a/core/src/saros/concurrent/management/ConcurrentDocumentClient.java
+++ b/core/src/saros/concurrent/management/ConcurrentDocumentClient.java
@@ -31,23 +31,23 @@ public class ConcurrentDocumentClient implements Startable {
 
   private final JupiterClient jupiterClient;
 
-  private final DeletedResourceFilter deletedResourceFilter;
+  private final ResourceActivityFilter resourceActivityFilter;
 
   public ConcurrentDocumentClient(ISarosSession sarosSession) {
     this.sarosSession = sarosSession;
     this.jupiterClient = new JupiterClient(sarosSession);
 
-    this.deletedResourceFilter = new DeletedResourceFilter(sarosSession, this::reset);
+    this.resourceActivityFilter = new ResourceActivityFilter(sarosSession, this::reset);
   }
 
   @Override
   public void start() {
-    deletedResourceFilter.initialize();
+    resourceActivityFilter.initialize();
   }
 
   @Override
   public void stop() {
-    deletedResourceFilter.dispose();
+    resourceActivityFilter.dispose();
   }
 
   /**
@@ -80,8 +80,8 @@ public class ConcurrentDocumentClient implements Startable {
       return jupiterClient.withTimestamp(checksumActivity);
 
     } else {
-      deletedResourceFilter.handleResourceDeletion(activity);
-      deletedResourceFilter.handleResourceCreation(activity);
+      resourceActivityFilter.handleFileDeletion(activity);
+      resourceActivityFilter.handleFileCreation(activity);
 
       return activity;
     }
@@ -106,9 +106,9 @@ public class ConcurrentDocumentClient implements Startable {
     List<IActivity> activities = new ArrayList<IActivity>();
 
     try {
-      deletedResourceFilter.handleResourceCreation(activity);
+      resourceActivityFilter.handleFileCreation(activity);
 
-      if (deletedResourceFilter.isFiltered(activity)) {
+      if (resourceActivityFilter.isFiltered(activity)) {
         log.debug("Ignored activity for already deleted resource: " + activity);
 
         return activities;
@@ -123,7 +123,7 @@ public class ConcurrentDocumentClient implements Startable {
         activities.add(activity);
       }
 
-      deletedResourceFilter.handleResourceDeletion(activity);
+      resourceActivityFilter.handleFileDeletion(activity);
 
     } catch (Exception e) {
       log.error("Error while transforming activity: " + activity, e);
--- a/core/src/saros/concurrent/management/ConcurrentDocumentClient.java
+++ b/core/src/saros/concurrent/management/ConcurrentDocumentClient.java
@@ -93,6 +93,9 @@ public class ConcurrentDocumentClient implements Startable {
    * <p>This method will transform them back from Jupiter-specific activities to locally executable
    * activities. @GUI Must be called on the GUI Thread to ensure proper synchronization
    *
+   * <p>Drops activities that are reported as filtered out by {@link
+   * ResourceActivityFilter#isFiltered(IActivity)}.
+   *
    * @host and @client This is called whenever activities are received from REMOTELY both on the
    *     client and on the host
    * @param activity The activity to be transformed
--- a/core/src/saros/concurrent/management/ConcurrentDocumentServer.java
+++ b/core/src/saros/concurrent/management/ConcurrentDocumentServer.java
@@ -34,7 +34,7 @@ public class ConcurrentDocumentServer implements Startable {
 
   private final JupiterServer server;
 
-  private final DeletedResourceFilter deletedResourceFilter;
+  private final ResourceActivityFilter resourceActivityFilter;
 
   /** {@link ISessionListener} for updating Jupiter documents on the host. */
   private final ISessionListener sessionListener =
@@ -55,36 +55,36 @@ public class ConcurrentDocumentServer implements Startable {
     this.sarosSession = sarosSession;
     this.server = new JupiterServer(sarosSession);
 
-    Consumer<SPath> deletedResourceHandler =
+    Consumer<SPath> deletedFileHandler =
         resource -> {
           LOG.debug("Resetting jupiter server for " + resource);
           server.removePath(resource);
         };
 
-    this.deletedResourceFilter = new DeletedResourceFilter(sarosSession, deletedResourceHandler);
+    this.resourceActivityFilter = new ResourceActivityFilter(sarosSession, deletedFileHandler);
   }
 
   @Override
   public void start() {
     sarosSession.addListener(sessionListener);
-    deletedResourceFilter.initialize();
+    resourceActivityFilter.initialize();
   }
 
   @Override
   public void stop() {
     sarosSession.removeListener(sessionListener);
-    deletedResourceFilter.dispose();
+    resourceActivityFilter.dispose();
   }
 
   /**
-   * Calls {@link DeletedResourceFilter#handleResourceDeletion(IActivity)} and {@link
-   * DeletedResourceFilter#handleResourceCreation(IActivity)} with the given activity.
+   * Calls {@link ResourceActivityFilter#handleFileDeletion(IActivity)} and {@link
+   * ResourceActivityFilter#handleFileCreation(IActivity)} with the given activity.
    *
    * @param activity the activity to handle
    */
   public void handleResourceChange(IActivity activity) {
-    deletedResourceFilter.handleResourceDeletion(activity);
-    deletedResourceFilter.handleResourceCreation(activity);
+    resourceActivityFilter.handleFileDeletion(activity);
+    resourceActivityFilter.handleFileCreation(activity);
   }
 
   /**
@@ -106,7 +106,7 @@ public class ConcurrentDocumentServer implements Startable {
 
     final List<QueueItem> result = new ArrayList<QueueItem>();
 
-    if (deletedResourceFilter.isFiltered(activity)) {
+    if (resourceActivityFilter.isFiltered(activity)) {
       LOG.debug("Ignored activity for already deleted resource: " + activity);
 
       return result;
--- a/core/src/saros/concurrent/management/ConcurrentDocumentServer.java
+++ b/core/src/saros/concurrent/management/ConcurrentDocumentServer.java
@@ -6,9 +6,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import org.apache.log4j.Logger;
 import saros.activities.ChecksumActivity;
-import saros.activities.FileActivity;
 import saros.activities.IActivity;
-import saros.activities.IActivityReceiver;
 import saros.activities.JupiterActivity;
 import saros.activities.QueueItem;
 import saros.activities.SPath;
@@ -35,6 +33,8 @@ public class ConcurrentDocumentServer implements Startable {
 
   private final JupiterServer server;
 
+  private final DeletedResourceFilter deletedResourceFilter;
+
   /** {@link ISessionListener} for updating Jupiter documents on the host. */
   private final ISessionListener sessionListener =
       new ISessionListener() {
@@ -53,6 +53,8 @@ public class ConcurrentDocumentServer implements Startable {
   public ConcurrentDocumentServer(final ISarosSession sarosSession) {
     this.sarosSession = sarosSession;
     this.server = new JupiterServer(sarosSession);
+
+    this.deletedResourceFilter = new DeletedResourceFilter(server::removePath);
   }
 
   @Override
@@ -66,24 +68,15 @@ public class ConcurrentDocumentServer implements Startable {
   }
 
   /**
-   * Dispatched the activity to the internal ActivityReceiver. The ActivityReceiver will remove
-   * FileDocuments when the file has been deleted.
+   * Calls {@link DeletedResourceFilter#handleResourceDeletion(IActivity)} and {@link
+   * DeletedResourceFilter#handleResourceCreation(IActivity)} with the given activity.
    *
-   * @param activity Activity to be dispatched
+   * @param activity the activity to handle
    */
-  public void checkFileDeleted(final IActivity activity) {
-    activity.dispatch(hostReceiver);
+  public void handleResourceChange(IActivity activity) {
+    deletedResourceFilter.handleResourceDeletion(activity);
+    deletedResourceFilter.handleResourceCreation(activity);
   }
-
-  private final IActivityReceiver hostReceiver =
-      new IActivityReceiver() {
-        @Override
-        public void receive(final FileActivity activity) {
-          if (activity.getType() == FileActivity.Type.REMOVED) {
-            server.removePath(activity.getPath());
-          }
-        }
-      };
 
   /**
    * Transforms the given activities on the server side and returns a list of QueueItems containing
@@ -104,9 +97,13 @@ public class ConcurrentDocumentServer implements Startable {
 
     final List<QueueItem> result = new ArrayList<QueueItem>();
 
-    try {
-      activity.dispatch(hostReceiver);
+    if (deletedResourceFilter.isFiltered(activity)) {
+      LOG.debug("Ignored activity for already deleted resource: " + activity);
 
+      return result;
+    }
+
+    try {
       if (activity instanceof JupiterActivity) {
         result.addAll(receive((JupiterActivity) activity));
 
--- a/core/src/saros/concurrent/management/ConcurrentDocumentServer.java
+++ b/core/src/saros/concurrent/management/ConcurrentDocumentServer.java
@@ -91,6 +91,9 @@ public class ConcurrentDocumentServer implements Startable {
    * Transforms the given activities on the server side and returns a list of QueueItems containing
    * the transformed activities and there receivers.
    *
+   * <p>Drops activities that are reported as filtered out by {@link
+   * ResourceActivityFilter#isFiltered(IActivity)}.
+   *
    * @host
    * @sarosThread Must be executed in the Saros dispatch thread.
    * @notGUI This method may not be called from SWT, otherwise a deadlock might occur!!
--- a/core/src/saros/concurrent/management/ConcurrentDocumentServer.java
+++ b/core/src/saros/concurrent/management/ConcurrentDocumentServer.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.function.Consumer;
 import org.apache.log4j.Logger;
 import saros.activities.ChecksumActivity;
 import saros.activities.IActivity;
@@ -54,17 +55,25 @@ public class ConcurrentDocumentServer implements Startable {
     this.sarosSession = sarosSession;
     this.server = new JupiterServer(sarosSession);
 
-    this.deletedResourceFilter = new DeletedResourceFilter(server::removePath);
+    Consumer<SPath> deletedResourceHandler =
+        resource -> {
+          LOG.debug("Resetting jupiter server for " + resource);
+          server.removePath(resource);
+        };
+
+    this.deletedResourceFilter = new DeletedResourceFilter(sarosSession, deletedResourceHandler);
   }
 
   @Override
   public void start() {
     sarosSession.addListener(sessionListener);
+    deletedResourceFilter.initialize();
   }
 
   @Override
   public void stop() {
     sarosSession.removeListener(sessionListener);
+    deletedResourceFilter.dispose();
   }
 
   /**
--- a/core/src/saros/concurrent/management/ResourceActivityFilter.java
+++ b/core/src/saros/concurrent/management/ResourceActivityFilter.java
@@ -22,29 +22,29 @@ import saros.session.ISarosSession;
 import saros.session.ISessionListener;
 import saros.session.User;
 
-/** Class to handle resource deletions and filter out activities for already deleted resources. */
-class DeletedResourceFilter {
-  private static final Logger log = Logger.getLogger(DeletedResourceFilter.class);
+/** Class to handle file deletions and filter out resource activities for already deleted files. */
+class ResourceActivityFilter {
+  private static final Logger log = Logger.getLogger(ResourceActivityFilter.class);
 
   private final ISarosSession sarosSession;
 
-  /** Method passed by the constructor caller that can be used to react to resource deletions. */
-  private final Consumer<SPath> resourceDeletionHandler;
+  /** Method passed by the constructor caller that can be used to react to file deletions. */
+  private final Consumer<SPath> fileDeletionHandler;
 
   /**
-   * A map of shared resources that were deleted during the session onto the pending acknowledgments
-   * from other participants. It is used to filter out activities for such resources that were
-   * created before the other participants ran the corresponding resource deletion activity locally.
+   * A map of shared files that were deleted during the session onto the pending acknowledgments
+   * from other participants. It is used to filter out activities for such files that were created
+   * before the other participants ran the corresponding file deletion activity locally.
    *
-   * <p>The set is updated once the resource is recreated as we then want to handle activities for
-   * it again. Furthermore, it is updated once all acknowledgments for a resource deletion were
-   * received as the filter is then no longer necessary.
+   * <p>The set is updated once the file is recreated as we then want to handle activities for it
+   * again. Furthermore, it is updated once all acknowledgments for a file deletion were received as
+   * the filter is then no longer necessary.
    *
    * <p>This way of filtering can lead to issues when the order of activities is not preserved, e.g.
    * when we receive the content change for a new file is received before the file creation. Such
    * activities will be dropped, leading to inconsistencies.
    */
-  private final Map<SPath, List<User>> deletedResources;
+  private final Map<SPath, List<User>> deletedFileFilter;
 
   /** Activity consumer processing received deletion acknowledgments. */
   private final IActivityConsumer activityConsumer =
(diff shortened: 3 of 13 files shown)


tobous:
I am still not completely happy with this as the set of deleted resources is never cleared (unless the resource is recreated), meaning it continuously grows. Furthermore, we store references that might not be needed anymore as all participants might already have executed the deletion/move activity.

I only see two options for this:
- Hold a timestamp for each entry and prune it periodically.
- Introduce an ACK for resource deletions/moves so that we know when every participant has successfully processed it.

Pruning seems easier, but ACKs would be a nicer solution as it guarantees that all activities are handled correctly (which could be an issue if we prune to early).

@Drakulix  @m273d15 @srossbach Any opinions?

srossbach:
I updated my comment yesterday regarding the issue. You cannot simply delete something. If the vector time mismatch the Watchdog is unable to detect inconsistencies. As an example. When you perform a recovery (which includes a reset on the client as well on the server side) the whole session is locked/blocked, i.e nobody can edit files and modify something in the editor.

As you are primary testing under IntelliJ and your plugin only supports 2 people you MUST NOT assume that if something works for n = 2 users that this will work for n > 2 users.

core/src/saros/concurrent/management/ConcurrentDocumentClient.java
```diff
@@ -123,16 +160,55 @@ private IActivity receiveChecksum(ChecksumActivity activity) {
...
+      SPath removedFile = fileActivity.getPath();
+
+      deletedResources.add(removedFile);
+      jupiterClient.reset(removedFile);
+
+    } else if (fileActivity.getType() == Type.MOVED) {
+      SPath removedFile = fileActivity.getOldPath();
+
+      deletedResources.add(removedFile);
+      jupiterClient.reset(removedFile);
```
srossbach: And now you send the activity and in the meantime you receive and JupiterActivity for such file, what will happen ?

core/src/saros/concurrent/management/ConcurrentDocumentClient.java
```diff
@@ -32,10 +35,23 @@
...
+   */
+  // FIXME this could introduce issues when the order of activities is changed, i.e. we receive the
+  //  content change before the file creation
+  private final Set<SPath> deletedResources;
 
+  public ConcurrentDocumentClient(ISarosSession sarosSession) {
     this.sarosSession = sarosSession;
     this.jupiterClient = new JupiterClient(sarosSession);
+
+    this.deletedResources = new CopyOnWriteArraySet<>();
```
srossbach: This is single threaded, both the Client and the Server should only be accessed by a dedicated thread (yes I know the reset stuff does it not correctly)

srossbach:
So after taking a quick look you changed the logic to:

When I see a create / delete activity the specific document gets a state which decide if an activity is applied. If you are not in that state do the old logic and create documents on the fly ?

tobous:
> When I see a create / delete activity the specific document gets a state which decide if an activity is applied. If you are not in that state do the old logic and create documents on the fly ?

Yes. When a file is deleted, before the vector time is reset, it is added to the set of deleted resources. This happens for the server and the client whenever they receive a deletion/move activity. Whenever the server or client receives an `IResourceActivitiy` for a resource contained in the set, it is simply dropped. This was added to avoid the case that we receive an activity for the deleted file after its deletion, thereby recreating its jupiter document.

Resources are removed from the set when they are recreated, either through a file creation or a file move. This ensures that activities will be processed again for the newly created resource.

And yes, I did not change the old logic, so jupiter documents are still created on the fly. But due to the filter, it should not matter.

core/src/saros/concurrent/management/ConcurrentDocumentClient.java
tobous: Such activities should be filtered out by the logic in [l.124](https://github.com/saros-project/saros/pull/714/files#diff-cfaccc9315d374b587ed16f4876e129fR124).

core/src/saros/concurrent/management/ConcurrentDocumentClient.java
tobous: Done. But this only applies to the client. The server does not have the same guarantees according to the javadoc of `transformIncoming(...)`.

srossbach: The client should only be accessed through SWT/AWT/Application Thread. The server (if I am not mistaken) is accessed only the the NetworkDispatchThread (which IMO is pretty bad as it blocks every message send through the IPacketListeners). But I remember there was a dirty hook (client / server ?) that reacts on user leave events and that is main reason why there are concurrent collections / synchronized stuff although it should not.

Drakulix:
> I am still not completely happy with this as the set of deleted resources is never cleared (unless the resource is recreated), meaning it continuously grows. Furthermore, we store references that might not be needed anymore as all participants might already have executed the deletion/move activity.
> 
> I only see two options for this:
> 
>     * Hold a timestamp for each entry and prune it periodically.
> 
>     * Introduce an ACK for resource deletions/moves so that we know when every participant has successfully processed it.
> 
> 
> Pruning seems easier, but ACKs would be a nicer solution as it guarantees that all activities are handled correctly (which could be an issue if we prune to early). So I think I am currently leaning towards adding an ACK.
> 
> @Drakulix @m273d15 @srossbach Any opinions?

While I like the overall approach, this remaining problem needs to be solved. Otherwise, we are introducing a new memory leak to fix an issue, that was mostly caused by a fix to an even worse leak.

(But reading through the patch, this is something any activity could actually trigger, it is just much more likely with the HeartbeatDispatcher, right?)

I like the ACK-solution, as it is cleaner, but I think the prune is more robost. With the ACK we could keep entries in cases, where we miss the ACK (e.g. an unexpected disconnect). One can absolutely argue, that this is negligible. I guess I am fine with either solution.

core/src/saros/concurrent/management/ConcurrentDocumentClient.java
tobous: Hmm. But wouldn't it then be the better option to keep it as a synchronized access in both classes and add a todo to make it unsynchronized once we are sure that it's accessed correctly? The overhead from having the set synchronized should be minimal.

tobous:
> (But reading through the patch, this is something any activity could actually trigger, it is just much more likely with the HeartbeatDispatcher, right?)

No, this issue is independent from the hearbeat stuff. It just has the same solution.
The underlying issues are that
- the jupiter document/vector time is not dropped for deleted resources on the server. (The logic was present but faulty. See [old l.78](https://github.com/saros-project/saros/pull/714/files#diff-3747eb5ea0b67332ed770d4509a0015dL78).)
- the jupiter time is only dropped when receiving a deletion activity, not when sending it. I.e. the state of the participant that deleted the resource is not updated.
- straggling activities sent before the deletion was applied locally where not ignored.

This issue is that this causes a desync when subsequently recreating a deleted resource as the vector times no longer match.

The issue for #710 is that the participant that deleted the resource still sends heartbeats for the deleted resource as the local jupiter document was never reset.

Both issues are solved by this patch as it adjusts the behavior to also drop the vector time on outgoing deletion activities.

tobous:
Updated the change by introducing `DeletedResourceFilter` to encapsulate the filter logic. This was done to avoid code duplication and to provide a central place to put later additions of the logic (like pruning or ACK processing).

tobous:
I am not sure whether adding ACKs is really the best solution. I implemented it and it seems to work (but I only did *very* limited testing with 2 participants; but that is not the point I am trying to make), but it adds a lot of logic (about 220 lines of code before cleanup and adding javadoc), a new activity type, and a lot of network traffic (as each user has to send an ACK for every other participant besides the creator of the deletion activity). Furthermore, as this adds to the concurrent nature of the problem, it is annoying to debug and to ensure that it correctly works for multiple clients.

The prune solution on the other hand needs about 70 LOC and it ensures that the list of deleted resources is always cleared.

I still think that using ACKs is a more elegant solution, but I am still pondering on whether it is worth the hassle (testing troubles and increased code complexity, which makes it harder to maintain).

tobous:
The current approach only looks at file activities. This is problematic when a folder is deleted as the vector time of the contained files will not be reset. However, this issue can not be solved with the current logic as the resources will already have been deleted on the creator's side, meaning we can't access the child resources.

The only solution I can think of is to require the creator of the activity to always completely tear down deleted folders (i.e. create deletion activities for all contained resources that are dispatched before the actual folder is deleted).

This is currently not handled in such a way in Saros/I, but there is already a TODO for it (as this is also needed for other cleanup stuff) and its a relatively easy change.

In Saros/E, the logic in [`ProjectDeltaVisitor`](https://github.com/saros-project/saros/blob/master/eclipse/src/saros/project/ProjectDeltaVisitor.java) seem to already handle the child resources in such a way (as children of folders are visited and the activities are sorted to ensure that files are deleted before folders). But I am not completely sure on this. @srossbach Any input?

Furthermore, if this is now a requirement for folder deletions, this should be put in writing somewhere. Probably the javadoc of the folder deletion activity.

srossbach:
> 
> In Saros/E, the logic in [`ProjectDeltaVisitor`](https://github.com/saros-project/saros/blob/master/eclipse/src/saros/project/ProjectDeltaVisitor.java) seem to already handle the child resources in such a way (as children of folders are visited and the activities are sorted to ensure that files are deleted before folders). But I am not completely sure on this. @srossbach Any input?
> 

We have no guideline on how to send file deletions when they are triggered from another delete operation (e.g delete a folder). Unless this is specified somewhere I would not make any assumption which data in which order you get. E.g you could just send a folder delete activity, but I guess this would blow up the partial sharing as it also have to keep track of the current resources.

Drakulix:
> I am not sure whether adding ACKs is really the best solution. I implemented it and it seems to work (but I only did _very_ limited testing with 2 participants; but that is not the point I am trying to make), but it adds a lot of logic (about 220 lines of code before cleanup and adding javadoc), a new activity type, and a lot of network traffic (as each user has to send an ACK for every other participant besides the creator of the deletion activity). Furthermore, as this adds to the concurrent nature of the problem, it is annoying to debug and to ensure that it correctly works for multiple clients.
> 
> The prune solution on the other hand needs about 70 LOC and it ensures that the list of deleted resources is always cleared.

That is what I was fearing. I would prefer the prune solution in that case.

tobous:
> We have no guideline on how to send file deletions when they are triggered from another delete operation (e.g delete a folder). Unless this is specified somewhere I would not make any assumption which data in which order you get. E.g you could just send a folder delete activity, but I guess this would blow up the partial sharing as it also have to keep track of the current resources.

@srossbach I know that we *currently* don't have any guidelines on this. My question was rather whether my understanding of the Eclipse handling (i.e. that child resources are explicitly handled) is correct. From looking at the code, it looks that way to me, but I am not familiar with the details of the Eclipse `IResourceDeltaVisior` that is used in the `ProjectDeltaVisior`. But, from the javadoc, it seems that way as [`visit(...)`](https://github.com/saros-project/saros/blob/master/eclipse/src/saros/project/ProjectDeltaVisitor.java#L65) always returns `true` for folders and the activities should be sorted by the [`sort(...)`](https://github.com/saros-project/saros/blob/master/eclipse/src/saros/project/ProjectDeltaVisitor.java#L306) method.

As I see no other way of handling this case, I would just make this explicit handling of child resources a new guideline.

tobous:
Hmmmm... a big problem with the prune/cleanup solution is that we don't have a timeout for clients. In other words, it is completely acceptable for a client to be stuck for a while and the process everything that happened in the meantime without causing a desync.

An example of this happening in a normal session is the developer executing some blocking actions (e.g. opening the settings menu in IntelliJ) that block the IDE execution thread. This blocks the execution of received activities (like file deletions).

This makes it basically impossible to set a cleanup window (i.e. after what time to throw away the deletion filter for a particular resource) that is guaranteed to work.

tobous:
> Lets see. First behavior like you explained above. I then changed it (because of StopActivities) that activities are executed regardless of the model state of the IDE. You changed it back.

Hmmm... I did? I don't remember. Do you know which patch/PR that happened in? I am drawing a complete mental blank on that one.

> This is mainly an IntelliJ problem. On Eclipse activities are always executed (causing some issues sometime however when a modal dialog is running the dispatch queue).

The mentioned example is mainly an IntelliJ problem, but a client not responding for a while (for example due to a very slow connection) is still a reasonable scenario, even for eclipse.

srossbach:
> 
> 
> > We have no guideline on how to send file deletions when they are triggered from another delete operation (e.g delete a folder). Unless this is specified somewhere I would not make any assumption which data in which order you get. E.g you could just send a folder delete activity, but I guess this would blow up the partial sharing as it also have to keep track of the current resources.
> 
> @srossbach I know that we _currently_ don't have any guidelines on this. My question was rather whether my understanding of the Eclipse handling (i.e. that child resources are explicitly handled) is correct. From looking at the code, it looks that way to me, but I am not familiar with the details of the Eclipse `IResourceDeltaVisior` that is used in the `ProjectDeltaVisior`. But, from the javadoc, it seems that way as [`visit(...)`](https://github.com/saros-project/saros/blob/master/eclipse/src/saros/project/ProjectDeltaVisitor.java#L65) always returns `true` for folders and the activities should be sorted by the [`sort(...)`](https://github.com/saros-project/saros/blob/master/eclipse/src/saros/project/ProjectDeltaVisitor.java#L306) method.
> 
> As I see no other way of handling this case, I would just make this explicit handling of child resources a new guideline.

It does not always return true. It does not even matter what it returns to understand the logic. After the DeltaVisitor finished its job it contains all resources that were modified in any way and that resources are shared. Returning false simply is a speedup as you do not traverse the tree any deeper.

tobous:
Added the acknowledgment logic. This logic still needs to be tested in a multi-participant session. This will hopefully happen next Monday.

tobous:
Synchronization of access to map of deleted files might still need work. I am still doing some testing.

core/src/saros/concurrent/management/ConcurrentDocumentServer.java
```diff
@@ -53,38 +54,47 @@ public void userLeft(final User user) {
...
+          LOG.debug("Resetting jupiter server for " + resource);
+          server.removePath(resource);
+        };
+
+    this.resourceActivityFilter = new ResourceActivityFilter(sarosSession, deletedFileHandler);
   }
 
   @Override
   public void start() {
+    if (!sarosSession.isHost()) {
```
srossbach: I know this is easy, but why do you not prevent this component being added to the container instead ?

core/src/saros/session/internal/DeletionAcknowledgmentDispatcher.java
```diff
@@ -0,0 +1,90 @@
+package saros.session.internal;
```
srossbach: Any reason to for the session package instead of the concurrent package ?

core/src/saros/session/internal/DeletionAcknowledgmentDispatcher.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Checks whether the passed activity contains the deletion of a file and sends an acknowledgment
+   * to all other participants.
+   *
+   * <p>Activities that contain file deletions are {@link FileActivity file activities} of the type
+   * {@link Type#REMOVED} or {@link Type#MOVED}.
+   *
+   * @param activity the activity to check and acknowledge if applicable
+   */
+  private void acknowledgeDeletionActivity(IActivity activity) {
```
srossbach: I do not see any code that calls this method with anything but just FileActivities.

core/src/saros/concurrent/management/ResourceActivityFilter.java
```diff
@@ -0,0 +1,263 @@
...
+import saros.session.User;
+
+/** Class to handle file deletions and filter out resource activities for already deleted files. */
+class ResourceActivityFilter {
+  private static final Logger log = Logger.getLogger(ResourceActivityFilter.class);
+
+  private final ISarosSession sarosSession;
+
+  /** Method passed by the constructor caller that can be used to react to file deletions. */
+  private final Consumer<SPath> fileDeletionHandler;
```
srossbach: There is only one method implementation. Any reason not pass the ConcurrentDocumentServer and call the specific method instead ?

core/src/saros/concurrent/management/ConcurrentDocumentServer.java
tobous: Because this causes issues with the PicoContainer. I see three possibilities to avoid adding it to the context:
1. Add `null` when not the host.
This causes the session container to throw an exception when initializing the context as `null` values are not accepted.
2. Don't add it at all when not the host.
This causes the session container to also throw an exception as it can no longer resolve all dependencies (even though the logical branches are never used, the logic requiring/using the server still exists on the client side).
3. Write a mock document server that is just a NOP for every method and add it when not the host.
Possible, but I didn't think it would make the situation much better. Also, this would require the introduction of the NOP class as well as an interface for the server and NOP-server to implement.

Or did I miss anything?

core/src/saros/concurrent/management/ResourceActivityFilter.java
tobous: This class is created for both the [`ConcurrentDocumentClient`](https://github.com/saros-project/saros/blob/570e6310f088fe1d13da50e2db90c0ddb4db7e44/core/src/saros/concurrent/management/ConcurrentDocumentClient.java#L40) as well as the [`ConcurrentDocumentServer`](https://github.com/saros-project/saros/blob/570e6310f088fe1d13da50e2db90c0ddb4db7e44/core/src/saros/concurrent/management/ConcurrentDocumentServer.java#L64). The two classes are using different methods, which is why I generalized this.

core/src/saros/session/internal/DeletionAcknowledgmentDispatcher.java
tobous: I was very unsure where to put this. I had it in the concurrent package at first. My reasoning to move it here was that  *technically* the ACKs are not a specific component dealing with concurrency issues, it is just currently the only usage.

But, as I already said, I am unsure on this. If you thing it fits better in the concurrency package, I don't have any issues with moving it there.

core/src/saros/session/internal/DeletionAcknowledgmentDispatcher.java
tobous: You are right. Will change it to just use `FileActivity` as the parameter type. (I wrote this when I wasn't sure whether I would also have to handle folder activities and forgot to update it.)

tobous: Done.

core/src/saros/session/internal/DeletionAcknowledgmentDispatcher.java
tobous: This split on where to put it also stems from me being a tiny bit unhappy with the current solution.

It seems weird to me to filter activities in the concurrency handling classes and not in the `ActivityHandler`, which *handles activities*. This would also allow us to only run one filter on the host side (instead of two with the current solution, one for the document client and one for the document server).

But I couldn't get that solution to work as I am not familiar enough with the inner workings of the activity handler (i.e. how activities are routed). I tried to just add the functionality to the beginning of `handleIncomingActivities(...)` and `handleOutgoingActivities(...)`, but this lead to weird feedback loops (the deletion activities were dropped by the sender as they seemed to be passed through the outgoing method twice).

I tried to debug it but for a while but then just gave up as it isn't that much of an improvement.

core/src/saros/concurrent/management/ConcurrentDocumentServer.java
srossbach: I was interested in answer 2. Looks like the ActivityHandler is not decoupled well enough.

I would suggest either add an TODO here, stating that this behavior is wrong or you can try to add another CTOR to the ActivityHandler (without the Server) which would lead to null checks if the ConcurrentDocumentServer is available in the ActivityHandler.

tobous: There already is a comment about this in the [`SarosCoreSessionContextFactory`](https://github.com/saros-project/saros/blob/master/core/src/saros/session/SarosCoreSessionContextFactory.java#L42). But I can add a comment for it here as well.

In general, I don't think this should be resolved in this PR. I will open an issue for it.

core/src/saros/session/internal/DeletionAcknowledgmentDispatcher.java
srossbach: I have not written the routing, but yes the whole ActivityHandler is a bit tricky as it is coupled with the session. Taking a quick look it goes:

ActivitySequencer -> SarosSession -> ActivityHandler#handleIncoming (when Server then) -> callback (which is the session) -> may call ActivityHandler#handleOutgoing

General rule for activites. Every activity (unless it is a targeted activity) is send to all other users (the generator will never see it unless it is a targeted activity for himself). If the activity is a targeted activity it is send to the host who will "route" it.

srossbach: Seems I had similar issues with this and added a comment. Maybe it is helpful

https://github.com/saros-project/saros/blob/master/core/src/saros/session/internal/SarosSession.java#L751

core/src/saros/concurrent/management/ConcurrentDocumentServer.java
tobous: I created the issue #719 for this. I also added a FIXME to the `start()` method referencing the issue.

core/src/saros/session/internal/DeletionAcknowledgmentDispatcher.java
srossbach: When the host user creates an activity the sequence is: Producer->Session->ActivityHandler#handleOutgoing->Session->ActivitySequencer->Session->ActivityHandler#handleIncoming->Session->ActivitySequencer

Where for every other user the sequence is:
Producer->Session->ActivityHandler#handleOutgoing->Session->ActivitySequencer

tobous: Thanks for the info. But all of this just makes the approach of adding the filter to the activity dispatcher seem way more difficult than the current solution.

Furthermore, I am uncertain whether a shared filter for the client and server on the host instance would work.
My guess is that the current implementation only works as the "first loop" (call to incoming->outgoing) on the host side sets the client filter and the "second loop" sets the server filter.

Because otherwise I can't explain why the current implementation works but my approach for moving it into the `ActivityHandler` did not. The current implementation should run into the same issues as the other approach just simply moves the calls out of the respective client/server method into the `ActivityHandler` method that calls them. So the unification of the client and server filter on the host side is most likely the issue.

Even still, I don't think investing more time into this is the right move as it would require a lot more testing and debugging without providing *that* much of an architectural improvement.

@srossbach What is your opinion? (Also what is your opinion on the placement of the `DeletionAcknowledgementDispacher` with the current approach?)

srossbach: I am too tired to review this fully for now but:

1. Are you aware that a file activity can be a recovery ? Have not checked if this could mess something up.

2. You can play around with the STF that also have some methods where you can suspend and resume network traffic. https://github.com/saros-project/saros/blob/master/stf/src/saros/stf/server/rmi/controlbot/manipulation/INetworkManipulator.java

3. What happens if no acknowledge arrives, e.g a user got disconnected.

tobous: > 1. Are you aware that a file activity can be a recovery ? Have not checked if this could mess something up.

Good point. I have not thought about that. Will check if it matters.

From the top of my head, the recovery stuff might be an issue when the host did not receive a deletion activity. The resulting recovery activity for the client would be a creation activity, which would not be filtered, so the recovery part itself is not the issue. But, as the client filters out all activities for the deleted resource, it would drop the checksum activities used to detect filesystem discrepancies, meaning the desync would never be detected. I think.... (I agree that it is too late for a more detailed analysis right now :wink:)
As I already said, will do more analysis/testing.

> 2. You can play around with the STF that also have some methods where you can suspend and resume network traffic. https://github.com/saros-project/saros/blob/master/stf/src/saros/stf/server/rmi/controlbot/manipulation/INetworkManipulator.java

I will have a look at it. But for that I would first have to get the STF to run. It doesn't seem to play nice with my window manager (it always gets stuck in a loop opening and closing the settings menu). I think there is a way to run it in a docker container, which should resolve the issue, but I haven't had a look at the necessary setup yet.

> 3. What happens if no acknowledge arrives, e.g a user got disconnected.

When the acknowledgment is just dropped/lost, the filter will remain until the file is re-created. But, as this should happen very rarely (without making the whole session unstable), the resulting memory overhead until the end of the session should be acceptable.

When a user leaves the session (or a project is removed from the session), the list of filers is updated accordingly. This is implemented through a [session listener](https://github.com/saros-project/saros/blob/a394fc7754ff362097b16dae0b3804d4b5d80c18/core/src/saros/concurrent/management/ResourceActivityFilter.java#L95).

tobous:
Excluded checksum activities from filter. But this seems to be too strict of an adjustment. Checksum activities acknowledging the file deletion are needed to correctly reset the consistency state after a desync (if the client did not receive a file deletion). Furthermore, these acknowledgements should not hurt as they are only sent after the file was deleted, meaning the vector time will already have been reset correctly, so there is no reason to filter them.

But, if checksum activities for existing files are left through, it can still lead to the situation where the client deletes a file and the host sends a checksum activity before receiving the file deletion, leading to the client thinking it is in an inconsistent state.
I still have to test some more whether dropping such checksum activities is a valid approach or whether it could lead to issues detecting superfluous files. (However, this seems to be an issue in general. From my testing, it seems like such file inconsistencies where one participant has a file that is not present for the other participants is only detected if the corresponding file is open in an editor. Otherwise, no checksum activity will be created for it, meaning the inconsistency will never be detected.)

tobous:
Rebased onto current master and dropped commit adjusting start/stop of `ConcurrentDocumentServer` as it is no longer necessary now that #720 is merged.

tobous:
Adjusted the filter logic to detect checksum activities for existing files as filtered again. This is needed as such activities will still contain the old vector time, meaning they lead to a vector time mismatch exception.

tobous:

tobous:

saros-infrastructure:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- core/src/saros/concurrent/management/ConcurrentDocumentClient.java  1
- core/src/saros/concurrent/management/ResourceActivityFilter.java  6
- core/src/saros/session/internal/DeletionAcknowledgmentDispatcher.java  3
- core/src/saros/concurrent/management/ConcurrentDocumentServer.java  1
- core/src/saros/activities/DeletionAcknowledgmentActivity.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/saros/saros/pullRequest?prid=4251619&bid=14523810)

tobous:
Rebased onto current master without any interaction. Also added the commit `[FIX][CORE] Exclude editor closed activities from filter` to fix the handling of editor closed activities.

saros-infrastructure:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- core/src/saros/concurrent/management/ResourceActivityFilter.java  7
- core/src/saros/concurrent/management/ConcurrentDocumentServer.java  1
- core/src/saros/concurrent/management/ConcurrentDocumentClient.java  1
- core/src/saros/activities/DeletionAcknowledgmentActivity.java  1
- core/src/saros/session/internal/DeletionAcknowledgmentDispatcher.java  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/saros/saros/pullRequest?prid=4251619&bid=14523810)

tobous:
Man, this PR grew a lot from its initial size (mostly because of the addition of the deletion ACK). Now that it has been tested multiple times in a multi-user session and the found issues were resolved, I think it is time to again start reviewing this PR.

@Drakulix @m273d15 @srossbach @stefaus Can I interest anyone of you to again review this PR?

*As a note for the review:*
I think the most crucial point where this logic can actually break something is when activities that are needed for the teardown of the Saros state for the deleted resource are filtered out. This was, for example, the cause for #780 (which has been fixed with https://github.com/saros-project/saros/pull/714/commits/8ed15b5c3a1b26ae72fe91926784b91775355512).

The filter logic is located in [`ResourceActivityFilter#isFiltered(IActivity)](https://github.com/saros-project/saros/blob/8ed15b5c3a1b26ae72fe91926784b91775355512/core/src/saros/concurrent/management/ResourceActivityFilter.java#L253). I went through all existing activities and checked whether they were important for the teardown, but maybe I missed something. So if you see any other possible issues with needed activities being filtered that I might have missed (i.e. that are not handled correctly by the logic), please say so.

saros-infrastructure:


```diff

Issues
======
+ Solved 1
           

Complexity increasing per file
==============================
- core/src/saros/concurrent/management/ConcurrentDocumentClient.java  1
- core/src/saros/activities/DeletionAcknowledgmentActivity.java  1
- core/src/saros/session/internal/DeletionAcknowledgmentDispatcher.java  3
- core/src/saros/concurrent/management/ResourceActivityFilter.java  7
- core/src/saros/concurrent/management/ConcurrentDocumentServer.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/Saros/saros/pullRequest?prid=4251619&bid=14523810)

tobous:
It seems like there is no interest in reviewing (or re-reviewing) this PR. (I want to make very clear that this is **not** meant as any form of criticism, just an observation! I know this PR isn't easy to review and a lot of work went into the review process already.)

But, as I would still like to move the process along, after conferring with @m273d15, I have decided to proceed with merging this PR without a complete review unless anybody voices objections.

I have now tested these changes many times and have removed any outstanding issues. I have also re-run the STF to confirm that all tests (besides the known faulty ones) pass. I feel like I have done my due diligence on that front.

If you don't want that to happen, you have at least till Monday evening (the 28th of January 2020) to either voice some objections/concerns or at least declare your intentions of reviewing this PR (in a somewhat timely manner :wink:). Otherwise, I will merge this some time next week.

tobous:
Rebased onto current master without any interaction in preparation of merge.

PR #718, opened 2019-10-11
Explicitly handle child resources when deleting folders

tobous:
#### [REFACTOR][I] Split resource deletion handling

Split resource deletion handling into separate methods for file and
folder deletions. This was done in preparation for the adjustment of the
folder deletion logic to explicitly handle child resources as this new
logic requires additional calls to the file deletion logic.

Moves the creation of a VirtualFileFilter to separate method. This is
done as it will also be used by the logic iterating the contents of the
deleted folder.

#### [FIX][I] Explicitly send deletion activities for child resources

Adjusts LocalFilesystemModification.generateFolderDeletion(...) to
explicitly create deletion activities for the contained resources. This
ensures that the resource deletion of the child resources is handled
correctly for all other participants. Furthermore, it ensures that the
local state of all deleted child resources is also correctly torn down.

This is done as a fix for an issue related to and discussed in #714.

#### [DOC][I] Add javadoc on the order of child resource activity creation

Amends the javadoc of generateFolderDeletionActivity(...) and
generateFolderMoveActivity(...) to explain the ordering in which the
activities for contained child resources are created/dispatched.

#### [REFACTOR][I] Move file state handling to separate methods

Tries to partly encapsulate the handling of the internal Saros state for
a file in LocalFilesystemModificationHandler by moving it to separate
methods. This makes it easier to uniformly adjust the logic in the
future and offers a base for a possible extraction of the logic in the
future.

### DIFF
--- a/intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
+++ b/intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
@@ -378,6 +378,12 @@ public class LocalFilesystemModificationHandler extends AbstractActivityProducer
   /**
    * Generates and dispatches a folder deletion activity for the deleted folder.
    *
+   * <p>Also explicitly creates deletion activities for all contained resources belonging to the
+   * same module. Contained files are processed first (in ascending depth order), followed by the
+   * contained folders in descending depth order. This ensures that deletion activities for folders
+   * are only send once the activities for all contained resources were sent, allowing the receiver
+   * to also explicitly handle all resource deletions.
+   *
    * @param deletedFolder the folder that was deleted
    */
   private void generateFolderDeletionActivity(@NotNull VirtualFile deletedFolder) {
@@ -507,7 +513,16 @@ public class LocalFilesystemModificationHandler extends AbstractActivityProducer
 
   /**
    * Generates and dispatches matching creation and deletion activities replicating moving the given
-   * directory. Also dispatches activities for all contained resources belonging to the same module.
+   * directory.
+   *
+   * <p>Also explicitly creates activities to move all contained resources belonging to the same
+   * module. When creating the new resources, the creation/move activities are created in ascending
+   * depth order. This ensures that creation/move activities are only send once the creation
+   * activities for all parent resources were sent. When creating the deletion activities, the
+   * contained files are processed first (in ascending depth order), followed by the contained
+   * folders in descending depth order. This ensures that move activities for folders are only send
+   * once the activities for all contained resources were sent, allowing the receiver to also
+   * explicitly handle all resource deletions.
    *
    * <p>How the resources (including the given directory) are handled depends on whether the source
    * and target directory is shared.
--- a/intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
+++ b/intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
@@ -393,13 +393,50 @@ public class LocalFilesystemModificationHandler extends AbstractActivityProducer
 
     User user = session.getLocalUser();
 
-    IActivity activity;
+    Deque<IActivity> queuedDeletionActivities = new ConcurrentLinkedDeque<>();
 
-    // TODO create deletion activities for child resources
-    // TODO clean up editor pool and annotations for child resources
-    activity = new FolderDeletedActivity(user, path);
+    VirtualFileFilter virtualFileFilter = getVirtualFileFilter(deletedFolder);
 
-    dispatchActivity(activity);
+    /*
+     * Defines the actions executed on the base directory and every valid
+     * child resource (defined through the virtualFileFilter).
+     *
+     * Returns whether the content iteration should be continued.
+     */
+    ContentIterator contentIterator =
+        fileOrDir -> {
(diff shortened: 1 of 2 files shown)


intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
```diff
@@ -367,35 +368,103 @@ private void generateResourceDeletionActivity(@NotNull VirtualFileEvent virtualF
...
+        LOG.trace("Ignoring non-shared folder deletion: " + deletedFolder);
       }
 
       return;
     }
 
     User user = session.getLocalUser();
 
-    IActivity activity;
+    Deque<IActivity> queuedDeletionActivities = new ConcurrentLinkedDeque<>();
```
srossbach: Why must this be a concurrent collection ?

tobous: Because I wasn't sure whether there is any kind of parallelization when iterating the resources with the `ContentIterator`. The interface/method descriptions don't make any statement on this, so I wanted to make sure that there aren't any issues with concurrent writes.

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
```diff
@@ -367,35 +368,103 @@ private void generateResourceDeletionActivity(@NotNull VirtualFileEvent virtualF
...
-      annotationManager.removeAnnotations(path.getFile());
+      return;
     }
 
+    User user = session.getLocalUser();
+
+    IActivity activity =
+        new FileActivity(user, Type.REMOVED, FileActivity.Purpose.ACTIVITY, path, null, null, null);
+
+    editorManager.removeAllEditorsForPath(path);
```
srossbach: Although I do not like this option but you could if you like (not this patch) invert the direction, i.e you attach an activity listener from the editorManager and annotationManager to the LocalFilesysteModificationHandler and react on those changes.

The reason I do not like it is because you could create an anonymous producer and so not ever component that fires activities can be observerd.

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
```diff
@@ -367,35 +368,103 @@ private void generateResourceDeletionActivity(@NotNull VirtualFileEvent virtualF
...
+      generateFileDeletion(deletedVirtualFile);
+    }
+  }
+
+  /**
+   * Generates and dispatches a folder deletion activity for the deleted folder.
+   *
+   * @param deletedFolder the folder that was deleted
+   */
+  private void generateFolderDeletion(@NotNull VirtualFile deletedFolder) {
```
srossbach: The naming of the two new methods does not conform to the existing ones in this class.

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
```diff
@@ -367,35 +368,103 @@ private void generateResourceDeletionActivity(@NotNull VirtualFileEvent virtualF
...
+    /*
+     * Calls the above defined contentIterator on the base directory and
+     * every contained resource.
+     * Directories are only stepped into if they match the above defined
+     * virtualFileFilter. This also applies to the base directory.
+     */
+    VfsUtilCore.iterateChildrenRecursively(deletedFolder, virtualFileFilter, contentIterator);
+
+    while (!queuedDeletionActivities.isEmpty()) {
+      dispatchActivity(queuedDeletionActivities.pop());
```
srossbach: Hmm I was going to ask where the activities are sorted until I realized that the activitied are fired during the iteration. Not that easy to spot.

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
tobous: Well, it does match the existing naming convention for the split case (see the move methods), where the methods explicitly handling file/folder logic do not have the 'Activity' suffix.

I created it this way to differentiate between methods that were called by the handler from those that were called 'internally' in the handling logic. But you are right, a uniform naming for all methods is preferable.

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
tobous: I didn't see a reason to also queue file deletions. They are guaranteed to be leaves, meaning they can be processed by the other participants directly without running the risk of deleting child resources in the process.

I extended the javadoc to include information on how the activity order is handled.

tobous:

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
tobous: I am not convinced that such an approach would really improve the code. But I did move all the stuff dealing with the Saros state for a file (e.g. the editor pool and annotations) to separate methods. This makes it easier to uniformly adjust the logic once such an adjustment is necessary. Furthermore, it makes it easier to extract the logic if we were to decide to do so at some point in the future.

Issue #719, opened 2019-10-12
Remove 'ConcurrentDocumentServer' from client session context


tobous:
The `ConcurrentDocumentServer` is only needed/used on the Saros instance that host the session.

However, due to issues with the PicoContainer, it is always added to the session context, even for client instances. This is due to the class being required by the constructor of `ActivityHandler`, meaning the PicoContainer detects an unresolvable dependency when it is not added to the session context.

This came to be an issue with #714, where active components were added to the `ConcurrentDocumentServer`. As a workaround, the `start()` and `stop()` methods were adjusted in order to ensure that the class is only active on the host instance.

#### Suggested solution approach
Adjust the `ActivityHandler` by adding a second constructor that does not require the `ConcurrentDocumentServer`. Subsequently, all parts of the logic that are using the server will then have to do a null-check first (if they are not already doing a is-host-check).

This approach was suggested during the discussion in #714.

tobous:
Fixed by #720.

PR #720, opened 2019-10-12
[FIX] #719 Remove 'ConcurrentDocumentServer' from client session context

srossbach:
Also removed the access through the ISarosSession interface.

### DIFF
--- a/core/src/saros/session/SarosCoreSessionContextFactory.java
+++ b/core/src/saros/session/SarosCoreSessionContextFactory.java
@@ -35,11 +35,8 @@ public class SarosCoreSessionContextFactory implements ISarosSessionContextFacto
   public final void createComponents(ISarosSession session, MutablePicoContainer container) {
 
     // Concurrent Editing
-    /*
-     * As Pico Container complains about null, just add the server even in
-     * client mode as it will not matter because it is not accessed.
-     */
-    container.addComponent(ConcurrentDocumentServer.class);
+    if (session.isHost()) container.addComponent(ConcurrentDocumentServer.class);
+
     container.addComponent(ConcurrentDocumentClient.class);
     container.addComponent(HeartbeatDispatcher.class);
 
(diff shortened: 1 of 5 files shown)


core/src/saros/session/SarosCoreSessionContextFactory.java
```diff
@@ -48,11 +45,10 @@ public final void createComponents(ISarosSession session, MutablePicoContainer c
...
     // Watchdogs
-    // FIXME this should only be added to the host context
-    container.addComponent(ConsistencyWatchdogHandler.class);
-
-    if (session.isHost()) container.addComponent(ConsistencyWatchdogServer.class);
-    else container.addComponent(ConsistencyWatchdogClient.class);
+    if (session.isHost()) {
+      container.addComponent(ConsistencyWatchdogServer.class);
+      container.addComponent(ConsistencyWatchdogHandler.class);
+    } else container.addComponent(ConsistencyWatchdogClient.class);
```
tobous: I would prefer to use the braces for the else case as well to improve readability.

```suggestion
    } else {
      container.addComponent(ConsistencyWatchdogClient.class);
    }
```

core/src/saros/session/internal/ActivityHandler.java
```diff
@@ -383,6 +393,8 @@ public void run() {
    */
   private TransformationResult directServerActivities(List<IActivity> activities) {
 
+    assert session.isHost() && documentServer != null;
```
tobous: Also, you could add an assertion message to make debugging easier in case this ever were to fail.

tobous:
@srossbach Is there a reason not to merge this PR?

srossbach:
No, but I had a heavy migraine that knocked me out.

tobous:
> No, but I had a heavy migraine that knocked me out.

Uh, that sucks. I hope you are feeling better now.

PR #722, opened 2019-10-14
[BUILD] Update to gradle 5.6.2

m273d15:
All changes are generate by:
```
./gradlew wrapper --gradle-version=5.6.2
```

This update fixes the duplicate classpath entries in .classpath [issue](https://github.com/gradle/gradle/issues/10393).


tobous:
The issue does not seem to be related to remaining deprecated build artifacts. Still happens when I clone a fresh version of the Saros repo and delete the workspace before importing. (Setup: Delete Workspace -> Clone -> Check out this branch -> `gradlew prepareEclipse` -> open eclipse -> import from git excluding intellij package)
The issue also doesn't get resolved by subsequently calling `prepareEclipse` or `test` and reloading the package in Eclipse (using F5).

m273d15:
Thank you, I was able to reproduce this issue with a fresh eclipse installation and a fresh clone.

m273d15:
I think switching back to the previous version is the easiest fix of duplicate entries and this issue.
The reason seems to be that gradle introduced a unique on the classpath entries, but groups by the path attribute of a classpathentry. Therefore, the dependencies to saros.eclipse and saros.eclipse:testing are the same in the remove duplicate entries process.

I will open an issue in the gradle project. I don't want to use the old version forever.

PR #726, opened 2019-10-24
Handle resource move corner cases

### LINKED ISSUE #684: Renaming the content root of the shared module removes the entire content for all other participants
When the content root (the actual folder) of the shared module is renamed, it is no longer tracked by Saros. This leads to Saros detecting it to be empty, therefore deleting every shared file of the content root for all other participants.

#### Steps to Recreate
- Alice and Bob start a session
- Alice renames the content root of the shared module

#### Possible Solution Approaches
It might be enough to simply not hold the content root of the shared module in the `IntelliJProjectImpl` and instead always dynamically requesting it when it is required. But, this still leaves the issue that the move of the folder could create activities for all contained files. This issue still needs to be investigated more.

tobous:
#### [INTERNAL][I] Remove static reference to module content root

Replaces the static references to the content root in
IntelliJProjectImpl with dynamic requests. This ensures that the correct
content root is used after it was moved.

#### [FIX][I] #684 Filter move activities for the content root

Adjusts LocalFilesystemModificationHandler to filter out folder move
activities for the content root of the shared module. This allows the
participants to freely move/rename the shared module without causing
desyncs.

Fixes #684.

#### [FIX][I] Filter move activities for module files

Adjusts the logic to not send activities when the module file is moved
back into the shared module.

This logic is currently enough to handle the special case that the
module file of a shared module is moved outside of the shared module.
This is no longer true once multiple modules can be shared at the same
time, as we then also have to handle the case where the module file is
moved into a different shared module. Furthermore, it gets more
complicated when the module file is moved between (different) shared
modules, neither of which are actually associated with the module file,
as this would be seen as a normal file move. Also, modifications made to
the file while it is located in a different shared module would be
shared.

### DIFF
--- a/intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
+++ b/intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
@@ -710,6 +710,11 @@ public class LocalFilesystemModificationHandler extends AbstractActivityProducer
    * @param newFileName the new name for the file or null if it was not renamed
    * @see #generateFolderMoveActivity(VirtualFile, VirtualFile, VirtualFile, String)
    */
+  /*
+   * TODO adjust the logic ignoring module moves once multiple modules can be shared
+   *  Even though module files are ignored by their module, they won't be ignored when moved in/
+   *  between other shared modules.
+   */
   private void generateFileMoveActivity(
       @NotNull VirtualFile oldFile,
       @NotNull VirtualFile oldBaseParent,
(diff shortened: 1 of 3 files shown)


tobous:
Changed base to master and rebased onto current master.

intellij/src/saros/intellij/eventhandler/filesystem/LocalFilesystemModificationHandler.java
```diff
@@ -640,6 +650,31 @@ private void generateFolderMoveActivity(
...
+   *     associated with, <code>false</code> if it represents a file, does not match one of the
+   *     content roots or is not associated with a module
+   */
+  private boolean isContentRootDirectory(@NotNull VirtualFile virtualFile) {
+    if (!virtualFile.isDirectory()) {
+      return false;
+    }
+
+    Module module =
+        Filesystem.runReadAction(() -> ModuleUtil.findModuleForFile(virtualFile, project));
```
stefaus: Nit: Eclipse creates a warning: "The static method should be accessed directly" and would like you to use `ModuleUtilCore.findModuleForFile(virtualFile, project));`

tobous: I am not sure whether the read access is strictly necessary here, but I just wanted to use it to make sure. The call `ModuleUtilCore.findModuleForFile(virtualFile, project))` internally just calls `ProjectFileIndex.SERVICE.getInstance(project).getModuleForFile(file)` which, according to its javadoc, should always be called from a read action. The `ModuleUtilCore` however does not honor this restriction.

PR #737, opened 2019-10-31
[CORE] Network: Reduce Number of TCP Packets

stefaus:
To reduce user input latency, the network socket is configured to send
a new TCP packet each time flush() is called on its provided stream.
The downside is more packets are generated, thus increasing overhead for
additional headers, processing time and thelikelihood of packet loss.

The saros network protocol uses four message types, three of which
describe the payload and one the DATA type carrying the payload. Since
the describing messages are only useful for handling the payload, they
do not need to be sent separately.

### DIFF
--- a/core/src/saros/net/internal/BinaryChannelConnection.java
+++ b/core/src/saros/net/internal/BinaryChannelConnection.java
@@ -220,8 +220,6 @@ public class BinaryChannelConnection implements IByteStreamConnection {
       Integer elementNameId;
 
       synchronized (this) {
-        boolean sendUpdate = false;
-
         final String namespace = data.getNamespace();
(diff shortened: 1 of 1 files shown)


srossbach:
The namespace and element packets are only sent once. The commit message is a bit misleading.

srossbach:
> 
> 
> LGTM, The code is hard to understand. We should refactor this.
> I am not sure whether a `#TODO` is useful or we should try to
> manage these tasks also in the issue tracker.

It might luck ugly because the code also mux and demux. This class was not touched since years, it works fine, does not cause any trouble and just after watching this patch it should be refactored ? If you really want to refactor this class then you should also optimize in such way, that the receiving buffers are reused because the generate much garbage for the garbage collector.

PR #738, opened 2019-11-01
[Server] Introducing ServerReferencePointmanager

pdler:
This PR contains the introducing of the ServerReferencePointManager. It has following features:
- creates an IReferencePoint given by Java.IO.File or Path
- Maps IReferencePoint to Java.Io.File (in this case folder)
- Handling activities, like creating, moving files etc.

The decision of handling activities is, that if an activity is executed by Saros Server, the Saros resources are called directly from the SPath (ex. spath.getFile()). So the idea is, the ServerReferencePointMangaer takes the execution of activities.

### DIFF
--- a/core/src/saros/activities/SPath.java
+++ b/core/src/saros/activities/SPath.java
@@ -96,7 +96,10 @@ public class SPath {
    * Returns a handle for an IFolder represented by this SPath.
    *
    * @return the IFolder contained in the associated IProject for the given project relative path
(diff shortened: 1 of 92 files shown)


pdler:
After rebasing `BasedOnReferencePoints`, this PR got many merge conflict and 40+ commits, so that I created new branch based on the rebased `BasedOnReferencePoints` branch and took on the "new" commits: https://github.com/saros-project/saros/pull/745

PR #741, opened 2019-11-03
Drop re-parsing events

### LINKED ISSUE #740: Randomly closing Tabs
**Describe the bug**
Sometimes when using Saros tabs of CLion randomly close.

**To Reproduce**
Unknown, seems to happen randomly.

**Expected behavior**
The tabs should not randomly close.

**Log files**
[Both log files in one gist](https://gist.github.com/Janrupf/90df33f26162faec584928c618d8b79f)

**Screenshots**
Not applicable

**Environment (please complete the following information):**
 - OS: Arch Linux
 - IDE CLion 2019.2.5
 - Saros Version: [commit 2249ded2e281d92990747064dd5d20e3c93ea5bd](https://github.com/saros-project/saros/commit/2249ded2e281d92990747064dd5d20e3c93ea5bd)

**Additional context**
I built saros from source to allow it to run in CLion. I don't think its CLion which is causing the trouble here, since everything else seems fine.

tobous:
#### [FIX][I] #740 Drop events triggered by file re-parsing

Adjusts the handling of local file property change events to correctly
filter out events triggered by a re-parsing of the file. This explicit
check is needed as such events are reported as re-namings onto the same
name, leading to issues in our logic (as this creates move activities
onto the same resource, resulting in a resource clash on the receiving
side).

This check is currently done by comparing the new and old name. Once
backwards compatibility is dropped below 2019.2, we should consider
using the dedicated method FilePropertyEvent.isRename() for this purpose
instead.

Fixes #740.

#### [INTERNAL][I] Amend JetBrains plugin dependencies

Amends the commented out list of plugin dependencies to adhere to the
plugin SDK guidelines.

The support for other JetBrains IDEs still remains disabled for now.

Even though the plugin now passes a smoke test in many JetBrains IDEs
(CLion, Pycharm, WebStorm), I still don't feel comfortable marking the
plugin as compatible as there still might be implementation differences
(see issue #740, which does not seem to occur in IntelliJ).
Furthermore, the plugin is still in a somewhat unstable state, meaning
we still might make major architectural changes in the future. Enabling
support for other IDEs and therefore guaranteeing compatibility with
these other IDEs increases the testing effort by a lot.

#### [NOP][I] Update links to IntelliJ documentation in plugin.xml HTTPS

### DIFF
--- a/intellij/resources/META-INF/plugin.xml
+++ b/intellij/resources/META-INF/plugin.xml
@@ -35,6 +35,7 @@
 
     <!-- please see http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/plugin_compatibility.html on how to target different products -->
     <!-- uncomment to enable plugin in all products
+    <depends>com.intellij.modules.platform</depends>
     <depends>com.intellij.modules.lang</depends>
     -->
 
--- a/intellij/resources/META-INF/plugin.xml
+++ b/intellij/resources/META-INF/plugin.xml
@@ -30,10 +30,10 @@
         Saros Project
     </vendor>
 
-    <!-- please see http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html for description -->
+    <!-- please see https://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html for description -->
     <idea-version since-build="182.5107.16"/>
 
-    <!-- please see http://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/plugin_compatibility.html on how to target different products -->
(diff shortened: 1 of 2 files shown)


tobous:
Ug, I forgot that the method was only added recently, meaning it breaks backwards compatibility. :weary:

Will have to check what the latest compatible version is and think about whether this is worth it or whether using the more complicated solution is preferable (as it allows us to keep backwards compatibility).

tobous:
This would set the minimum compatibility to `2019.2`, which was released July 2019.

tobous:
I avoided the issue by simply comparing the old and new name (in hindsight, its a pretty obvious solution). This should be sufficient for now, but we should consider using the dedicated method for this check one a backwards compatibility of 2019.2 or newer is more acceptable.

tobous:

Issue #579, opened 2019-07-22
Simplify the execution of stf tests via docker


m273d15:
The docker setup for executing the stf tests in Travis CI can be used to execute the stf tests locally.
The current workflow is not user-friendly and should be simplified in order to reduce the barrier to execute the stf tests.
#333  should be fixed, because the race-condition is likely if executed locally (with better HW).

m273d15:
The current steps in order to execute the tests are:
* Start the required containers and run the following command in the repository root dir:
```bash
export CONFIG_DIR=travis/config SCRIPT_DIR=travis/script/stf
$PWD/travis/script/stf/setup_stf_container.sh $PWD
```
* Run the tests:
```bash
## Start container bash
docker exec -it stf_master bash
## Inside the container:
cd /home/ci/saros_src
./travis/script/stf/master/start_stf_tests.sh
docker network rm stf_test_network
```

After running the tests you might want to remove the following resources (otherwise the next execution of the stf tests (as described above) will fail):
* The `stf_ws` dir
* The running docker containers
* The network `stf_test_network` (`docker network rm stf_test_network`)

PR #742, opened 2019-11-05
Add XStream Security Framework Setup

### LINKED ISSUE #209: Add XStream security framework setup
XStream 1.4.7 and above include a security framework that can be used to white-list certain actions. Such a security framework should be set up to prevent exploits using XStream. More details about the issue are given on the XStream security page: http://x-stream.github.io/security.html

Details (as well as an example) on how to configure the security framework are given in the section http://x-stream.github.io/security.html#framework

To correctly configure this white-list, we will first have to determine the minimal set of permissions need for our use case.

Our XStream provider is defined in the core: [saros/misc/xstream/XStreamExtensionProvider](https://github.com/saros-project/saros/blob/master/core/src/saros/misc/xstream/XStreamExtensionProvider.java)

### XStream version compatibility
We have to ensure that all Saros versions are using compatible XStream versions.

Saros/I is currently relying on the XStream version supplied by IntelliJ (XStream 1.4.10 since IntelliJ 2018.2) as the version contained in the core does not seem to be compatible with Java 8.

The core is currently using a snapshot version of XStream: xstream-1.4-20101008.163445-1.jar
I am not sure which minor release it is, but it is probably a snapshot of the 1.4 development branch as the version 1.4 was released in August 2011 and this name of the snapshot suggests that it is from October 2010. I am not sure whether there is a reason to use this snapshot or if an update would break stuff.

As we are currently working towards moving all Saros components to Java 8, I would suggest to update the XStream version of the core to a proper release above 1.4.7. But I am not sure whether XStream 1.4.7 and above still support Java 6.

### TODOs
- [x] determine XStream version compatibility
  - [x] update the XStream library contained in the core if necessary
- [x] determine the minimal set of needed permissions
- [x] add the security manager setup to the [XStreamExtensionProvider](https://github.com/saros-project/saros/blob/master/core/src/saros/misc/xstream/XStreamExtensionProvider.java)

tobous:
#### [FIX][CORE] #209 Add XStreamFactory to set up security framework

Adds XStreamFactory as a factory class providing XStream objects whose
security framework was already setup to our needs. The security
framework is set up to allow the following objects:
 - null
 - Java primitives and their wrappers
 - Class extending 'Collection'
 - Classes extending 'Map'
 - String
 - Classes in any package starting with 'saros.**'

This should currently allow any existing logic to function as before.

The security framework was introduces to prevent misuse of the
functionality. See https://x-stream.github.io/security.html

Base for fixing #209.

#### [FIX][CORE] #209 Use XStreamFactory to obtain XStream objects

Adjusts the core logic to always use the XStreamFactory to obtain
XStream objects.

This change was also applied to internal usages of XStream that do not
handle any network messages. It is not strictly necessary here but does
not hurt and is still good practice. Furthermore, it prevents the logs
from becoming polluted by warning messages about missing security
framework setups for XStream objects.

This change was also applied to test usages of XStream to ensure that
faulty configurations of the security framework will be reported by the
tests.

Fixes #209.

### DIFF
--- a/core/src/saros/misc/xstream/XStreamFactory.java
+++ b/core/src/saros/misc/xstream/XStreamFactory.java
@@ -1,0 +1,62 @@
+package saros.misc.xstream;
+
+import com.thoughtworks.xstream.XStream;
+import com.thoughtworks.xstream.io.HierarchicalStreamDriver;
+import com.thoughtworks.xstream.security.NoTypePermission;
+import com.thoughtworks.xstream.security.NullPermission;
+import com.thoughtworks.xstream.security.PrimitiveTypePermission;
+import java.util.Collection;
+import java.util.Map;
+
+/** Factory class providing XStream objects with the correctly set-up security framework. */
+public class XStreamFactory {
+
+  /**
+   * Returns an <code>XStream</code> object with an set-up security framework.
+   *
+   * @return an <code>XStream</code> object with an set-up security framework
+   */
+  public static XStream getSecureXStream() {
+    XStream xStream = new XStream();
+
+    setUpSecurityFramework(xStream);
+
+    return xStream;
+  }
+
+  /**
+   * Returns an <code>XStream</code> object with an set-up security framework. The passed <code>
+   * HierarchicalStreamDriver</code> will be used when instantiating the <code>XStream</code> object
+   *
+   * @return an <code>XStream</code> object with an set-up security framework
+   */
+  public static XStream getSecureXStream(HierarchicalStreamDriver hierarchicalStreamDriver) {
+    XStream xStream = new XStream(hierarchicalStreamDriver);
+
+    setUpSecurityFramework(xStream);
+
+    return xStream;
+  }
+
+  /**
+   * Sets up the security framework for the passed <code>XStream</code> object.
+   *
+   * @param xStream the <code>XStream</code> object to set the security framework up for
+   * @see <a
+   *     href="https://x-stream.github.io/security.html">https://x-stream.github.io/security.html</a>
+   */
+  private static void setUpSecurityFramework(XStream xStream) {
+    // forbid all classes by default
+    xStream.addPermission(NoTypePermission.NONE);
+
+    // allow default java stuff
+    xStream.addPermission(NullPermission.NULL);
+    xStream.addPermission(PrimitiveTypePermission.PRIMITIVES);
+    xStream.allowTypeHierarchy(Collection.class);
+    xStream.allowTypeHierarchy(Map.class);
+    xStream.allowTypes(new Class[] {String.class});
+
+    // allow all saros classes
+    xStream.allowTypesByWildcard(new String[] {"saros.**"});
+  }
+}
--- a/core/src/saros/account/XMPPAccountStore.java
+++ b/core/src/saros/account/XMPPAccountStore.java
@@ -31,6 +31,7 @@ import javax.crypto.spec.SecretKeySpec;
 import org.apache.commons.io.IOUtils;
 import org.apache.log4j.Logger;
 import saros.annotations.Component;
+import saros.misc.xstream.XStreamFactory;
 import saros.net.xmpp.JID;
 
 /**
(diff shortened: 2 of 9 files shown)


tobous:
I have not let the STF run with these changes, but did some manual testing for basic session setups in Saros/E and Saros/I. Still, STF test should be done before this is merged.

Will see what I can do (as the STF does not run on my system and I don't know how to run the docker version). If somebody else wants to run it, that would be a nice solution as well. :wink:

core/src/saros/misc/xstream/XStreamFactory.java
```diff
@@ -0,0 +1,62 @@
...
+
+    // allow default java stuff
+    xStream.addPermission(NullPermission.NULL);
+    xStream.addPermission(PrimitiveTypePermission.PRIMITIVES);
+    xStream.allowTypeHierarchy(Collection.class);
+    xStream.allowTypeHierarchy(Map.class);
+    xStream.allowTypes(new Class[] {String.class});
+
+    // allow all saros classes
+    xStream.allowTypesByWildcard(new String[] {"saros.**"});
```
srossbach: In my opinion we should reduce this further to allow only primitive types or special classes that are only containing primitive data

tobous: I think already implemented it that way at first by going through all fields that were annotated with an xsteam alias, but this way seemed cleaner. Especially since it does not require constant updates if you add new field types.

But I will push the other version later today (probably around 1 pm) so that you can have a look at it and compare the two.

m273d15:
@tobous I added the description of the necessary steps to start the stf tests to the issue #579

core/src/saros/misc/xstream/XStreamFactory.java
tobous: As far as I can tell, this is the minimal setup. Furthermore, it requires additional setup in `SPathConverterTest`. The amended PR is available on https://github.com/saros-project/saros/tree/alternative_xstream_security.

```suggestion
    // allow classes needed for Saros network logic
    xStream.allowTypeHierarchy(IActivity.class);
    xStream.allowTypeHierarchy(SarosPacketExtension.class);
    xStream.allowTypeHierarchy(Operation.class);

    Class[] allowedSarosNetworkClasses =
        new Class[] {
          FileList.class,
          JID.class,
          JupiterVectorTime.class,
          ProjectNegotiationData.class,
          SPath.class,
          User.class,
          UserListExtension.UserListEntry.class,
          VersionExchangeExtension.class,
          XStreamPacketExtension.class,
        };
    xStream.allowTypes(allowedSarosNetworkClasses);

    String[] allowedPrivateSarosNetworkClassNames = {
      FileList.class.getCanonicalName() + "$File", FileList.class.getCanonicalName() + "$MetaData"
    };
    xStream.allowTypes(allowedPrivateSarosNetworkClassNames);

    // allow classes needed for ColorIDSetStorage
    Class[] allowedColorIDSetStorageClasses = new Class[] {ColorIDSet.class, UserColorID.class};
    xStream.allowTypes(allowedColorIDSetStorageClasses);

    // allow classed needed for XMPPAccountStore
    Class[] allowedXMPPAccountStoreClasses = new Class[] {XMPPAccount.class};
    xStream.allowTypes(allowedXMPPAccountStoreClasses);

    String[] allowedPrivateXMPPAccountStoreClassName = {
      XMPPAccountStore.class.getCanonicalName() + "$AccountStoreInformation"
    };
    xStream.allowTypes(allowedPrivateXMPPAccountStoreClassName);
```

tobous:
Ok, after a lot of patience I can finally say that both patches pass the STF (except for some flickering and a test that fails on master as well when executed as part of the testsuite). But man, running the STF is annoying.

Now the question is only which version to use. @srossbach Any preferences?

srossbach:
I prefer the simple one for now, as the more detailed my cause issues if people are going to add new activities or packets.

tobous:
As a caveat that I still wanted to mention: Technically, the PR did not strictly "pass" the STF as there were some failing tests. But these test failures are all also present on the current master, so this should not be related to the PR. I created issue #744 for this if anybody wants to have a look.

PR #750, opened 2019-11-08
Prepare for update to Java 11

tobous:
#### [BUILD][CORE] Update testing libraries

Updates EasyMock and PowerMock to the newest compatible versions. This
was done to keep the number of issues related to java 11 to a minimum.

#### [FIX] Add PowerMockIgnore annotations for javax.xml.*

These annotations are necessary to avoid an internal issue with
powermock. This workaround will most likely become unnecessary at some
point in the future.

#### [FIX] Make SarosSessionManagerTest$DummyError package-protected

This was done to avoid an issue with the CTOR accessibility occurring
when using Java 11.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -91,10 +91,10 @@ configure(projectsToConfigure) { projectToConf ->
   dependencies {
     addLog4JDependency(configurations.testConfig)
     testConfig junitVersion
-    testConfig 'org.easymock:easymock:3.5.1'
-    testConfig 'org.powermock:powermock-core:2.0.0'
-    testConfig 'org.powermock:powermock-module-junit4:2.0.0'
-    testConfig 'org.powermock:powermock-api-easymock:2.0.0'
+    testConfig 'org.easymock:easymock:4.0.1'
+    testConfig 'org.powermock:powermock-core:2.0.5'
+    testConfig 'org.powermock:powermock-module-junit4:2.0.5'
+    testConfig 'org.powermock:powermock-api-easymock:2.0.5'
   }
 
   // generate lib directory that contains all release dependencies
--- a/core/test/junit/saros/session/SarosSessionManagerTest.java
+++ b/core/test/junit/saros/session/SarosSessionManagerTest.java
@@ -28,7 +28,7 @@ import saros.session.internal.SarosSession;
 @PowerMockIgnore({"javax.xml.*"})
 public class SarosSessionManagerTest {
 
-  private class DummyError extends Error {
+  class DummyError extends Error {
     private static final long serialVersionUID = 1L;
   }
 
--- a/core/test/junit/saros/session/SarosSessionManagerTest.java
+++ b/core/test/junit/saros/session/SarosSessionManagerTest.java
@@ -11,6 +11,7 @@ import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.powermock.api.easymock.PowerMock;
+import org.powermock.core.classloader.annotations.PowerMockIgnore;
 import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.PowerMockRunner;
 import saros.communication.connection.ConnectionHandler;
@@ -24,6 +25,7 @@ import saros.session.internal.SarosSession;
 
 @RunWith(PowerMockRunner.class)
 @PrepareForTest({SarosSession.class, SarosSessionManager.class})
+@PowerMockIgnore({"javax.xml.*"})
 public class SarosSessionManagerTest {
 
   private class DummyError extends Error {
--- a/eclipse/test/junit/saros/SarosEclipseContextFactoryTest.java
+++ b/eclipse/test/junit/saros/SarosEclipseContextFactoryTest.java
@@ -5,6 +5,7 @@ import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.powermock.core.classloader.annotations.MockPolicy;
+import org.powermock.core.classloader.annotations.PowerMockIgnore;
 import org.powermock.core.classloader.annotations.PrepareForTest;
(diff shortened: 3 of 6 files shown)


srossbach:
Do you need Java 11 for IntelliJ or what is the purpose of bumping the Java version that high ?

tobous:
This PR still needs some work. The STF needs to be adjusted to run on newer Eclipse versions. Subsequently, we need to define the new lowest compatible Eclipse version.
This should be done before this PR is merged so that the STF can be run for it to make sure that the update doesn't introduce any unwanted side effects.

Also, we need to update the build infrastructure to use the new Java and Eclipse versions. The build will fail in the current setup.

srossbach:
I do not want to disappoint you but even the newest version of Eclipse only required Java 8. You cannot bump the used Java version of the core higher than that.

tobous:
> Do you need Java 11 for IntelliJ or what is the purpose of bumping the Java version that high ?

No, I don't need 11.

But I think it is time for an update and Java 11 is the next LTS release after 8 (yay Oracle support plan). Furthermore, anything between Java 8 and Java 11 is currently already out of support.

From what I have seen/heard from other projects, the jump from Java 8 to 11 is the norm when updating from Java 8.

I already had a look at this a while back, but the PoweMock support was still way to buggy. It still has issue as of release 2.0.4 (https://github.com/saros-project/saros/pull/750/commits/8ca8dea46e21e2932b884c50c1c621770529b145), but it seemed stable enough for us to use it.

srossbach:
The Oracle "support" plan is not that easy to understand because the support is no longer free.

tobous:
> The Oracle "support" plan is not that easy to understand because the support is no longer free.

I was also surprised that JDK 8 still has non-commercial support. I through that also ended long ago. But hey, at least it's something.

I still think that we should update to JDK 11 if possible. Updating won't get easier with time and we don't want to again end up in the situation where we missed the update window and have to deal with annoying issues when updating (like 6->7 and 7->8).



> I do not want to disappoint you but even the newest version of Eclipse only requires Java 8. You cannot bump the used Java version of the core higher than that.

Well that sucks. But that was probably to be expected. IntelliJ also only somewhat recently switched to Java 11 (from Java 8). Do you know whether there are any plans on switching to Java 11 for Eclipse? I couldn't find anything.

srossbach:
I did not find anything either regarding Eclipse but you can still update the IntelliJ version to Java 11 if you like.

srossbach:
You could also build the core against different Java Version but good luck that it will never require to change any line of production code. Also this is confusing if someone wants to use some Java 11 features and gets a big disappointment if the CI compiles the code.

tobous:
> I did not find anything either regarding Eclipse but you can still update the IntelliJ version to Java 11 if you like.

I will think about it. But I am not sure whether splitting the builds again is a good idea. And, as already mentioned, the update isn't strictly necessary, I just don't want us to fall behind too far. So this will probably just have to wait until Eclipse updates or there is a real (urgent) reason to update. But now that we have this PR, the update itself shouldn't be an issue. At that point, we may also no longer need the PowerMock workarounds, so that's at least something 😉.

srossbach:
I do not see any problems running Saros/I with JDK 11. Saros/I was running with JDK 7 over a year while the Core and Saros/E was still using JDK 6.

tobous:
> I do not see any problems running Saros/I with JDK 11. Saros/I was running with JDK 7 over a year while the Core and Saros/E was still using JDK 6.

I am not sure whether that is a good idea yet. After you mentioned Eclipse releases, I had another look at IntelliJ releases. Even though there are Java 11 builds (and my linux distro provides only those builds), there still are Java 8 releases as well. I am not sure what the "default" version is, but either way, updating now would split the user base.

m273d15:
If the new libraries etc. are still compatible with JDK 8 we could think about using a build matrix with JDK 8 and JDK 11 in the master branch. We would execute the build and the unit tests with both JDKs. Nobody would try to use JDK 11 features which are not available (because the used JDK is 8), but we could check that our dependencies etc. are still ready for JDK 11.

tobous:
@m273d15 Are you still interested in adding a multi-build that runs periodically to check whether we keep java 11 compatibility? If so, I can update this PR and remove the actual plugin config stuff (basically just remove 00dda58e69a60976f5a76e7273af549ce6ee7c70).

As already mentioned, the changes in the other PRs are just workarounds for issues in Powermock that might become obsolete in the future. But they don't really hurt and are easy to remove once they are no longer needed.

tobous:
Adjusted PR to only prepare for migration to Java 11 without actually changing the configuration.

This was done to allow us to implement a check whether we could theoretically build using java 11.

tobous:
I also tested whether anything has changes with the update to PowerMock 2.0.5. But we still need the workarounds introduced in this PR.

tobous:
Theoretically, the powermock ignores added in https://github.com/saros-project/saros/pull/750/commits/558718423775148a77ddaa07d739fec7d15cbdec can be replaces with a [global powermock ignore](https://github.com/powermock/powermock/wiki/PowerMock-Configuration#global-powermockignore) in the core test. This would then apply to all tests that include the core test classpath.

The advantage of this approach is that there is only a single location where the configuration is made and where we can add a comment that it is only a workaround.

The disadvantage, on the other hand, is that the workaround is then somewhat hidden. As this is only defined in a separate directory in the core, it is harder to understand why the tests in other packages behave in a certain way.

This would be an issue if we ever were to add tests that actively want to mock classes in `java.xml.*`, which would clash with the global powermock ignore for such packages.

I am not sure which approach to take. @m273d15 @srossbach @stefaus Any opinions?

m273d15:
I think the annotation is fine. The hidden configurations might lead to confusion.

tobous:
Rebased onto current master and added references to the matching issues in PowerMock to the workaround commits.

@m273d15 Any remaining issues with this PR?

PR #752, opened 2019-11-14
Always close editor when file is deleted

### LINKED ISSUE #758: Dirty editor is not always closed when file is deleted
Editors for shared files are not always closed when the corresponding file is deleted while it is being modified by a session participant. Instead, it remains open in a dirty state. When trying to save the editor, the IDE correctly reports that there is no local representation and asks where to create the file.

#### Steps to reproduce

- Start a session between Alice and Bob, both running on different machines
- Alice and Bob open an editor for the same file 
- Alice creates continuous uninterrupted inputs in the editor (i.e. holds down a keyboard key)
- Bob deletes the shared file for the editor
(Weirdly enough, Eclipse does not always let you deleted files with a dirty editor. This only seems to be possible after the file was saved at least once locally (remote saves do not count) after it was created.)

##### Expected behavior

The editor is closed for both participants.

##### Actual behavior

Alice or Bob (somewhat inconsistent) still has a dirty open editor for the deleted file, even though it no longer exists on disk. This sometimes also triggers a session inconsistency.

tobous:
#### [INTERNAL][E] Add logic to close editor without saving

Preparation for fixing #758.

#### [FIX][E] #758 Always close editor when file is deleted

Adjusts the logic handling file deletion actions on the sending and
receiving side. Ensures that the editor for the deleted file is always
closed properly without saving.

Skipping the save is important as the editor will only still be open for
the deleted file if it was in an unsaved state at the time of deletion.

Closing such unsaved editors for deleted files is important as they
otherwise create issues with the consistency logic. Since the watchdog
logic is based on the open editors, the remaining editor is interpreted
as the resource still being present, leading to an inconsistent state.

Fixes #758.

### DIFF
--- a/eclipse/src/saros/project/ProjectDeltaVisitor.java
+++ b/eclipse/src/saros/project/ProjectDeltaVisitor.java
@@ -220,6 +220,8 @@ final class ProjectDeltaVisitor implements IResourceDeltaVisitor {
   private void generateRemoved(IResource resource) {
 
     if (resource instanceof IFile) {
+      editorManager.closeEditor(new SPath(ResourceAdapterFactory.create(resource)), false);
+
       addActivity(
           new FileActivity(
               user,
--- a/eclipse/src/saros/editor/EditorManager.java
+++ b/eclipse/src/saros/editor/EditorManager.java
@@ -755,6 +755,20 @@ public class EditorManager implements IEditorManager {
 
   @Override
   public void closeEditor(final SPath path) {
+    closeEditor(path, true);
+  }
+
+  /**
+   * Closes the editor for the given resource. If <code>save=true</code> is specified, the content
+   * of the editor will be written to disk before it is closed.
+   *
+   * <p>Does nothing if the resource is not shared or there is no open editor for the resource.
+   *
+   * @param path the resource whose editor to close
+   * @param save whether to write the editor content to disk before closing it
+   * @see EditorAPI#closeEditor(IEditorPart, boolean)
+   */
+  public void closeEditor(final SPath path, boolean save) {
     if (path == null) throw new IllegalArgumentException("path must not be null");
 
     SWTUtils.runSafeSWTSync(
@@ -763,7 +777,7 @@ public class EditorManager implements IEditorManager {
           @Override
           public void run() {
             for (IEditorPart part : editorPool.getEditors(path)) {
-              EditorAPI.closeEditor(part);
+              EditorAPI.closeEditor(part, save);
             }
           }
         });
--- a/eclipse/src/saros/editor/internal/EditorAPI.java
+++ b/eclipse/src/saros/editor/internal/EditorAPI.java
@@ -184,18 +184,23 @@ public class EditorAPI {
   }
 
   /**
-   * Closes the given editor part.
+   * Closes the given editor part. If <code>save=true</code> is specified, the content of the editor
+   * will be written to disk before it is closed.
    *
    * <p>Needs to be called from an UI thread.
+   *
+   * @param part the editor part to close
+   * @param save whether to write the editor content to disk before closing it
+   * @see IWorkbenchPage#closeEditor(IEditorPart, boolean)
    */
-  public static void closeEditor(IEditorPart part) {
+  public static void closeEditor(IEditorPart part, boolean save) {
     IWorkbenchWindow window = getActiveWindow();
 
     if (window == null) return;
 
     IWorkbenchPage page = window.getActivePage();
     // Close AND let user decide if saving is necessary
-    page.closeEditor(part, true);
+    page.closeEditor(part, save);
   }
 
   /**
--- a/eclipse/src/saros/project/FileActivityConsumer.java
+++ b/eclipse/src/saros/project/FileActivityConsumer.java
@@ -141,7 +141,11 @@ public class FileActivityConsumer extends AbstractActivityConsumer implements St
   }
 
   private void handleFileDeletion(FileActivity activity) throws CoreException {
-    final IFile file = toEclipseIFile(activity.getPath().getFile());
+    SPath path = activity.getPath();
+
+    editorManager.closeEditor(path, false);
+
+    final IFile file = toEclipseIFile(path.getFile());
 
     if (file.exists()) FileUtils.delete(file);
     else LOG.warn("could not delete file " + file + " because it does not exist");
--- a/eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
+++ b/eclipse/src/saros/ui/wizards/AddProjectToSessionWizard.java
@@ -309,7 +309,8 @@ public class AddProjectToSessionWizard extends Wizard {
                       /*
                        * close all editors to avoid any conflicts.
                        */
-                      for (final IEditorPart editor : editorsToClose) EditorAPI.closeEditor(editor);
+                      for (final IEditorPart editor : editorsToClose)
+                        EditorAPI.closeEditor(editor, true);
                     }
                   });
 


eclipse/src/saros/project/ProjectDeltaVisitor.java
```diff
@@ -220,6 +220,8 @@ private void generateMoved(
   private void generateRemoved(IResource resource) {
 
     if (resource instanceof IFile) {
+      editorManager.closeEditor(new SPath(ResourceAdapterFactory.create(resource)), false);
```
srossbach: As I side note although it is not obvious. The code executed here should (regarding the Eclipse documentation) not do heavy work, we already do heavy work here and even sync on the SWT thread which is deadlock prone (it currently does not happen). I do not know what happens if you invoke Eclipse logic here.

And the other question ? When does this even happen. If I delete a file that is dirty (the editor) Eclipse closes the editor. So this can just happen when this is automated, e.g it could caused by another plugin (maybe by accident).

tobous: > As I side note although it is not obvious. The code executed here should (regarding the Eclipse documentation) not do heavy work, we already do heavy work here and even sync on the SWT thread which is deadlock prone (it currently does not happen). I do not know what happens if you invoke Eclipse logic here.

I added the logic here as it was the obvious spot to put it. I am not familiar with the Eclipse event listener logic. Do you know of a better way of reacting to a local file deletion which allows me to do "heavy work".

> And the other question ? When does this even happen. If I delete a file that is dirty (the editor) Eclipse closes the editor.

My assumption was that it happens when eclipse encounters an editor on deletion that is still dirty *after* the warning dialog for a dirty editor (i.e. the dialog saves the editor to allow for a clean deletion but our activities make it dirty again before the actual deletion takes place). But I am not completely sure. An "easy" test setup that let me reproduce the issue consistently is:
- Start a session between Alice and Bob, both running on different machines
- Both open an editor for the same file
- Alice creates continuous uninterrupted inputs in the editor (i.e. holds down a keyboard key)
- Bob deletes the shared file for the editor
(Weirdly enough, Eclipse does not always let you deleted files with a dirty editor. This only seems to be possible after the file was saved at least once locally (remote saves do not count) after it was created.)

The result of this is that Alice or Bob (somewhat inconsistent) still has a dirty open editor for the deleted file, even though it no longer exists on disk.

This behavior no longer appears with this workaround.

> So this can just happen when this is automated, e.g it could caused by another plugin (maybe by accident).

My local Eclipse installation is just the default "Eclipse IDE for Eclipse Contributes" (release 2019-09). I don't have anything else installed besides "GEF Legacy". So even if this issue is caused by another plugin, it is most likely a pre-installed one (at least for the contributes edition), meaning we should handle the issue either way in my opinion.

tobous:
I created the issue #758 for this and mentioned it in the commit messages to make it easier to understand issue/reasoning behind the commit when looking at the commit later on.

eclipse/src/saros/project/ProjectDeltaVisitor.java
srossbach: The scenario that you described should be fixed with the modification you made to FileActivityConsumer. I do not recommend to close editors in the ProjectDeltaVisitor as you do not know what caused the deletion of the file.

tobous: > The scenario that you described should be fixed with the modification you made to FileActivityConsumer.

But isn't the `FileActivityConsumer` only responsible for incoming deletion activities? I distinctly remember also having the same issue on the side of the participant deleting the resource. (On the other hand, I could not recreate this during my recent tests. Only the receiver got the issue, which would match your suggestion.)

I will do some more testing with and without this part of the fix and report back.

tobous: I can confirm that the issue still occurs on host side after removing the added line from `ProjectDeltaVisitor`. (Will try again with the line and update the comment.)

Sadly, I haven't found a way of consistently triggering the issue. With the described setup, it only works about 30-40% of the time.

I had a look at the logs but couldn't find anything out of the ordinary besides some jupiter vector time errors (some of which were caused by the faulty heartbeat logic; also fixed by #714). But I only looked at loglevel warning and above (as Saros/E logs way to much as trace is enabled by default for the devel build).

tobous: Ok, got it to work first try. (GitHub does not allow me to upload files right now, so I will try again later on.)

Test setup was as follows:

Both using Eclipse for Contributors 2019-09, building Saros from source.
Alice: user `intellij_test_2`, archlinux
Bob: user `intellij_test_1`, Windows 10

- Alice (host) and Bob start a session on an existing project.
- Alice creates the file `ToDeleteClass`.
- Alice makes some changes (inserting whitespace) and saves the file.
- Bob opens the editor for `ToDeleteClass`.
- Bob creates continuous inputs by holding down a letter key on the keyboard.
- Alice selects file in project view, presses delete and selects continue option in dialog.

Result:
Editor for file `ToDeleteClass` is still open for Alice after file was deleted on disk. When IDE regains focus, Alice is asked whether to save or drop the editor contents. (Still, this leads to errors when the other user is in follow mode.)

srossbach: Looking at the log:

TRACE 01:26:41,720 [saros-activity-dispatcher] (ActivityHandler.java:373) dispatching [#3] optimized activities [mode = 0] : [TextSelectionActivity(offset: 0, length: 0, src: intellij_test_1@saros-con.imp.fu-berlin.de/Saros, path: SPath [project=P/TestProject (EclipseProjectImpl), projectRelativePath=src/pkg2/plan]), ViewportActivity(path: SPath [project=P/TestProject (EclipseProjectImpl), projectRelativePath=src/pkg2/plan], range: (0,1)), EditorActivity(type: CLOSED, path: SPath [project=P/TestProject (EclipseProjectImpl), projectRelativePath=src/pkg1/ToDeleteClass.java])]
TRACE 01:26:41,730 [Worker-10: Building workspace] (FileSystemChecksumCache.java:85) invalidating checksum for existing file: /TestProject/bin/pkg1/ToDeleteClass.class [0xACAC1FFE1EA25B4031585DB7F89C9510]
TRACE 01:26:41,730 [Worker-10: Building workspace] (SharedResourcesManager.java:206) received resource change event caused by job  ='Building workspace'
TRACE 01:26:41,730 [Worker-10: Building workspace] (SharedResourcesManager.java:218) received resource delta 'ResourceDelta(/)' contains:
C(0) /
C(0) /TestProject
C(0) * /TestProject/bin
C(0) * /TestProject/bin/pkg1
R /TestProject/bin/pkg1/ToDeleteClass.class

TRACE 01:26:41,730 [Worker-10: Building workspace] (SharedResourcesManager.java:260) generated resource activities for current resource delta 'ResourceDelta(/)' : []
TRACE 01:26:41,734 [BinaryChannel-intellij_test_1] (BinaryChannelConnection.java:308) processing opcode 0xFA [TFD]: id=0, chunks=1
TRACE 01:26:41,734 [BinaryChannel-intellij_test_1] (BinaryChannelConnection.java:346) processing opcode 0xFB [DATA]: id=0, DATA len=245 bytes
TRACE 01:26:41,734 [saros-DispatchContext] (NetworkManipulatorImpl.java:101) intercepting incoming packet from: intellij_test_1@saros-con.imp.fu-berlin.de/Saros
TRACE 01:26:41,735 [saros-DispatchContext] (XMPPReceiver.java:214) received binary XMPP extension: TransferDescription [elementName=ados, namespace=saros, recipient=intellij_test_2@saros-con.imp.fu-berlin.de/Saros, sender=intellij_test_1@saros-con.imp.fu-berlin.de/Saros, compress=true], size: 245, RX time: 0 ms [SOCKS5 (D)]
TRACE 01:26:41,735 [saros-DispatchContext] (ActivitySequencer.java:571) rcvd (001) intellij_test_1@saros-con.imp.fu-berlin.de/Saros -> [JupiterActivity(timestamp: [0,0], operation: Insert(0,'d',0), source: intellij_test_1@saros-con.imp.fu-berlin.de/Saros)]
TRACE 01:26:41,750 [main] (RemoteWriteAccessManager.java:168) .updateConnectionState(SPath [project=P/TestProject (EclipseProjectImpl), projectRelativePath=src/pkg1/ToDeleteClass.java])

It seems that the the file is deleted before the the Editor Close Activity is executed resulting in a document provider still attached to that file.

srossbach: My big guess is that it has something to do with the RemoteWriteAccessManager keeping references on the DocumentProviders although the file is no longer present.

srossbach: Looking at the log:

delete file
-> send delete file
<- receive editor closed
execute documentprovider disconnect which results in releasing the handle to the buffer

Just wondering what will happen if we release the handle(s) before the file is deleted. I guess this would also fix some stuff regarding the watchdog logic, e.g imagine you send a file delete but never receive and editor closed for this activity, the buffer is still read.

tobous: Ok, I am not really sure what to do with that information. Could you provide a patch/PR that should fix the underlying issue? If you prefer, I can then still test it for you (or we can test/debug it together).

srossbach: After thinking about this issue again it seems this is currently the best option to solve this problem. It also ensures that an EditorClosedActivity is send before the FileDeletionActivity is executed. Lets hope it do not cause any deadlocks.

tobous: Would spawning another thread and executing the editor close on it help, just to not run it in the thread the `ProjectDeltaVisitor` is run on?

This would remove the "synchronous" part of the execution, meaning the editor closed activity would not necessarily be send before the deletion activity. But I don't know how much the exact timing/order of the two activities matters.

tobous: > Would spawning another thread and executing the editor close on it help, just to not run it in the thread the `ProjectDeltaVisitor` is run on?

I will merge this for now. If you do prefer a solution utilizing another thread, I can still add it in a separate PR.

PR #756, opened 2019-11-15
[DOC] Correct CSS selector for gap between tab and content

fzieris:
The old selector `.tab-content :first-child` matched *all* elements that are the first child of their respective parent, such as the first list items in all lists. Furthermore, the intended padding was only added for the Eclipse tab (which was such a `:first-child`) but not the IntelliJ tab (which is always the second child, after the Eclipse tab).


fzieris:
@m273d15 Feel free to merge this 😃

PR #759, opened 2019-11-17
[DOC][BUILD] Add information about Windows drive restrictions

tobous:


srossbach:
I guesses you mean drive. A partition (the disk one) can contain multiple drives.

tobous:
> I guesses you mean drive. A partition (the disk one) can contain multiple drives.

You are right. Will adjust the wording.

PR #760, opened 2019-11-17
[DOC] Set title in head

tobous:
Sets the page or site title as the title in the HEAD section of the
generated HTML site. This allows the title to be displayed correctly in
the browser tab name.


tobous:
Instead of setting the title ourselves, we could also consider using other jekyll plugin [jekyll-seo-tag](https://github.com/jekyll/jekyll-seo-tag), which sets the title and other data used for search engine optimization. (It is on the [list of allowed jekyll plugins](https://pages.github.com/versions/) for github pages.)

PR #764, opened 2019-11-19
[BUILD] Add nightly Saros/I release for all IDEs

m273d15:
* Adds a GitHub Action that releases the state of
  current master branch as Saros/I release which
  also supports all JetBrains IDEs.
* Action is triggered each day at 08:00.
* The action uncomments the <depends> tags in plugin.xml.
  Therefore, the commenting style is changed.

This approach uses an existing release and always releases the new release as asset of the release. The approach has the advantage that releases are not created and deleted automatically (which leads to a chaotic GitHub release page).

I created the PR, because in #740 a user already asked for a build that is installable in CLion, @Drakulix and me are interested in a PyCharm build.

This is still a draft, because I want to clarify the structure of the release entry. My proposal:

-------------------------------------------
## Saros/I nightly
This is a nightly Saros/I release that is also installable in any other JetBrains IDE.

Please, be aware that:
  * We are currently **testing only with the IntelliJ IDEA**. Therefore, **don't hesitate to [contact us](https://gitter.im/saros-project/saros/user) or [report issues](https://github.com/saros-project/saros/issues/new/choose)** you encounter with Saros/I (especially in other IDEs).
  * You and your participant **use the same Saros/I nightly release**.

`This is a pre-release`: true<br/>
`tag`: current master (disadvantage of the approach, the tag is contant)


m273d15:
I know that the uncommenting logic is not the perfect solution, but I assume that it is good enough. We don't change this file very frequently. Futhermore, we would notice issue at the next day (I would introduce a corresponding badge in the readme).

.github/workflows/nightly-rel.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Build Saros/I
+        run: ./gradlew sarosIntellij
+          
+      - name: Deploy Saros/I Release
+        uses: WebFreak001/deploy-nightly@v1.0.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # automatically provided by github actions
+        with:
+          upload_url: <TODO> # find out this value by opening https://api.github.com/repos/<owner>/<repo>/releases in your browser and copy the full "upload_url" value including the {?name,label} part
```
tobous: What about these two TODOs? Or are they dependent on the actual URL of the release entry?

m273d15: Yes this URL defines the upload URL to the specific release. The of is the release identifier.

tobous:
As seen in #824, we can access the sha of the used commit when executing the action. So it should be possible to use it for the version number. This makes sure that nightly build based on different commits are not compatible (as this could introduce a lot of unexpected issues otherwise).

For the version number, I would prefer something along the lines of `[MAJOR].[MINOR].[PATCH].NIGHTLY-[COMMIT_SHA]` (i.e. for a nightly based on the [current master](https://github.com/saros-project/saros/commit/ee5457752c255f8b9ec31705a792ee02e9ec2264), this would be `0.2.2-NIGHTLY-ee5457752c255f8b9ec31705a792ee02e9ec2264`). An abbreviation of the commit sha would also be possible/useful (see [`--abbrev-commit`](https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection)) to not make the version string too long.

As this fits in the current Saros versioning scheme (`major.minor.micro.qualifier`), this would also not cause any issues.

*Update:* Ok, the current version manager completely ignores the "qulifier" part of the version. Yeay. So it makes no difference with the current version comparison logic. I would have to agree with @stefaus that the current version comparison logic is way overbuild. It is completely irrelevant which version is newer/older. Just comparing versions and deciding based on that (e.g. solely on MAJOR version) should be enough.

And I don't see a big reason to provide that much compatibility support anyways. There aren't a lot of reasons not to update the plugin once a new version is out. And updating takes seconds.

m273d15:
PR #824 will already contain the functionality.

Issue #744, opened 2019-11-06
Failing STF tests on master


tobous:
On the current master, there are multiple tests that consistently fail. Some of them could be related to my local setup, so I grouped them into two categories:

#### ~~saros.stf.test.account.AccountPreferenceTest#testRemoveAccountButton()~~

~~The STF test fails when its run as part of the test class (AccountPreferenceTest) but succeeds when it is run on its own.~~

**Update:** Fixed by #747.

<details><summary>Stacktrace</summary>

```
AccountPreferenceTest
saros.stf.test.account.AccountPreferenceTest
testRemoveAccountButton(saros.stf.test.account.AccountPreferenceTest)
org.eclipse.swtbot.swt.finder.widgets.TimeoutException: Timeout after: 5000 ms.: The widget (of type 'Button' and with mnemonic 'Finish' and with style 'SWT.PUSH') was not enabled.
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.waitUntil(SWTBotFactory.java:522)
	at org.eclipse.swtbot.swt.finder.SWTBot.waitUntil(SWTBot.java:1)
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.waitUntil(SWTBotFactory.java:496)
	at org.eclipse.swtbot.swt.finder.SWTBot.waitUntil(SWTBot.java:1)
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.waitUntil(SWTBotFactory.java:484)
	at org.eclipse.swtbot.swt.finder.SWTBot.waitUntil(SWTBot.java:1)
	at org.eclipse.swtbot.swt.finder.widgets.AbstractSWTBot.waitForEnabled(AbstractSWTBot.java:694)
	at org.eclipse.swtbot.swt.finder.widgets.SWTBotButton.click(SWTBotButton.java:66)
	at saros.stf.server.rmi.superbot.impl.SuperBot.confirmShellAddXMPPAccount(SuperBot.java:220)
	at saros.stf.server.rmi.superbot.component.menubar.menu.submenu.impl.SarosPreferences.addAccount(SarosPreferences.java:63)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #765, opened 2019-11-19
[BUILD] Execute stf tests on pr/stf/* branches

m273d15:
* Introduce GitHub action in order to execute
  the stf tests on branches that improve the
  stf tests.
* The travis stf runs terminate after ~45min
  and don't allow to run the tests completely


srossbach:
Any changes to only execute modified or new tests instead of all tests ?

m273d15:
No I want to create a setup that allows to modify the tests in pr/stf branches.
The Travis setup terminates the test and only runs nightly. Follow-up PRs will change the set of stf test to be executed

.github/workflows/stf.yml
```diff
@@ -0,0 +1,33 @@
+name: STF Tests
+
+on:
+  push:
+    # Temporary restriction: Run only stf tests on branches that work on stf
+    branches:    
+      - 'pr/stf/*'
+
+# Restriction when STFs ha stable results
```
tobous:
```suggestion
# Restriction when STF has stable results
```

m273d15:
> The action fails with an exception regarding the yaml file:

I fixed this with a direct rebase push. The indention of the file was broken. It is the parse that failed not the action. The [GitHub documentation](https://help.github.com/en/github/setting-up-and-managing-organizations-and-teams/repository-permission-levels-for-an-organization#repository-access-for-each-permission-level) says that you need at least "write" access in order to create/edit an action.

tobous:
> The [GitHub documentation](https://help.github.com/en/github/setting-up-and-managing-organizations-and-teams/repository-permission-levels-for-an-organization#repository-access-for-each-permission-level) says that you need at least "write" access in order to create/edit an action.

Ok, good.

> I fixed this with a direct rebase push. The indentation of the file was broken. It is the parse that failed not the action.

Is there a way to re-trigger the action so that we can make sure that it runs correctly before merging this?

m273d15:
I created the branch `pr/stf/test_gha` with the same commits. Therefore, the result is also published on this PR, because the same commit is build. The result: STF GitHub Action badge is as expected, because of #744

.github/workflows/stf.yml
tobous: Still unresolved.

tobous:
It still feels kind of weird adding a check that we know will always fail in the current setup, meaning the builds that trigger them will always fail (until all issues with the current STF tests are resolved). But I guess that's how it'll be for now. At least having any job that runs the STF completely is an improvement.

PR #767, opened 2019-11-20
Use ConnectionHandler instead of XMPPConnectionService

stefaus:
`XMPPConnectionService` and its Listener `IConnectionListener` expose smack objects. To avoid this in code where they are not needed the `ConnectionHandler` and `IConnectionStateListener` is used (as the javadoc from `ConnectionHandler` already advises) .

--
Handler and Listener may also expose smack exception objects which will be changed in future PRs.


core/src/saros/communication/connection/ConnectionHandler.java
```diff
@@ -113,6 +113,16 @@ public String getConnectionID() {
     return jid.toString();
   }
 
+  /**
+   * Returns the local JID for the current connection.
+   *
+   * @return the local JID for the current connection or <code>null</code> if no connection is
+   *     established or the connection has no jid
+   */
+  public JID getLocalJID() {
```
srossbach: This basically violates the design. This class should hide the any internals. What is the problem to use getConnectID and handle this in the affected places ?

stefaus: I know, but as the whole UI is currently violating the whole design I added this.
For example `OpenChatAction` is comparing JIDs, using `getConnectID()` this would lead to jid -> string -> jid conversions without a benefit because in the end everything its still limited to using jids.

I could add a `@deprecated` remark, but I think that this is superfluous because all JID usage in UI should be deprecated.

So yes this can look like a little step back, but from my point of view its better to do so for now. My current approach is to focus on moving smack stuff to core and than removing all this JID refs. Are you ok with this?

saros-infrastructure:


```diff

Issues
======
- Added 2
           

Complexity decreasing per file
==============================
+ eclipse/src/saros/ui/widgets/viewer/session/XMPPSessionDisplayComposite.java  -1
+ eclipse/src/saros/ui/actions/OpenChatAction.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/saros/saros/pullRequest?prid=4536207&bid=15264454)

PR #770, opened 2019-11-22
Scale images correctly when viewed on mobile

tobous:
#### [DOC][FIX] Scale images correctly when viewed on mobile

Adjusts some remaining pictures to scale correctly when viewed on
mobile.

#### [DOC][NOP] Fix minor formatting of heading definitions

### DIFF
--- a/docs/contribute/processes/review.md
+++ b/docs/contribute/processes/review.md
@@ -126,7 +126,7 @@ In case you don't know yet:
 
 #### Which merge option should be used?
 
-![GitHub merge options](../images/processes/GitHub_choose_merge_method.png)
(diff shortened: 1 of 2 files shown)


tobous:
@m273d15 Is there an option to apply this option side-wide, so that it doesn't have to be set every time a picture is added? I think you mentioned something about that a while back but had some reservation, but I don't know what exactly the issue was.

m273d15:
It should be possible to add the class to all images (it might be a problem that the bootstrap classes are loaded via the header link and the other styling is defined via scss. Therefore, I am not sure, whether the SCSS can access the bootstrap classes) , but IIRC this change has undesired effects on the logo and small icons, but if you like we can try it.

PR #788, opened 2019-12-12
Fix #571

### LINKED ISSUE #571: STF contains old paths
The [STF documentation](https://www.saros-project.org/contribute/saros-testing-framework.html) contains old paths:
* the location of the `configuration.properties` file 
* the location of the launch configs

m273d15:


docs/contribute/saros-testing-framework.md
```diff
@@ -47,7 +47,7 @@ Therefore the current workaround is to execute the Gradle task `generateLibAll`
 ### Run Tests
 
 1.  Start the **requiered launch configurations** which are located
-    in the directory `test/resources/launch` by **right clicking the
+    in the directory `stf/launch` by **right clicking the
```
srossbach: Last time I used a launch configuration I had multiple, one in the STF and one in the Saros/E plugin. Is this still the case ?

tobous: Yes. The second one in `stf/launch` was introduced when the stf was moved into the separate package `stf`. As far as I know, the second set of launch configs are just copies of the ones located in the eclipse package.

We already discussed this in a standup a while back and I though about deleting one of the sets, but I wasn't quite sure which one.

As the launch configs seem to be primarily aimed at the STF execution, it would make sense to keep these ones and delete the ones in the `eclipse` package. However, due to ease of use, I quite often (exclusively, to be honest) use the stf launch config to test Saros/E. So removing them from the `eclipse` package could reduce the "visibility" of the launch configs for testing purposes (primarily for new developers).

srossbach: I always use them because of the different log level and some other goodies (different XMPP account store)

tobous: @srossbach Do you see any issue with removing the launch config from the eclipse package (located in `saros/eclipse/test/resources/launch/`)?

srossbach: I really do not care as long as the developers will provide proper log files with the full log level. So yes just remove them.

Maybe we can make this clear (have not read the documentation yet) that using such launch configuration does not run any tests but only allows you to access the remote control functionality offered by the STF. We should also describe some changes that are made to Saros when executing the plugin in the mode. 

See https://github.com/saros-project/saros/blob/master/stf/src/saros/stf/server/STFController.java#L93

Basically it switches the account store (which caused issues in the HTML GUI 😆 ) and sets an option for auto stopping empty sessions without asking the user.

tobous: > Basically it switches the account store (which caused issues in the HTML GUI  ) and sets an option for auto stopping empty sessions without asking the user.

Huh, I though that was just the default :laughing:. I guess I really didn't use the normal Saros/E config all that often.

> Maybe we can make this clear (have not read the documentation yet) that using such launch configuration does not run any tests but only allows you to access the remote control functionality offered by the STF. We should also describe some changes that are made to Saros when executing the plugin in the mode.

I don't think we have any information on manual testing in Saros/E on our website. From having a short look at the "Contribute" sections, I could only find the page on [STF testing](https://www.saros-project.org/contribute/saros-testing-framework.html), which does not mention any specifics regarding the launch configs.

But yes, that would be a useful section to have.

m273d15: Opened #791

PR #824, opened 2020-01-28
Build and test via GitHub Actions

m273d15:
This PR disables the Travis CI usage and uses GitHub Actions for:
* Build and test
  * Upload build results if build and tests are successful
  * Upload test results if the tests failed
  * Don't build if only documentation is changed

The artifacts are [stored for 90 days](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/persisting-workflow-data-using-artifacts#about-workflow-artifacts).

Attention: This setup does not use the `ci_build` image as the Travis version. Therefore, the used IntelliJ version is only defined by the intellij gradle plugin.

In a follow up PR I will fix:
* rename the dir `travis` to `ci`
* remove superfluous scripts


m273d15:
Example [artifacts](https://github.com/saros-project/saros/actions/runs/32227845) of this PR. (We can use this feature to upload the screenshots and reports of the STF tests)

tobous:
@m273d15 Is there any way of manually re-starting jobs? From what I have seen with the recent Travis builds, the downloads of the formatter are quite flaky, meaning it would be nice to have such an option.

I couldn't find a corresponding button in the UI. Or can this only be done through other means? Or only for failed builds?

tobous:
Nevermind: https://stackoverflow.com/questions/56435547/how-do-i-re-run-github-actions

m273d15:
>  the downloads of the formatter are quite flaky

Therefore, the used action caches the jar in the GitHub Action for the next run.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+
+  formatter:
+    runs-on: ubuntu-latest
+    name: Check java format
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+    - uses: m273d15/java-format-action@v1.1-rc
```
tobous: The old formatting action also contained a linter for javascript (ESLint). This is no longer the case with this action.

I am not sure what exactly the linter did an whether it can adequately be replaced through codacy rules (which also has ESLint integration).

Even though this is currently not necessary, with the re-introduction of javascript (or rather typescript) components through the VSCode implementation, this will become useful again.

docs/contribute/processes/continuous-integration.md
```diff
@@ -13,8 +13,8 @@ The results of the analysis are available through the GitHub interface or can be
 
 ## Pull-Request-Triggered Jobs
 
-Travis CI also runs when a new pull request is created or an existing pull request is updated.
-See the [Travis documentation](https://docs.travis-ci.com/user/pull-requests/#%E2%80%98Double-builds%E2%80%99-on-pull-requests) for more information.
+GitHub Actions also runs when a new pull request is created or an existing pull request is updated.
```
tobous: There are still more references to Travis CI builds that need to be updated in the section `Push-Triggered Jobs`.

It is also mentioned in the section `Daily Jobs`, but, As the CRON job would still be active after this, such references can remain as far as I can tell.

README.md
```diff
@@ -1,5 +1,5 @@
 # <img src="docs/assets/images/logo/saros-logo.svg" alt="drawing" align="right" width="100"/> Saros - Distributed Party Programming
-[![Travis](https://travis-ci.com/saros-project/saros.svg?branch=master)](https://travis-ci.com/saros-project/saros/branches)
+[![Build](https://github.com/saros-project/saros/workflows/Build/badge.svg)](https://github.com/saros-project/saros/actions?query=workflow%3ABuild)
```
tobous: Is this supposed to display builds for all branches?

With the previous badge, only builds of the master were displayed. Builds for other branches/PRs can routinely break without having any impact (as they might be WIP). Displaying build results for them could leave a bad impression of the state of the repo/development when visiting the README.

The [documentation](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository) offers options on how to do this:
```suggestion
[![Build](https://github.com/saros-project/saros/workflows/Build/badge.svg?branch=master)](https://github.com/saros-project/saros/actions?query=workflow%3ABuild+branch%3Amaster)
```

tobous:
Also, once this is merged, we mustn't forget to update the branch protection rules to configure (one of) the GitHub action builds as necessary.

tobous:
With the IntelliJ version now only being set through the properties file, we don't have any way of guaranteeing backwards compatibility. So it might be a good idea to actually use the [intellij plugin verifyer](https://github.com/JetBrains/intellij-plugin-verifier) as you already suggested.

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+        java-version: 1.8
+
+    # Build
+    - name: Build and test with Gradle
+      run: ./gradlew -PskipTestSuites=true -PaggregateTestResults=true -DmaxParallelForks=4 --no-daemon --parallel sarosEclipse sarosIntellij sarosServer
+
+    # Post-Build
+    # Success-Case
+    - name: Upload eclipse build
+      uses: actions/upload-artifact@v1
```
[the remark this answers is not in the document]
m273d15: I adjusted the upload logic and only upload the artifacts if the build is successful and on the master branch.

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- buildSrc/src/main/java/saros/gradle/eclipse/SarosEclipsePlugin.java  1
- buildSrc/src/main/java/saros/gradle/eclipse/configurator/EclipseVersionConfigurator.java  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=4901837&bid=16228737)

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+      if: success()
+      with:
+        name: saros-server-${{ github.sha }}
+        path: build/distribution/server
+
+    # Failure-Case
+    - name: Check test report
+      id: check-report
+      if: failure()
+      run: |
```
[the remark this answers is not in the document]
m273d15: I use a multi-line command because the first char "[" is reserved and the [yaml parser has issues](https://github.com/m273d15/saros/actions/runs/38705953) with the command and I don't want to quote the whole command.

.github/workflows/build.yml
m273d15: We have to think about how we want to integrate eslint, but there is a github action for executing eslint and codacy also provides a mechanism to execute eslint with a config in your repository.

m273d15:
I still have to check:
* whether the missing `<version>` tag causes warnings in the intellij dev environment
* which checks in GitHub should be required (because the build action is not triggered if only the documentation is changed)

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- buildSrc/src/main/java/saros/gradle/eclipse/SarosEclipsePlugin.java  1
- buildSrc/src/main/java/saros/gradle/eclipse/configurator/EclipseVersionConfigurator.java  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=4901837&bid=16228737)

build.gradle
```diff
@@ -108,6 +108,7 @@ configure(projectsToConfigure) { projectToConf ->
 
   projectToConf.ext.intellijHome = projectToConf.hasProperty('intellijHome') ? intellijHome : System.getenv('INTELLIJ_HOME')
   projectToConf.ext.intellijSandboxDir = System.getenv('SAROS_INTELLIJ_SANDBOX')
+  projectToConf.ext.versionQualifier = projectToConf.hasProperty('versionQualifier') ? '.' + versionQualifier : ''
```
tobous: Does the qualifier count as "set" if it is defined as empty? I.e. does this change mean that the versions will always end in a '.' as the property `versionQualifier` is set by default (even though it is set to be empty).

*Update:* This is in fact the case. You can't start a local instance using the `runIde` task using this branch as it fails due to an illegal plugin version modifier `0.2.2.`, meaning the trailing `.` is included when no version qualifier is set.

tobous: Removing the default entry in `gradle.properties` resolves this issue.

.github/workflows/build.yml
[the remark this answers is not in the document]
tobous: The only issues is that we now can't test the uploaded artifacts (which changed as part of this PR update) without merging this into the master :sweat_smile:. So we should either adjusts this for testing or create a separate testing branch that uploads the artifacts for all branches.

build.gradle
tobous: We could consider using "DEV" as a default to make it easier to determine whether we are working on a dev build or not (and make it easier to tell if user report issues on their own build).

The issue with that is that I don't know whether you can "un-set" a parameter. We would still need a way of building artifacts with the normal version for release. So that should be a possibility without having to modify the source (i.e. removing the `gradle.properties` entry).

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- buildSrc/src/main/java/saros/gradle/eclipse/SarosEclipsePlugin.java  1
- buildSrc/src/main/java/saros/gradle/eclipse/configurator/EclipseVersionConfigurator.java  3
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=4901837&bid=16228737)

tobous:
When building from source using Eclipse and having a default version qualifier set, the core manifest is modified permanently to include the modifier. If this is the wanted behavior, this should be documented somewhere.

Furthermore, the actual version of the Eclipse plugin is not updated correctly. It is still shown as `15.0.0` in the running eclipse instance.

tobous:
And the automatic scripts cause the last empty line of the core manifest to be dropped for whatever reason....

build.gradle
m273d15: Thank you :+1:

> The issue with that is that I don't know whether you can "un-set" a parameter.

You can call `./gradlew -PversionQualifier= sarosEclipse` looks weird but results in an empty property value.

m273d15:
> When building from source using Eclipse and having a default version qualifier set, the core manifest is modified permanently to include the modifier. If this is the wanted behavior, this should be documented somewhere.

I never used the PR with a default qualifier, because I created the logic for publishing nightly builds with qualifiers and tested that version is used in drop-in and zip.

I will test and enhance the logic in order to support a default qualifier.

tobous:
> And the automatic scripts cause the last empty line of the core manifest to be dropped for whatever reason.... (at least for me, the line was removed after I build and ran Saros/I and Saros/E).

This issues is caused when running the `runIde` task, not the Eclipse build.

Also, as another issue, when the manifest was already modiefied (i.e. already contains a version along the lines of `0.1.0.DEV`), it is no longer found by the update logic, meaning subsequent changes to the `versionQualifier` value are no longer honored by the build plugin/resulting artifact.

.github/workflows/build.yml
```diff
@@ -0,0 +1,101 @@
...
+        sed -i 's/<!--depends/<depends/g' intellij/resources/META-INF/plugin.xml
+        sed -i 's/depends-->/depends>/g' intellij/resources/META-INF/plugin.xml
+
+    # Build
+    - name: Build and test with Gradle
+      run: |
+        ./gradlew \
+          -PskipTestSuites=true \
+          -PaggregateTestResults=true \
+          -PversionQualifier=$GITHUB_SHA \
```
tobous: How about including a "DEV" (or "NIGHTLY", but I think I would prefer "DEV" in this situation; a matter of taste I guess) suffix to make it more clear that the build is a developer build. In that context, the appended SHA will also make more sense.

Also, as mentioned before, is there a way of sensibly shortening the SHA? Otherwise, the version string will be absurdly long.

```suggestion
          -PversionQualifier=DEV-$GITHUB_SHA \
```

m273d15: > Also, as mentioned before, is there a way of sensibly shortening the SHA? Otherwise, the version string will be absurdly long.

I could try `git rev-parse --short <sha>`.

> How about including a "DEV".

Sounds reasonable

tobous: > I could try `git rev-parse --short <sha>`.

Sounds good.

tobous:
Also, your current regex also won't work for the STF (and STF self-test) as it still has the version `0.1.0.DEVEL`. But that should not be an issue as the `DEVEL` should have already been dropped as part of the move to semantic versioning anyways.

m273d15:
> Also, as another issue, when the manifest was already modified (i.e. already contains a version along the lines of 0.1.0.DEV), it is no longer found by the update logic, meaning subsequent changes to the versionQualifier value are no longer honored by the build plugin/resulting artifact.

The PR was based on the assumptions (which reduced the complexity):
* We use this logic only in CI for creating artifacts with qualifiers (We do not use qualifiers in our repository).

Therefore, It was not important to preserve the formatting of the metafiles (because the formatting completely changes during the jar creation process) and overwriting qualifiers was not necessary.

With the idea of using a default qualifier in the dev process I have to rewrite the whole logic, because this is a feature that was not intended. I have to consider:
* If the qualifier property is empty and the metafiles contain a qualifier -> remove the qualifier and the leading '.'
* If the qualifier property is filled and the metafiles contain a qualifier -> overwrite the existing qualifier
* Use a more specific regex
* Preserve the formatting
* Check integration in the development processes

I would like to adjust the qualifier (DEV-<short sha>) and add the default qualifier handling in a follow-up PR.

buildSrc/src/main/java/saros/gradle/eclipse/configurator/EclipseVersionConfigurator.java
```diff
@@ -0,0 +1,50 @@
...
+  private void replaceLines(String regex, String replacement) {
+    try {
+      Path path = manifest.toPath();
+      Charset charset = StandardCharsets.UTF_8;
+      List<String> oldContentLines = Files.readAllLines(path, charset);
+      String newContent =
+          oldContentLines
+              .stream()
+              .map(line -> line.replaceAll(regex, replacement))
+              .collect(Collectors.joining("\n"));
```
tobous: This causes the newline at the end of the file to be dropped. A dirty fix would be to just add it back manually.

```suggestion
              .collect(Collectors.joining("\n"));

      newContent += "\n";
```

buildSrc/src/main/java/saros/gradle/eclipse/SarosEclipsePlugin.java
```diff
@@ -24,7 +27,34 @@ public void apply(Project project) {
...
+    project.afterEvaluate(
+        p -> {
+          configureEclipseAfterEvaluate(p, e);
+          createPluginVersionChangeTask(p, e);
+        });
+  }
+
+  private void createPluginVersionChangeTask(Project p, SarosEclipseExtension e) {
+    String qualifier = e.getPluginVersionQualifier();
+    if (e.getPluginVersionQualifier() == null || qualifier.isEmpty()) return;
```
stefaus: Nit:
```suggestion
    if (qualifier == null || qualifier.trim().isEmpty()) return;
```

m273d15:
Example of a run on [master](https://github.com/m273d15/saros/actions/runs/38825739)
Example of a run on [not_master](https://github.com/m273d15/saros/actions/runs/38833420)

tobous:
Also, the naming of the build artifact for Saros/I is pretty generic. The downloadable ZIP just contains another zip called "Saros.zip".

Is it possible to set a more descriptive name? At least `saros-intellij.zip` would be nice. I know this is kind of nit-picking as the main zip is named sensibly, it still seems confusing to the just have a "Saros.zip" lying around.

.github/workflows/build.yml
```diff
@@ -0,0 +1,107 @@
...
+          --parallel \
+          sarosEclipse sarosIntellij sarosServer sarosLsp
+
+    # Post-Build
+    # Success-Case
+    - name: Upload eclipse build
+      uses: actions/upload-artifact@v1
+      if: success() && github.ref == 'refs/heads/master'
+      with:
+        name: saros-eclipse-dev-${{ steps.short-sha.outputs.short_sha }}
```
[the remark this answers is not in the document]
m273d15: I don't think so:
* SHORT_SHA is an environment variable I set in the runner environment
* short_sha is a step output that is accessible in the configuration. Using an environment variable in the config is not possible (as far as I know).

m273d15:
> Is it possible to set a more descriptive name? At least saros-intellij.zip would be nice. I know this is kind of nit-picking as the main zip is named sensibly, it still seems confusing to the just have a "Saros.zip" lying around.

Sure, this is an issue independent of the PR, but we can do this. The name is currently defined in the [custom plugin](https://github.com/saros-project/saros/blob/b411c88e1a7687507263708b381b5a82b602c8f9/buildSrc/src/main/java/saros/gradle/intellij/SarosIntellijExtension.java#L24) (I will remove this configuration. The custom plugin should not wrap the whole IntelliJ plugin and we should use the original gradle intelliJ plugin whenever possible.).
However, we should check whether this option also changes the plugin.xml.

m273d15:
> Also, the naming of the build artifact for Saros/I is pretty generic. The downloadable ZIP just contains another zip called "Saros.zip".

I think the jar file's name is currently the same as the plugin name defined in the

tobous:
From what I can tell, changing the `pluginName` in `SarosIntellijExtension` does not influence the `plugin.xml` of the build result. The fields `name` and `id` still remain the same. So it only seems to influence zip name. But I can't guarantee that as I am not familiar with the used API.

m273d15:
Renamed the zip in #832

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- buildSrc/src/main/java/saros/gradle/eclipse/configurator/OsgiBundleVersionConfigurator.java  3
- buildSrc/src/main/java/saros/gradle/eclipse/SarosEclipsePlugin.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=4901837&bid=16228737)

PR #829, opened 2020-02-12
[FIX][DOC] Use SVG logos for website

tobous:
#### [FIX][DOC] Use SVG logos for website

Replaces the PNG versions of the used logos with SVG versions to provide
scalability.

Manually sets the size of the IDE icons in the tab headers to have a
height of 30. This was done to match the old height as this is required
by the current layout. Otherwise, there are gaps in the tab frame.

IntelliJ logo was take from official site:
https://www.jetbrains.com/company/brand/logos/

Usage should be cleared under brand guidelines:
https://www.jetbrains.com/company/brand/

Eclipse logo was taken from official site:
https://www.eclipse.org/artwork/

Text was removed and graphic was cropped using Inkscape. I hope this
does not count as "modifying the logo" and is therefore still covered by
the usage guidelines:
https://www.eclipse.org/legal/logo_guidelines.php

However, if this is not covered by the usage guidelines, the existing
usage would not be covered either. So this change should not hurt either
way.

#### [FIX][DOC] Replace base page comic with higher res version

Sets the width to 400 to preserve the old layout. As the image is still
not that large and not scalable, keeping it relatively small seems like
the best option.

It is still not quite clear where the image came from and whether the
usage of base material taken from the "Geek and Poke" comics was cleared
with the author. So we might have to look for a better image to use for
our base page.


tobous:
This still leaves the [initial comic on the home page](https://github.com/saros-project/saros/blob/master/docs/assets/images/comics/dpp_0.png). But I couldn't find the initial source we used to create it. @m273d15 Any idea?

Also, the ["Marketplace" button](https://marketplace.eclipse.org/sites/all/themes/solstice/public/images/marketplace/btn-install.png) for the Eclipse installation is very low res on the [release page](https://raw.githubusercontent.com/saros-project/saros/master/docs/releases/index.md). But that is the only image provided by Eclipse I could find. And it is still the one linked to by the official market place when creating a new "external button".

m273d15:
> This still leaves the initial comic on the home page. But I couldn't find the initial source we used to create it. @m273d15 Any idea?

No it looks like it is created based on [geek and poke](http://geek-and-poke.com/)

PR #831, opened 2020-02-14
Remove Whiteboard

m273d15:
The current whiteboard is buggy and the HTML WB has the same
issues as the [HTML GUI](https://www.saros-project.org/contribute/html-gui.html).
We have no resources to maintain this approach.

Follow-up todos:
* remove corresponding GitHub label
* close corresponding issues


eclipse/plugin.xml
```diff
@@ -1469,12 +1469,4 @@
          </adapter>
       </factory>
    </extension>
-   <extension
```
srossbach: This is not BUILD related.

m273d15: True, I will move the change into another commit.

tobous:
As the whiteboard build files were previously covered by the gitignore, they are not automatically removed with this PR, meaning developers will still have to remove the `whiteboard` folder and its remaining contents themselves after checking out the newest changes (if they ever built the whiteboard locally before).

I don't know whether this is really necessary, but you could consider writing a short note on the mailing list or developer gitter after this is merged that some minor local cleanup is still necessary, just to avoid confusion.

tobous:
You could consider adding an entry to the website pointing to the commit still containing the code in case somebody is interested in the whiteboard at some point in the future.

If I remember correctly, you also wanted to include the issues with the whiteboard with that. If you don't want to do that with this PR, you could just open an issue to remind us that the entry still has to be written.

PR #835, opened 2020-02-18
[BUILD] Avoid redundant jars in intellij zip

### LINKED ISSUE #830: Saros/I build artifact contain duplicate library jars
When looking at the [Saros build artifacts](https://github.com/m273d15/saros/actions/runs/38825739), I noticed that the artifact for Saros/I is considerably larger than the one for Saros/E or the Saros Server (8mb vs 5mb).  

This seems to stem from many libraries (like XStream, Smack, Commons libraries, our re-packaged picocontainer, etc.) being included twice. Once in the general zip and once in the contained `saros.core.jar`.

This isn't a huge issue as 8mb still isn't much, but it would still be nice to avoid such duplication.

m273d15:
Fix #830

I tested:
* local execution
* import and build in IntelliJ
* import and build in Eclipse + launch config

### DIFF
--- a/core/build.gradle
+++ b/core/build.gradle
@@ -1,3 +1,7 @@
+configurations {
+  plain
+}
+
 sarosEclipse {
   manifest = file('META-INF/MANIFEST.MF')
   createBundleJar = true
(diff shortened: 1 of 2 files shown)


m273d15:
This leads to a renamed `saros.core.jar` -> `saros.core-plain.jar` in the zip file.

m273d15:
> The plain jar (for Saros/I) still contains the old meta file stating the contained and exported JARs that are actually no longer present.

True, we can think about providing a simple manifest (without the osgi stuff, just the name and version) that is created by gradle as we already do in the server and lsp build

Issue #315, opened 2018-11-26
Build documentation in CI


m273d15:
Currently the documentation is only build if a PR is merged into the master branch. A change of a jekyll config (as `_config.yml`) could break this build and an invalid state would be on the master branch.
Therefore we should:
* Build **also** the documentation if a PR changes something in `docs` and somewhere else
* Build **only** the documentation if a PR changes only something in `docs`

m273d15:
There are multiple GitHub action (see [here](https://github.com/marketplace?utf8=%E2%9C%93&type=actions&query=github+pages)). We have to identify an action that:
* is maintained
* allows building the pages (not build+deploy in one step)

m273d15:
Done, documentation commits do trigger a build and deploy GitHub Action

PR #840, opened 2020-02-21
[BUILD] Build documentation

m273d15:
#315 Build documentation via GitHub pages.
In order to avoid that we introduce broken
documentation into the master branch, this
workflow builds the jekyll pages.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,6 @@ jobs:
 
   formatter:
     runs-on: ubuntu-latest
-    name: Check java format
     steps:
     - name: Checkout
       uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


tobous:
If we also want to make the java format check required (which I am definitely for), we also have to introduce an equivalent task for the doc builds. In order to do this, we should also rename the task to something more general like "Check format" (instead of "Check java format").

What exactly the "check format" job does for doc builds, I don't know. If we can't think of anything useful, we can still just make it a NOP.

m273d15:
> What exactly the "check format" job does for doc builds, I don't know. If we can't think of anything useful, we can still just make it a NOP.

There are markdown checkers, but I don't want to enforce documentation formatting. I would say NOP.

tobous:
Besides the NOP job to replace the java formatter for doc builds, I don't see any issues with this PR.

@m273d15 Do you plan on addressing the formatter in this PR or a separate PR?

tobous:
Ok, the GitHub API seems to have some issues right now, which is why the actions aren't triggered.

m273d15:
Thank you for your additions :1st_place_medal:

PR #841, opened 2020-02-21
[DOC] Update screenshot

### LINKED ISSUE #166: Rework Screenshots documentation
Is the [current screenshot](http://saros-project.github.io/saros/documentation/screenshots.html) page useful?

If yes:
* Create new screenshots
* Adjust the image size so that the image is not bigger than the page body width (As now).

m273d15:
Used image from #212 (It
was not possible to re-use the PR)

fix #166


m273d15:
> Compared to the previous image, it is missing the second "4" next to the left viewport annotations. The viewport annotations are displayed on the left side and on the right side of the editor.

I will check whether this annotation still exists in the current version.

> More importantly, this is missing the text update that was also part of the original PR. The current text no longer represents the actual functionality.

I will add the text

tobous:
The information in the original PR was not correct as well. The needed adjustments for the current text are along the lines of:

Adjustment for `1.`: double-sided arrow marks files that are shared. files visible to a user (editor open and selected) are marked with a dot in the user color. dot is gray if visible to multiple users.

Adjustment for `5.` also shows which files the user has currently open and follow mode information (not 100% sure what exactly is shown).

srossbach:
The user is not displayed as: You

This was removed ages ago.

PR #844, opened 2020-02-25
Refactor EDT handling

tobous:
This PR is mainly a refactoring, so there isn't a lot of interest happening. The main thing is moving the EDT runner logic from `Filesystem` to the new class `EDTExecutor` and then replacing direct calls to the EDT logic with calls to the new class.

Furthermore, a minor issue is fixed where calls in`ProjectAPI` were not run from the EDT even though they were required to.

#### [REFACTOR][I] Rename Filesystem to FilesystemRunner

Renames Filesystem to FilesystemRunner to better represent the
functionality of the class.

Subsequently moves it to the runtime package.

#### [INTERNAL][I] Add EDTExecutor

Adds the class EDTExecutor offering methods to synchronously or
asynchronously run tasks on the event dispatcher thread (EDT).

This class was added to unify the logic needing access to such
capabilities. A followup commit will replace all current usage of the
direct logic to run tasks on the EDT with calls to the EDTExecutor.

#### [REFACTOR][I] Use EDTExecutor

Replaces all direct calls to Application.invokeAndWait(...) and
Application.invokeLater(...) with calls to the EDTExecutor.

Removes a couple of execution blocks for code that does not need to be
executed on the EDT.

#### [FIX][I] Execute code in EDT where necessary in ProjectAPI methods

Fixes some methods of the ProjectAPI that did not correctly run certain
calls on the EDT as required by their documentations.

Subsequently removes a now unused method from the FilesystemRunner that
was previously erroneously used to run code on the EDT.

### DIFF
--- a/intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java
+++ b/intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java
@@ -1,6 +1,5 @@
 package saros.core.ui.eventhandler;
 
-import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.application.ModalityState;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.wm.WindowManager;
@@ -10,6 +9,7 @@ import org.apache.log4j.Logger;
 import saros.core.monitoring.IStatus;
 import saros.core.monitoring.Status;
 import saros.intellij.SarosComponent;
+import saros.intellij.runtime.EDTExecutor;
 import saros.intellij.runtime.UIMonitoredJob;
 import saros.intellij.ui.Messages;
 import saros.intellij.ui.util.NotificationPanel;
@@ -78,30 +78,28 @@ public class NegotiationHandler implements INegotiationHandler {
   private void showIncomingInvitationUI(
       Project project, final IncomingSessionNegotiation negotiation) {
 
-    ApplicationManager.getApplication()
-        .invokeLater(
-            () -> {
-              JoinSessionWizard wizard =
-                  new JoinSessionWizard(project, getWindow(project), negotiation);
-              wizard.setModal(true);
-              wizard.open();
-            },
-            ModalityState.defaultModalityState());
+    EDTExecutor.invokeLater(
+        () -> {
+          JoinSessionWizard wizard =
+              new JoinSessionWizard(project, getWindow(project), negotiation);
+          wizard.setModal(true);
+          wizard.open();
+        },
+        ModalityState.defaultModalityState());
   }
 
   private void showIncomingProjectUI(
       Project project, final AbstractIncomingProjectNegotiation negotiation) {
 
-    ApplicationManager.getApplication()
-        .invokeLater(
-            () -> {
-              AddProjectToSessionWizard wizard =
-                  new AddProjectToSessionWizard(project, getWindow(project), negotiation);
+    EDTExecutor.invokeLater(
+        () -> {
+          AddProjectToSessionWizard wizard =
+              new AddProjectToSessionWizard(project, getWindow(project), negotiation);
 
-              wizard.setModal(false);
-              wizard.open();
-            },
-            ModalityState.defaultModalityState());
+          wizard.setModal(false);
+          wizard.open();
+        },
+        ModalityState.defaultModalityState());
   }
 
   private Window getWindow(Project project) {
@@ -215,11 +213,7 @@ public class NegotiationHandler implements INegotiationHandler {
                 MessageFormat.format(
                     Messages.NegotiationHandler_project_sharing_canceled_message, peerName);
 
-            ApplicationManager.getApplication()
-                .invokeLater(
-                    () ->
-                        NotificationPanel.showInformation(
-                            message, "Project sharing canceled remotely"));
+            NotificationPanel.showInformation(message, "Project sharing canceled remotely");
 
             return new Status(IStatus.CANCEL, SarosComponent.PLUGIN_ID, message);
 
--- a/intellij/src/saros/intellij/editor/ProjectAPI.java
+++ b/intellij/src/saros/intellij/editor/ProjectAPI.java
@@ -8,6 +8,7 @@ import com.intellij.openapi.project.Project;
 import com.intellij.openapi.roots.ProjectFileIndex;
 import com.intellij.openapi.vfs.VirtualFile;
 import org.jetbrains.annotations.NotNull;
+import saros.intellij.runtime.EDTExecutor;
 import saros.intellij.runtime.FilesystemRunner;
 
 /** Static utility class for interacting with the project-level Intellij editor API. */
@@ -55,10 +56,13 @@ public class ProjectAPI {
    */
   public static Editor openEditor(
       @NotNull Project project, @NotNull VirtualFile virtualFile, final boolean activate) {
-    return FilesystemRunner.runReadAction(
+
+    FileEditorManager fileEditorManager = getFileEditorManager(project);
+
+    return EDTExecutor.invokeAndWait(
         () ->
-            getFileEditorManager(project)
-                .openTextEditor(new OpenFileDescriptor(project, virtualFile), activate));
+            fileEditorManager.openTextEditor(
+                new OpenFileDescriptor(project, virtualFile), activate));
   }
 
   /**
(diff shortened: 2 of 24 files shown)


saros-infrastructure:


```diff

Issues
======
- Added 2
           

Complexity increasing per file
==============================
- intellij/src/saros/intellij/ui/views/buttons/ConsistencyButton.java  1
- intellij/src/saros/intellij/runtime/EDTExecutor.java  7
- intellij/src/saros/intellij/runtime/FilesystemRunner.java  1
- intellij/src/saros/intellij/runtime/IntellijUISynchronizer.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5068197&bid=16675566)

intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java
```diff
@@ -78,30 +78,28 @@ public void handleIncomingProjectNegotiation(AbstractIncomingProjectNegotiation
...
-            },
-            ModalityState.defaultModalityState());
+    EDTExecutor.invokeLater(
+        () -> {
+          JoinSessionWizard wizard =
+              new JoinSessionWizard(project, getWindow(project), negotiation);
+          wizard.setModal(true);
+          wizard.open();
+        },
+        ModalityState.defaultModalityState());
```
m273d15: Why do you use the default modality state (for multiple calls of EDTExecutor methods in this PR) instead of using your method with the signature `invokeLater(@NotNull Runnable)` that uses the default state?

intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java
```diff
@@ -215,11 +213,7 @@ protected IStatus run(IProgressMonitor monitor) {
                 MessageFormat.format(
                     Messages.NegotiationHandler_project_sharing_canceled_message, peerName);
 
-            ApplicationManager.getApplication()
-                .invokeLater(
-                    () ->
-                        NotificationPanel.showInformation(
-                            message, "Project sharing canceled remotely"));
+            NotificationPanel.showInformation(message, "Project sharing canceled remotely");
```
m273d15: (Not important for this PR:) The constant string can be moved into the messages.properties.

intellij/src/saros/intellij/runtime/FilesystemRunner.java
```diff
@@ -75,16 +53,13 @@ private FilesystemRunner() {
    * @see Application#runWriteAction(Runnable)
    */
   public static void runWriteAction(
-      @NotNull final Runnable runnable, @Nullable final ModalityState modalityState) {
-
-    final ModalityState chosenModalityState =
```
m273d15: Why is this check not necessary anymore?

intellij/src/saros/intellij/runtime/IntellijUISynchronizer.java
```diff
@@ -1,46 +1,42 @@
...
   private static final Logger log = Logger.getLogger(IntellijUISynchronizer.class);
 
   @Override
   public void asyncExec(Runnable runnable) {
-    exec(runnable, true);
+    try {
+      EDTExecutor.invokeLater(runnable);
+
+    } catch (ProcessCanceledException e) {
+      log.error("Synchronous execution on EDT interrupted - " + runnable, e);
```
m273d15:
```suggestion
      log.error("Asynchronous execution on EDT interrupted - " + runnable, e);
```

m273d15: Is this a possible scenario? I would expect that invokeLater directly returns and does not throw a ProcessCanceledException.

intellij/src/saros/intellij/runtime/EDTExecutor.java
```diff
@@ -0,0 +1,130 @@
...
+
+    application.invokeAndWait(runnable, modalityState);
+  }
+
+  /**
+   * Calls {@link #invokeAndWait(Runnable, ModalityState)} with {@link
+   * ModalityState#defaultModalityState()}.
+   */
+  public static void invokeAndWait(@NotNull Runnable runnable) {
+    invokeAndWait(runnable, ModalityState.defaultModalityState());
```
m273d15: There is already an [invokeAndWait method that uses the defaultModalityState](https://github.com/JetBrains/intellij-community/blob/5ce01ec91f0eae2b44a0f6e712eb219863967ce9/platform/core-api/src/com/intellij/openapi/application/Application.java#L386)
Same for `invokeLater`

intellij/src/saros/intellij/editor/ProjectAPI.java
```diff
@@ -56,7 +57,9 @@ public static boolean isOpen(@NotNull Project project, @NotNull Document documen
    * @return whether there is an open text editor for the given file
    */
   public static boolean isOpenInTextEditor(@NotNull Project project, @NotNull VirtualFile file) {
-    FileEditor[] fileEditors = getFileEditorManager(project).getEditors(file);
+    FileEditorManager fileEditorManager = getFileEditorManager(project);
+
+    FileEditor[] fileEditors = EDTExecutor.invokeAndWait(() -> fileEditorManager.getEditors(file));
```
m273d15: In contrast to the other methods you wrapped in an `invokeAndWait` the method `getEditors` has [no comment](https://github.com/JetBrains/intellij-community/blob/5ce01ec91f0eae2b44a0f6e712eb219863967ce9/platform/platform-api/src/com/intellij/openapi/fileEditor/FileEditorManager.java#L122) "Must be called from <a href="https://docs.oracle.com/javase/tutorial/uiswing/concurrency/dispatch.html">EDT</a>." Is it necessary to wrap this method?

intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java
tobous: The methods implicitly using the default modality state were meant more as convenience methods. As the default modality state is most often used, I thought it would be nice to have a method were we wouldn't have to specify it.

But I didn't feel like it was necessary to replace all prior usages explicitly stating to use the default modality state with this new call.

intellij/src/saros/core/ui/eventhandler/NegotiationHandler.java
tobous: There is a lot of stuff in the UI that needs cleaning up.

intellij/src/saros/intellij/editor/ProjectAPI.java
tobous: Yes, it is not stated in the javadoc. But in the actual [implementation](https://github.com/JetBrains/intellij-community/blob/5ce01ec91f0eae2b44a0f6e712eb219863967ce9/platform/platform-impl/src/com/intellij/openapi/fileEditor/impl/FileEditorManagerImpl.java#L1312) of the method, it still very much checks for it being run on the EDT.

intellij/src/saros/intellij/runtime/EDTExecutor.java
tobous: And, as far as I can tell, [they do exactly the same](https://github.com/JetBrains/intellij-community/blob/5ce01ec91f0eae2b44a0f6e712eb219863967ce9/platform/platform-impl/src/com/intellij/openapi/application/impl/ApplicationImpl.java#L485). But, with these being convenience methods, I thought it would be better to specify this assumption in our code and not rely on theirs.

intellij/src/saros/intellij/runtime/IntellijUISynchronizer.java
tobous: This was just copy-pasted (badly) from the old logic. And no, it does not seem like the async execution can throw such an exception. Will remove the check.

intellij/src/saros/intellij/runtime/FilesystemRunner.java
tobous: Because the new logic does not accept null values. The field is now marked as `NotNull`.

The old implementation was just a different approach to offer a convenience call: Specify the modality state or, if you don't (i.e. pass `null`), we will use the default modality state. I thought it would be clearer if there rather were two different methods, one requiring a modality state (that is non-null) and one that does not (and always uses the default modality state).

I also checked all calls of these methods to ensure that none of them are passing null when calling one of these methods.

Also, as no call ever passed null, I did not introduce any such convenience methods here. But that goes back to the other point of whether it is necessary to replace all old calls that use the default modality state with the convenience method.

saros-infrastructure:


```diff

Issues
======
- Added 2
           

Complexity increasing per file
==============================
- intellij/src/saros/intellij/runtime/IntellijUISynchronizer.java  2
- intellij/src/saros/intellij/ui/views/buttons/ConsistencyButton.java  1
- intellij/src/saros/intellij/runtime/EDTExecutor.java  7
- intellij/src/saros/intellij/runtime/FilesystemRunner.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5068197&bid=16675566)

tobous:
Rebased onto the current master. Changed order as the base PR has not yet been reviewed.

PR #766, opened 2019-11-20
Remove old Discovery Manager From Eclipse

stefaus:
This PR extends the `XMPPContactsService` by the retrieval of contact client features via the new `DiscoveryService` instead of the old `DiscoveryManager` and makes this information available as an attribute of an XMPPContact object.
This allows to remove the usage of the old `DiscoveryManager` outside of `saros.core`.

PR #630, opened 2019-08-07
New Contact Service and First Eclipse usage

stefaus:
First of all, Sorry for the patch size!

### Patch 1 is a completly new package providing a unified contact service

This contact service is called XMPPContactsService and listens on all connection and roster changes. It provides methods for accessing the current contact lists consisting of XMPPContact objects.

XMPPContact objects identify a contact by its base JID, remain the same object reference during a connection, and are updated by the XMPPContactsService when events occur. This way a object reference can be passed in the system and provide updated informations about availability, the Saros Resource or Nickname without the need of an additional service injection / query. This obsoletes additional UI logic to call the DiscoveryManager on changes, etc.

An XMPPContact returns a ContactStatus that bundles a new presence type that is a simplified view of all possible states that XMPP allows.

Instead of using the existing DiscoveryManager or RosterTracker, I decided to reimplement these features. This allows better predictability of concurrent events, and was easier than extending and refactoring them. This way their usage can probably faded out.

### Patch 2 replaces logic in the Eclipse saros.ui.model.roster package and some referencing parts

The RosterContentProvider was listing to roster updates, triggering feature queries and filtered the results. This is now replaced by the XMPPContactsService. Additionally some Smack references are already removed.

The goal of these patches is to create a good starting point for removing XMPP logic in the UI packages. It obsoletes implementations like in eclipse, html with saros.ui.core_facades and the roster handling in IntelliJ.

PR #847, opened 2020-02-26
Exchange Feature Info via Version Manager

stefaus:
The context of this PR is to add a generic way to share information like a Skype name outside of a session. Additionally add a platform info like eclipse or intellij to allow client checks.

### Current behavior
After xmpp connect a simple available presence update is send to the server.
When a available presence update of other contacts is retrieved, a service discovery via `DiscoveryService` is done. This returns a list of strings which is checked for 'saros' and 'saros.server'.
A version check is done via `VersionManager` at session start.

### Proposal
Remove the discovery service logic and use the `VersionManager` like [@srossbach already proposed](https://github.com/saros-project/saros/pull/766#issuecomment-558698991) to handle features. This allows to exchange key-value lists, include information like a Skype username and remove extra Skype info messages.

The `VersionManager` uses a saros specific message and a different client would just not response and this would work. But this behavior like the current discovery logic leads to a message flood to all available contacts and should be avoided.
[XEP-115](https://xmpp.org/extensions/xep-0115.html) proposes a way to exchange version information already via presence updates. This allows to get info about the client (saros), platform (like eclipse) and version string (like 15.0.0.dev) at time of contact availability. This limits the need to send version exchange requests to saros clients.

A thing to discuss is if the current version compatibility logic is needed.
The logic to check backwards compatibility in server→client and client→server direction may be working but is currently never used.
As @tobous mentioned a different qualifier should already make clients incompatible e.g. to avoid problems using different builds.
So a proposal would be to check if the complete version string matches and later if needed add a logic that allows compatibility within the same patch level or some other simple logic.

@srossbach @tobous @m273d15 any input on this?


saros-infrastructure:


```diff

Complexity increasing per file
==============================
- core/src/saros/versioning/ClientInfo.java  3
         

Complexity decreasing per file
==============================
+ eclipse/src/saros/communication/SkypeManager.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

m273d15:
> Remove the discovery service logic and use the VersionManager like @srossbach already proposed to handle features. This allows to exchange key-value lists, include information like a Skype username and remove extra Skype info messages.

I like the idea, but we should think about renaming and restructuring the code. A VersionExchangeExtension/VersionManager that is used to share the skype name (or other information that is shared in the future) might confuse developers. I think the perspective (and code structure) should be "It is a general feature for sharing client-specific information that is also used to share the version" instead of "It is a feature for sharing the version information that is also used to share general information".

> So a proposal would be to check if the complete version string matches and later if needed add a logic that allows compatibility within the same patch level or some other simple logic.

I think this solution would be sufficient for our current needs :+1:

stefaus:
> we should think about renaming and restructuring the code
Yes definitely! First thought (see todo remark at `VersionManager`) was to name it feature manager, but maybe better something in ClientInfoManager direction.

srossbach:
I do not like the approach of encoding / decoding the client capabilities (platform, version) into the node element of the presence extension. This is not the purpose of XEP-115.

The purpose of this extension is to generate a node and ver pair so you can lookup the supported features in your cache (did you even configure it ? see https://github.com/igniterealtime/Smack/blob/master/smack-extensions/src/main/java/org/jivesoftware/smackx/caps/EntityCapsManager.java#L265 but I do not know if Smack sets a default one) and so reduce packet sending and replies.

stefaus:
Yeah, I should have written that XEP-115 allows first of all to get a client name (in node element, which should be a URI) and a hash over service discovery values.
So according to the XEP draft this should just be something like "https://saros-project.org".
The hacky thing I done here is to add platform and version separated by whitespace.
If changed to a "/" we could use "https://saros-project.org/eclipse/15.0.0.dev" as URI and this would be more or less fine.

Using Caps Management just makes sense if service discovery is used. As as I understand your view this is the thing you wanted to avoid.
So my thought for including the version was to avoid the disco request. But as the version message is easy and should be stable just knowing that a client is a saros client is good enough.

---
So how about just "https://saros-project.org" as node, remove service discovery stuff and put platform and version into the version manager message?

srossbach:
I guess that it is the best approach. Using a simple node name to indicate that this is a Saros client (either a server or a normal one). The remaining data should be exchanged through IQ packets (or whatever the underlying network layer is), i.e there is central manager that broadcasts or requests several configurations (e.g version, skype name, whatever). We basically just have to decide if we push our data to other clients or if other clients will always pull us.

E.g you call VersionManager.requestVersion(contact) which will call some implementation of XYZFeatureBlaBla.getVersion(contact) which can then decide based on the node name if it even starts sending a packet to the destination or not.

stefaus:
I updated the code.
- as talked about in the Stand-Ups the Version Compatibility stuff is simplified to just check if the Version matches
  - to include a check for qualifier and platform is part of a later PR
- changed the Request/Get version handling to push the information to all contacts that are/become available and have saros support; This way a info like skypename is already available when the contact context menu is opened
- moved the skypename and server feature exchange to the VersionManager
- changed the check for saros availability from DiscoveryService to exchange via Caps Presence Extention
- removed stuff like DiscoveryService and ResourceFeature
- as compatibility with older versions is bricked here anyway, added NamedMapConverter to xstream for a shorter and better to read Map conversion
- Not included now for readability: would rename `VersionManager` to `InfoManager` and the package `saros.versioning` to `saros.communication.info`

core/src/saros/communication/extensions/VersionExchangeExtension.java
```diff
@@ -1,54 +0,0 @@
-package saros.communication.extensions;
```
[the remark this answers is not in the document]
stefaus: yes this whole pr breaks compatibility, but @tobous, @m273d15 and i are fine with that

older clients will newer think about sending this packet because, as the `saros` feature is removed from service discovery, they will not see saros support.

core/src/saros/misc/xstream/XStreamExtensionProvider.java
```diff
@@ -99,6 +100,27 @@ public XStreamExtensionProvider(String namespace, String elementName, Class<?>..
 
     xstream.registerConverter(BooleanConverter.BINARY);
     xstream.registerConverter(new UrlEncodingStringConverter());
+
+    /*
+     * set map conversion format to
```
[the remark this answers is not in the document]
stefaus: not necessary, but because this pr breaks compatibility anyway a chance to increase readability of our xmpp messages and make them shorter

so already changing this breaks the whole compatibility because old `VersionExchangeExtension` couldn't be parsed

core/src/saros/net/xmpp/XMPPConnectionService.java
```diff
@@ -34,6 +38,8 @@
  */
 @Component(module = "net")
 public class XMPPConnectionService {
+  public static final String XMPP_CLIENT_IDENTIFIER = "https://saros-project.org";
```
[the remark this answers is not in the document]
stefaus: I don't think so, if another protocol would use a different naming we wouldn't use it.
The resource part is also xmpp specific and should move to `saros.net.xmpp`

core/src/saros/versioning/VersionManager.java
```diff
@@ -1,330 +1,184 @@
...
-
-    final Version currentVersion = localVersion;
+  public void setLocalInfo(String key, String value) {
+    Objects.requireNonNull(key, "key is null");
+
+    if (value == null) localInfo.remove(key);
+    else localInfo.put(key, value);
+
+    if (remoteInfo.isEmpty()) return;
+    dispatchThread.executeAsDispatch(
```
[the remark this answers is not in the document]
stefaus: I just looked till
https://github.com/saros-project/saros/blob/f13bdc3ea5c5d295428c01db145bd8d5a41e189d/core/src/saros/net/internal/XMPPTransmitter.java#L131

and thought that synchronized doesn't look good.

I was thinking that I don't want to hang the ui on changing for example the skype name in prefs and remembered you mentioning this https://github.com/saros-project/saros/pull/630#discussion_r312391180

Could you explain where do you see something leaking, I don't see what you mean by internal stuff.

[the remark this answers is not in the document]
srossbach: ```sendPacket``` do not block even it is synchronized. The only reason the DispatchThread is visible outside the network package is so skip serialization if you send messages to yourself (the only access in the ActitvitySequencer). The only thing that had to be synchronized is the setting and sending. Otherwise you could trigger a send on contact update, generate a packet which is not send yet, then you set something and this packet is transmitted and afterwards the old packet is transmitted (guess that will never happen but theoretically it could).

srossbach:
> 
> 
> > Changing the version comparison logic should not be part of this PR.
> 
> I could but it shouldn't be merged without theses other changes, because this whole pr breaks compatibility multiple times (remove of COMPATIBILITY_KEY, rename message, remove of servicediscovery, xstream maps).

I do not think it matters what patch will break the versioning. Looking at the first 2 commits (that should be part of its own PR) of this PR I think it is sufficient to just set the COMPATIBILITY_KEY value in the response always to: TOO_OLD.

The XML is currently pretty bad formatted and some packet extensions could need an overhaul anyway. So I think it is pretty good to break any compatibility for now. Or did you plan to make a release in the near future (@tobous any plans regarding IntelliJ ?)

core/src/saros/versioning/VersionManager.java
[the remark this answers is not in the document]
stefaus: Ok, I removed the DispatchThread.

saros-infrastructure:


```diff

Issues
======
+ Solved 4
- Added 2
           

Complexity increasing per file
==============================
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
- core/src/saros/versioning/ClientInfo.java  4
         

Complexity decreasing per file
==============================
+ eclipse/src/saros/ui/eventhandler/ServerPreferenceHandler.java  -1
+ eclipse/src/saros/communication/SkypeManager.java  -2
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

tobous:
@stefaus For future reference: When you create a stacked pull request (i.e. a PR not based on the master), you have to tell codacy to analyze the base branch (see [settings](https://app.codacy.com/manual/Saros/saros/settings/branches)) beforehand as it otherwise can't run for the PR.

If you forgot, you can simply restart the codacy run after enabling the base branch analysis.

I have already done this for this PR.

core/src/saros/versioning/VersionManager.java
```diff
@@ -27,188 +32,144 @@
...
-                    if (versionExchange == null) return false;
+    return Optional.ofNullable(clientInfo.getInfo(key));
+  }
 
-                    return rqJID.toString().equals(packet.getFrom())
-                        && ((IQ) packet).getType() == IQ.Type.RESULT
-                        && exchangeID.equals(versionExchange.get(ID_KEY));
-                  }
-                }));
+  private void sendVersionInfo(XMPPContact contact) {
```
srossbach: Small info that non Saros contacts are ignored would be nice.

core/src/saros/versioning/VersionManager.java
```diff
@@ -27,188 +32,144 @@
...
+          contact.ifPresent(remoteInfo::remove);
+        } else if (updateType == UpdateType.NOT_CONNECTED) {
+          remoteInfo.clear();
         }
       };
 
   public VersionManager(
-      @SarosVersion String version, final IReceiver receiver, final ITransmitter transmitter) {
-
+      @SarosVersion String version,
```
srossbach: As @m273d15 already said, you are mixing concerns. This no longer just a VersionManager. In fact the whole versionlogic is now more or less a one-liner in that class.

eclipse/src/saros/communication/SkypeManager.java
```diff
@@ -4,199 +4,48 @@
...
-      };
-
-  public SkypeManager(XMPPConnectionService connectionService, IPreferenceStore preferenceStore) {
-    this.connectionService = connectionService;
+  public SkypeManager(VersionManager versionManager, IPreferenceStore preferenceStore) {
+    this.versionManager = versionManager;
     this.preferenceStore = preferenceStore;
-    this.connectionService.addListener(this);
+
+    versionManager.setLocalInfo(PreferenceConstants.SKYPE_USERNAME, getLocalSkypeName());
```
srossbach: Seems we have to implement startable to the container. Actually this is "heavy" work, at least it seems so, because nothing it actually send at that point anyway. Maybe add an TODO ?

saros-infrastructure:


```diff

Issues
======
+ Solved 1
- Added 3
           

Complexity increasing per file
==============================
- core/src/saros/communication/info/ClientInfo.java  4
- core/src/saros/communication/info/VersionCompatibilityResult.java  1
- core/test/junit/saros/communication/info/InfoManagerTest.java  1
- core/src/saros/communication/info/Compatibility.java  4
- eclipse/src/saros/ui/eventhandler/ServerPreferenceHandler.java  2
- core/src/saros/communication/info/Version.java  11
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
- core/src/saros/communication/info/InfoManager.java  5
         

Complexity decreasing per file
==============================
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
+ eclipse/src/saros/communication/SkypeManager.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

core/src/saros/net/xmpp/contact/DiscoveryService.java
```diff
@@ -1,114 +0,0 @@
...
-import saros.net.xmpp.JID;
-import saros.util.NamedThreadFactory;
-
-/**
- * A Simple Service to get and cache {@link ResourceFeature} Support of XMPP resources.
- *
- * <p>All methods are non-blocking and run in a separate thread, which should be stopped after usage
- * via {@link #stop()}.
- */
-class DiscoveryService {
```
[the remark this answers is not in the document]
stefaus: The old DiscoveryManager is already removed since #806

core/src/saros/versioning/VersionManager.java
stefaus: done

saros-infrastructure:


```diff

Issues
======
+ Solved 1
- Added 3
           

Complexity increasing per file
==============================
- core/src/saros/communication/info/Version.java  11
- core/src/saros/communication/info/VersionCompatibilityResult.java  1
- core/src/saros/communication/info/Compatibility.java  4
- core/test/junit/saros/communication/info/InfoManagerTest.java  1
- eclipse/src/saros/ui/eventhandler/ServerPreferenceHandler.java  2
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
- core/src/saros/communication/info/InfoManager.java  5
- core/src/saros/communication/info/ClientInfo.java  4
         

Complexity decreasing per file
==============================
+ eclipse/src/saros/communication/SkypeManager.java  -2
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

core/src/saros/versioning/VersionManager.java
stefaus: As I have written: >Not included now for readability: would rename VersionManager to InfoManager and the package saros.versioning to saros.communication.info

I added a Commit renaming the Class and package.

eclipse/src/saros/communication/SkypeManager.java
stefaus: Do you mean for the `preferenceStore.addPropertyChangeListener`?
We could add it but does it have a impact?

saros-infrastructure:


```diff

Issues
======
+ Solved 1
- Added 3
           

Complexity increasing per file
==============================
- eclipse/src/saros/ui/eventhandler/ServerPreferenceHandler.java  2
- core/src/saros/communication/info/Version.java  11
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
- core/src/saros/communication/info/InfoManager.java  5
- core/src/saros/communication/info/Compatibility.java  4
- core/src/saros/communication/info/VersionCompatibilityResult.java  1
- core/test/junit/saros/communication/info/InfoManagerTest.java  1
- core/src/saros/communication/info/ClientInfo.java  4
         

Complexity decreasing per file
==============================
+ eclipse/src/saros/communication/SkypeManager.java  -2
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

core/src/saros/communication/info/InfoManager.java
```diff
@@ -0,0 +1,182 @@
...
+import saros.net.xmpp.contact.XMPPContact;
+import saros.net.xmpp.contact.XMPPContactsService;
+
+/**
+ * Component for figuring out whether two Saros plug-in instances with known Version are compatible
+ * and to exchange Infos outside of Sessions.
+ *
+ * <p>This class compares if local and remote version (not checking qualifier) are the same.
+ *
+ * <p>If you want to implement backward compatibility in a later version, as a suggestion:
```
srossbach: This is not very helpful as Javadoc. Guess it is better to just use a simple comment here.

core/src/saros/communication/info/InfoManager.java
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * Determines the version compatibility with the given peer.
+   *
+   * @param rqJID the resource qualified JID of the peer
+   * @return the {@link VersionCompatibilityResult VersionCompatibilityResult}
+   */
+  public VersionCompatibilityResult determineVersionCompatibility(JID rqJID) {
+    Version remoteVersion =
+        contactsService
+            .getContact(rqJID.getRAW())
```
srossbach: It doesn't seem to matter if you use a FQ JID or not. I guess you could also use the FQJID of an instant messager and would get the Saros version as long as the user is also connected with a Saros client.

core/src/saros/communication/info/InfoManager.java
stefaus: done

core/src/saros/communication/info/InfoManager.java
stefaus: yes, changed to jid

saros-infrastructure:


```diff

Issues
======
+ Solved 1
- Added 3
           

Complexity increasing per file
==============================
- core/src/saros/communication/info/ClientInfo.java  4
- core/src/saros/communication/info/VersionCompatibilityResult.java  1
- core/src/saros/communication/info/Compatibility.java  4
- eclipse/src/saros/ui/eventhandler/ServerPreferenceHandler.java  2
- core/src/saros/communication/info/InfoManager.java  5
- core/src/saros/communication/info/Version.java  11
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
- core/test/junit/saros/communication/info/InfoManagerTest.java  1
         

Complexity decreasing per file
==============================
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
+ eclipse/src/saros/communication/SkypeManager.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

saros-infrastructure:


```diff

Issues
======
+ Solved 1
- Added 3
           

Complexity increasing per file
==============================
- core/src/saros/communication/info/Version.java  11
- eclipse/src/saros/ui/eventhandler/ServerPreferenceHandler.java  2
- core/src/saros/communication/info/ClientInfo.java  4
- core/src/saros/communication/info/Compatibility.java  4
- core/test/junit/saros/communication/info/InfoManagerTest.java  1
- core/src/saros/communication/info/VersionCompatibilityResult.java  1
- core/src/saros/communication/info/InfoManager.java  5
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
         

Complexity decreasing per file
==============================
+ eclipse/src/saros/communication/SkypeManager.java  -2
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

server/src/saros/server/net/ServerFeatureAdvertiser.java
```diff
@@ -1,46 +1,19 @@
...
-  }
-
-  private static void advertiseServerFeature(Connection connection) {
-    LOG.info("Starting to advertise ourselves as server");
-
-    ServiceDiscoveryManager discoveryManager = ServiceDiscoveryManager.getInstanceFor(connection);
-
-    discoveryManager.addFeature(ResourceFeature.SAROS_SERVER.getIdentifier());
+  public ServerFeatureAdvertiser(VersionManager versionManager) {
+    versionManager.setLocalInfo(PreferenceConstants.SERVER_SUPPORT, Boolean.TRUE.toString());
```
tobous: Don't you already do this in the `ServerPreferenceHandler`? Why is this logic present twice?

Or is the `ServerPreferenceHandler` the unnecessary one? According to my IDE, it is never referenced/used.

core/src/saros/net/xmpp/XMPPConnectionService.java
```diff
@@ -272,6 +284,15 @@ public synchronized void connect(
     }
   }
 
+  private void sendAvailablePresenceWithClientIdentifier() {
+    String ver = EntityCapsManager.getInstanceFor(connection).getCapsVersion();
+    CapsExtension caps = new CapsExtension(XMPP_CLIENT_IDENTIFIER, ver, "sha-1");
```
tobous: Move hashing algorithm choice to static constant to make it easier to notice when looking at the class?

core/src/saros/net/xmpp/contact/XMPPContact.java
```diff
@@ -226,19 +205,15 @@ boolean setNickname(String newNickname) {
...
 
-    Optional<Resource> oldResource =
-        resources.stream().filter(r -> compareRawJid(r.fullJid, fullJid)).findFirst();
-    oldResource.ifPresent(resources::remove);
-    resources.add(
-        new Resource(
-            fullJid,
-            status,
-            oldResource.map(r -> r.features).orElse(EnumSet.noneOf(ResourceFeature.class))));
+    resources.stream().filter(r -> compareRawJid(r.fullJid, fullJid)).map(resources::remove);
```
tobous: Does removing an element from a collection inside a stream operation work or can this also lead to `ConcurrentModificationExceptions`?

Use `removeIf()` instead?

Also, the new logic removes all matching entries while the old logic only removed the first found entry. Does this make a difference or is there always only one match?

core/test/junit/saros/communication/info/InfoManagerTest.java
```diff
@@ -19,28 +19,29 @@
...
 
   private ITransmitter aliceTransmitter;
-  private ITransmitter bobTransmitter;
-
   private IReceiver aliceReceiver;
-  private IReceiver bobReceiver;
-
-  private VersionManager versionManagerRemote;
-  private VersionManager versionManagerLocal;
+  private XMPPContactsService aliceContactsService;
```
tobous: Make local variable in `init` method as it is not used otherwise?

saros-infrastructure:


```diff

Issues
======
+ Solved 1
- Added 2
           

Complexity increasing per file
==============================
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
- eclipse/src/saros/ui/eventhandler/ServerPreferenceHandler.java  2
- core/test/junit/saros/communication/info/InfoManagerTest.java  1
- core/src/saros/communication/info/Compatibility.java  4
- core/src/saros/communication/info/InfoManager.java  5
- core/src/saros/communication/info/ClientInfo.java  4
- core/src/saros/communication/info/Version.java  11
- core/src/saros/communication/info/VersionCompatibilityResult.java  1
         

Complexity decreasing per file
==============================
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
+ eclipse/src/saros/communication/SkypeManager.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

saros-infrastructure:


```diff

Issues
======
+ Solved 1
- Added 1
           

Complexity increasing per file
==============================
- core/src/saros/communication/info/InfoManager.java  5
- core/src/saros/communication/info/VersionCompatibilityResult.java  1
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
- core/src/saros/communication/info/ClientInfo.java  4
- core/src/saros/communication/info/Compatibility.java  4
- core/test/junit/saros/communication/info/InfoManagerTest.java  1
- core/src/saros/communication/info/Version.java  11
         

Complexity decreasing per file
==============================
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
+ eclipse/src/saros/communication/SkypeManager.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

core/src/saros/net/xmpp/contact/XMPPContact.java
stefaus: There is always only one match, it worked but was a quick change. Changed to removeIf

core/src/saros/net/xmpp/XMPPConnectionService.java
stefaus: done

core/src/saros/communication/extensions/VersionExchangeExtension.java
```diff
@@ -19,30 +20,19 @@
...
    *
-   * @param key the key whose associated value is to be returned
-   * @return the value to which the specified key is mapped, or null if this map contains no mapping
-   *     for the key
+   * @return data map
    */
-  public String get(String key) {
-    return data.get(key);
+  public Map<String, String> getData() {
+    return data;
```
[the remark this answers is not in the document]
stefaus: changed

core/src/saros/versioning/VersionManager.java
```diff
@@ -1,214 +1,145 @@
...
+  private final ConcurrentHashMap<String, String> localInfo = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<XMPPContact, ClientInfo> remoteInfo = new ConcurrentHashMap<>();
+
+  private final PacketListener versionInfoListener =
+      packet -> {
+        log.debug("received version info from " + packet.getFrom());
+        handleInfo(packet);
+      };
+
+  private final IContactsUpdate contactsUpdateListener =
```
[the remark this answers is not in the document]
stefaus: yes, it will receive an update for the old resource going offline, and the new becoming available.

core/src/saros/versioning/VersionManager.java
```diff
@@ -102,6 +105,38 @@ public VersionCompatibilityResult determineVersionCompatibility(JID jid) {
...
+   */
+  public void setLocalInfo(String key, String value) {
+    Objects.requireNonNull(key, "key is null");
+
+    if (value == null) localInfo.remove(key);
+    else localInfo.put(key, value);
+
+    if (remoteInfo.isEmpty()) return;
+    // send new info to all known online contacts
+    Enumeration<XMPPContact> contacts = remoteInfo.keys();
```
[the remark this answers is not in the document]
stefaus: no, changed it

eclipse/src/saros/communication/SkypeManager.java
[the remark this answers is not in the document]
stefaus: I added a check, but don't think its important here.

Same doesn't work for the listener because we want to update if the skypename was removed.

core/src/saros/communication/info/InfoManager.java
```diff
@@ -0,0 +1,179 @@
...
+import saros.net.xmpp.contact.XMPPContactsService;
+
+/**
+ * Component for figuring out whether two Saros plug-in instances with known Version are compatible
+ * and to exchange Infos outside of Sessions.
+ *
+ * <p>This class compares if local and remote version (not checking qualifier) are the same.
+ */
+@Component(module = "core")
+public class InfoManager {
```
m273d15: What do you think about splitting the class into two classes:
* InfoManager providing the general methods for exchanging information
* VersionManager using the InfoManager to provide methods for exchanging the versions and determining compatibility

srossbach: I would suggest the same as.

stefaus: Yes, that's a good idea, and I've changed it accordingly.

saros-infrastructure:


```diff

Issues
======
+ Solved 3
- Added 8
           

Complexity increasing per file
==============================
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
- core/src/saros/communication/InfoManager.java  5
         

Complexity decreasing per file
==============================
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
+ core/src/saros/versioning/VersionManager.java  -2
+ eclipse/src/saros/communication/SkypeManager.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

saros-infrastructure:


```diff

Issues
======
+ Solved 3
- Added 1
           

Complexity increasing per file
==============================
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
- core/src/saros/communication/InfoManager.java  5
         

Complexity decreasing per file
==============================
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
+ eclipse/src/saros/communication/SkypeManager.java  -2
+ core/src/saros/versioning/VersionManager.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

core/src/saros/communication/InfoManager.java
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * This Component exchanges information in a key-value form with other contacts outside of Sessions.
+ *
+ * <p>Every info set via {@link #setLocalInfo(String, String)} is send / updated to available
+ * contacts with Saros support. Retrieval of available data with {@link #getRemoteInfo(XMPPContact,
+ * String)}.
+ */
+public class InfoManager {
+  private class ClientInfo {
+    private final Map<String, String> infos;
```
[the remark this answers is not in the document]
stefaus: I removed in the mean time because it has not a big value and I use ClientInfo mainly to avoid `Map<XMPPContact, Map<String, String>>` but yes probably better to keep it.

core/src/saros/communication/InfoManager.java
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * This Component exchanges information in a key-value form with other contacts outside of Sessions.
+ *
+ * <p>Every info set via {@link #setLocalInfo(String, String)} is send / updated to available
+ * contacts with Saros support. Retrieval of available data with {@link #getRemoteInfo(XMPPContact,
+ * String)}.
+ */
+public class InfoManager {
+  private class ClientInfo {
```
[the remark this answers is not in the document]
stefaus: done

core/src/saros/communication/InfoManager.java
```diff
@@ -0,0 +1,140 @@
...
+
+    // send new info to all known online contacts
+    for (XMPPContact contact : remoteInfo.keySet()) sendInfo(contact);
+  }
+
+  private void handleInfo(Packet packet) {
+    Optional<XMPPContact> contact = contactsService.getContact(packet.getFrom());
+    if (!contact.isPresent()) return;
+
+    InfoExchangeExtension versionInfo = InfoExchangeExtension.PROVIDER.getPayload(packet);
```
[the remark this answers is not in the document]
stefaus: done

core/src/saros/communication/InfoManager.java
```diff
@@ -0,0 +1,144 @@
...
+    String sarosJid = contact.getSarosJid().map(JID::toString).orElse(null);
+    if (sarosJid == null) return;
+
+    IQ packet = InfoExchangeExtension.PROVIDER.createIQ(new InfoExchangeExtension(localInfo));
+    packet.setType(IQ.Type.SET);
+    packet.setTo(sarosJid);
+    try {
+      transmitter.sendPacket(packet);
+    } catch (IOException e) {
+      log.error("could not send version response to " + sarosJid, e);
```
srossbach: error message is wrong

stefaus: done

saros-infrastructure:


```diff

Issues
======
+ Solved 3
- Added 1
           

Complexity increasing per file
==============================
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
- core/src/saros/communication/InfoManager.java  5
         

Complexity decreasing per file
==============================
+ core/src/saros/versioning/VersionManager.java  -2
+ eclipse/src/saros/communication/SkypeManager.java  -2
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

saros-infrastructure:


```diff

Issues
======
+ Solved 3
- Added 1
           

Complexity increasing per file
==============================
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
- core/src/saros/communication/InfoManager.java  5
         

Complexity decreasing per file
==============================
+ core/src/saros/versioning/VersionManager.java  -2
+ eclipse/src/saros/communication/SkypeManager.java  -2
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

saros-infrastructure:


```diff

Issues
======
+ Solved 3
- Added 1
           

Complexity increasing per file
==============================
- core/src/saros/communication/InfoManager.java  5
- core/src/saros/communication/extensions/InfoExchangeExtension.java  2
         

Complexity decreasing per file
==============================
+ eclipse/src/saros/communication/SkypeManager.java  -2
+ core/src/saros/versioning/VersionManager.java  -2
+ server/src/saros/server/net/ServerFeatureAdvertiser.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5072803&bid=16689379)

PR #851, opened 2020-02-26
Enable gradle build scan for github workflow

tobous:
#### [BUILD][I] Only adjust plugin.xml on master builds

Adjusts the logic modifying the plugin.xml to only run when building on
the master branch.

The plugin.xml is adjusted to enable compatibility with all JetBrains
IDEs for dev builds. Therefore, this adjustment is only needed when
actually publishing the artifacts, i.e. when building the master branch.

Otherwise, this is only included as a superfluous build step cluttering
up the log.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,7 @@ jobs:
       run: |
         ./gradlew \
           -PskipTestSuites=true \
+          -PuseBuildScan=true \
           -PaggregateTestResults=true \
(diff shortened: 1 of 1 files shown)


tobous:
@m273d15 I am not sure on the second commit. The build step is displayed either way.

On the one hand, the adjustment is unnecessary on any branch besides the master.

On the other hand, if there ever were to be an edge case where the adjustment of the `plugin.xml` causes an issue with the build, we would only notice after merging the PR onto the master, thereby causing the master to go red.

PR #860, opened 2020-03-03
[INTERNAL][CORE] Remove text edit listener from FollowModeManager

tobous:
Removes the logic listening for text edits and adjusting the viewport
accordingly from the FollowModeManager.

As the comment on the corresponding method already states, the listener
is not necessary as the viewport will be updated anyways for the text
selection activity caused by the caret movement (that in turn is caused
by the text edit).

The only advantage this logic would offer is that it speeds up the
viewport adjustment. As the selection activity is processed after the
text edit activity, the text edit will be applied before the viewport is
adjusted. But, as the delay is very short and the edited text will be
highlighted anyways, this advantage is negligible.

### DIFF
--- a/core/src/saros/editor/FollowModeManager.java
+++ b/core/src/saros/editor/FollowModeManager.java
@@ -10,7 +10,6 @@ import saros.activities.IActivity;
 import saros.activities.SPath;
 import saros.activities.StartFollowingActivity;
 import saros.activities.StopFollowingActivity;
-import saros.activities.TextEditActivity;
 import saros.activities.TextSelectionActivity;
 import saros.activities.ViewportActivity;
 import saros.context.IContextFactory;
@@ -171,21 +170,6 @@ public class FollowModeManager implements Startable {
         @Override
         public void receive(TextSelectionActivity activity) {
           TextSelection selection = new TextSelection(activity.getOffset(), activity.getLength());
-
-          editorManager.adjustViewport(activity.getPath(), followeeRange(), selection);
-        }
-
-        /**
-         * {@inheritDoc}
-         *
-         * <p>TODO Find out whether this is actually necessary. If we receive a
-         * TextSelectionActivity for each cursor movement, then we don't need to listen for edits as
-         * well.
-         */
-        @Override
-        public void receive(TextEditActivity activity) {
-          int cursorOffset = activity.getOffset() + activity.getText().length();
-          TextSelection selection = new TextSelection(cursorOffset, 0);
 
           editorManager.adjustViewport(activity.getPath(), followeeRange(), selection);
         }


tobous:
STF also passes: https://github.com/saros-project/saros/runs/481946353

srossbach:
Looking at the logic I think it handles cases of C&P stuff. Should we really remove it ? (I have not checked to logic of adjustViewport)

tobous:
Can you explain how this is related to copy and pasting stuff?

To be honest, my base reason to remove this is that it creates complications for the refactoring I am currently working on.

The refactoring moves the selection handling to a line and in-line offset based system (from the text offset based handling).
This method is an issue as it is in the core (meaning it can't resolve line and in-line offsets to real offsets easily) and tries to use location information from text edit activities (which will be migrated in a follow-up refactoring) to create a selection, meaning a conversion between the two positioning schemes is necessary. As a result, I would have to comment out this part of the code until the refactoring migrating the text activities to the position scheme is done.

But, as I had a closer look at the logic, I didn't see much of a purpose for it. As the comment already states, active text edits by the user should always be accompanied by a cursor movement.

Furthermore, I found some issues with the logic. I don't think it always works correctly. It works on the assumption that, if we receive a text edit, the caret/cursor of the user that created the activity must be at that position. This is, however, not always true. If, for example, the text edit was caused by a refactoring, the cursor of the other user might be nowhere near the changed text. (This is probably dependent on how refactorings are implemented by the IDE. But moving the cursor to every place that needs to be adjusted as part of the refactoring does not seem like a likely implementation to me.)

As a result (again, depending on how the refactoring logic is implemented in the local IDE), the follower might end up looking at a random part of the file, just because it was the last one edited by the refactoring.


Similar logic also exists in Eclipse in the [EditorManager](https://github.com/saros-project/saros/blob/master/eclipse/src/saros/editor/EditorManager.java#L718). Here, it has the same issue with refactorings, etc. but at least serves a purpose: It ensures that the cursor annotation is always in-sync with the edit. Otherwise, the cursor annotation would slightly lag behind the actual text change, looking a bit weird.

But this is also just a minor issue, so I also thought about removing it. But I wanted to discuss it first as it serves at least some purpose. (Also, the Eclipse logic doesn't bother me for the refactoring as it is located in the Eclipse part, meaning it can easily do the conversion.)

tobous:
Regarding the C&P stuff: As far as I can tell, pasting always inserts at the current cursor position, regardless of where I right-click to open the context menu and select paste. That is the only problem I could think of, so I don't see the issue with it.

srossbach:
I also had similar thoughts about the refactoring stuff. And looking at C&P. When I insert something my cursor also do not move. Looking at the TextSelectionActivity it clearly states the a selection with length 0 equals the current cursor position in the document.

PR #865, opened 2020-03-05
[CORE] XStream: Use NamedMapConverter

stefaus:
Set XStream map conversion format to `<map><entry key="#key>" value="#value"/></map>`
instead of `<map><entry><string>#key</string><string>#value</string></entry></map>`
to increase readability and reduce size of saros messages.

Breaks parsing compatibility with older clients.

### DIFF
--- a/core/src/saros/misc/xstream/XStreamExtensionProvider.java
+++ b/core/src/saros/misc/xstream/XStreamExtensionProvider.java
@@ -25,6 +25,7 @@ import com.thoughtworks.xstream.annotations.XStreamOmitField;
 import com.thoughtworks.xstream.converters.Converter;
 import com.thoughtworks.xstream.converters.SingleValueConverter;
 import com.thoughtworks.xstream.converters.basic.BooleanConverter;
+import com.thoughtworks.xstream.converters.extended.NamedMapConverter;
 import com.thoughtworks.xstream.io.xml.CompactWriter;
 import java.io.IOException;
 import java.io.StringWriter;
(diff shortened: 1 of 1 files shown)


core/src/saros/misc/xstream/XStreamExtensionProvider.java
```diff
@@ -99,6 +100,27 @@ public XStreamExtensionProvider(String namespace, String elementName, Class<?>..
...
+
+    /*
+     * set map conversion format to
+     * <map><entry key="#key>" value="#value"/></map>
+     * instead of
+     * <map><entry><string>#key</string><string>#value</string></entry></map>
+     *
+     * breaks parsing compatibility with older clients if changed
+     */
+    xstream.registerConverter(
```
srossbach: If you really want to save byte you could use:
```
new NamedMapConverter(xstream.getMapper(), "entry", "key", String.class, null, Integer.class, true, false, xstream.getConverterLookup());

 <map>
   <entry key="keyValue">0</entry>
 </map>
 ```
 
But what if we are using maps of type string-object or integer-object where object is of type string, number, or boolean ?

stefaus:
Ok this bricks handling of Maps different than String, String...
I expected a different handling when I read the xstream doc and think including the data type into to the xml output is stupid and avoidable, but maybe they have their reasons.

So I close this PR and leave xstream the way it is.

srossbach:
The reason is pretty simple. Lets take for example that we are using a custom number(integer) and boolean converter. The boolean converter converts false to 0 and true to 1. The  number converter converts the number using either hex(base16) or a much higher base.

In the end without additional information it is not possible to unmarshall a Xml map structure of type (x, object) because you do not know if the value represents a string, a boolean or a number.

Currently I think (while nice to have) it is not that important to optimize the map structure, because those packets are not frequently send. 

On the other hand, have you noticed the UrlConverter ? In my opinion this is not needed at all because XStream should already handle the escaping.

PR #866, opened 2020-03-06
Avoid not disappearing contribution annotations

m273d15:
The PR splits all annotations in annotations of size 1 and removes the old splitting logic. This is much easier and fixes the issue. However, I would assume that this was the initial solution in Saros and the split logic was an optimization, but during my tests (manual and with stf) I had not the impression that my change worsens the user experiences.

Furthermore, this slightly changes the history handling, because text insertions are multiple annotation insertions. Therefore a long insertion is only annotated at the end. This leads to the question "How should this feature work?".

@srossbach What are your thoughts? Did I miss a corner case that requires the other handling?

### DIFF
--- a/eclipse/src/saros/editor/EditorManager.java
+++ b/eclipse/src/saros/editor/EditorManager.java
@@ -1055,7 +1055,7 @@ public class EditorManager implements IEditorManager {
       // Try to replace
       try {
         // Attention: This method also alters the annotation model if an
-        // annotation exists at the offset and the length &gt; 1
+        // annotation exists at the offset and the length > 1
         doc.replace(offset, replacedText.length(), text);
       } catch (BadLocationException e) {
         LOG.error(
--- a/eclipse/src/saros/editor/EditorManager.java
+++ b/eclipse/src/saros/editor/EditorManager.java
@@ -1074,9 +1074,6 @@ public class EditorManager implements IEditorManager {
           contributionAnnotationManager.insertAnnotation(model, offset, text.length(), source);
         }
       }
-
-      IAnnotationModel model = provider.getAnnotationModel(input);
-      contributionAnnotationManager.insertAnnotation(model, offset, text.length(), source);
     } finally {
       provider.disconnect(input);
     }
--- a/eclipse/src/saros/editor/EditorManager.java
+++ b/eclipse/src/saros/editor/EditorManager.java
@@ -633,14 +633,6 @@ public class EditorManager implements IEditorManager {
 
     // inform all registered ISharedEditorListeners about this text edit
     editorListenerDispatch.textEdited(textEdit);
-
-    /*
-     * TODO Investigate if this is really needed here
-     */
-    IEditorInput input = changedEditor.getEditorInput();
-    IDocumentProvider provider = EditorAPI.getDocumentProvider(input);
-    IAnnotationModel model = provider.getAnnotationModel(input);
-    contributionAnnotationManager.splitAnnotation(model, offset);
   }
 
   private void execEditorActivity(EditorActivity editorActivity) {
--- a/eclipse/src/saros/editor/EditorManager.java
+++ b/eclipse/src/saros/editor/EditorManager.java
@@ -1054,6 +1054,8 @@ public class EditorManager implements IEditorManager {
 
       // Try to replace
       try {
+        // Attention: This method also alters the annotation model if an
+        // annotation exists at the offset and the length &gt; 1
         doc.replace(offset, replacedText.length(), text);
       } catch (BadLocationException e) {
         LOG.error(
--- a/eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java
+++ b/eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java
@@ -2,6 +2,7 @@ package saros.editor.internal;
 
 import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Queue;
@@ -85,6 +86,10 @@ class ContributionAnnotationHistory {
             + oldAnnotation
             + " in the current history for user: "
             + oldAnnotation.getSource());
+  }
+
+  protected void addNewEntry(ContributionAnnotation annotation) {
+    addNewEntry(Arrays.asList(annotation));
   }
 
   protected void addNewEntry(List<ContributionAnnotation> annotations) {
--- a/eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java
+++ b/eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java
@@ -2,7 +2,6 @@ package saros.editor.internal;
 
 import java.util.ArrayDeque;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Queue;
@@ -12,70 +11,59 @@ import org.apache.log4j.Logger;
 import org.eclipse.jface.text.source.IAnnotationModel;
 import saros.editor.annotations.ContributionAnnotation;
 
-/** */
+/** A history for the contribution annotations that has a fixed-size and uses a FIFO approach. */
 class ContributionAnnotationHistory {
 
   private static final Logger log = Logger.getLogger(ContributionAnnotationHistory.class);
-  Queue<List<ContributionAnnotation>> queue;
+  private Queue<List<ContributionAnnotation>> queue;
   private int maxHistoryLength;
 
-  protected ContributionAnnotationHistory(int maxHistoryLength) {
+  ContributionAnnotationHistory(final int maxHistoryLength) {
     this.queue = new ArrayDeque<>(maxHistoryLength);
     this.maxHistoryLength = maxHistoryLength;
   }
 
+  /** @return current size of the history */
+  int getSize() {
+    return queue.size();
+  }
+
   /**
    * Removes all entries in the history and returns an aggregated list of all contribution
-   * annotations in the history.
+   * annotations in the history
    *
-   * @return A flattened all history entries
+   * @return a flattened list of all history entries
    */
-  protected List<ContributionAnnotation> clear() {
-    List<ContributionAnnotation> content = new ArrayList<>();
+  List<ContributionAnnotation> clear() {
+    final List<ContributionAnnotation> content = new ArrayList<>();
     while (!queue.isEmpty()) {
       content.addAll(queue.remove());
     }
     return content;
   }
 
-  /** If the history */
-  protected Pair<IAnnotationModel, List<ContributionAnnotation>> removeHistoryEntry() {
-
-    Pair<IAnnotationModel, List<ContributionAnnotation>> removedEntry = null;
-
-    if (queue.size() == maxHistoryLength) {
-      List<ContributionAnnotation> annotations = queue.remove();
-      IAnnotationModel model = annotations.get(0).getModel();
-      for (ContributionAnnotation a : annotations) {
-        if (a.getModel() != model) throw new RuntimeException("Different Models");
-      }
-      removedEntry = new ImmutablePair<>(model, annotations);
-    }
-
-    return removedEntry;
-  }
-
-  /*
-   * Replaces the an annotation in the history with another one.
-   * E.g. if we want to replace the annotation D with D':
-   * [[A, B], [C], [D, E], [F], [G]] -> [[A, B], [C], [D', E], [F], [G]]
+  /**
+   * Replaces an annotation in the history with another one. E.g. if we want to replace the
+   * annotation D with D': [[A, B], [C], [D, E], [F], [G]] -> [[A, B], [C], [D', E], [F], [G]]
    *
+   * @param oldAnnotation the annotation that has to be replaced.
+   * @param newAnnotation the annotation that has to be inserted.
    */
-  protected void replaceInHistory(
+  void replaceInHistory(
       final ContributionAnnotation oldAnnotation, final ContributionAnnotation newAnnotation) {
 
-    for (List<ContributionAnnotation> entry : queue) {
+    for (final List<ContributionAnnotation> entry : queue) {
 
-      for (final ListIterator<ContributionAnnotation> annotationsLit = entry.listIterator();
-          annotationsLit.hasNext(); ) {
-        final ContributionAnnotation annotation = annotationsLit.next();
+      for (final ListIterator<ContributionAnnotation> annotationsListIt = entry.listIterator();
+          annotationsListIt.hasNext(); ) {
+        final ContributionAnnotation annotation = annotationsListIt.next();
 
         if (annotation.equals(oldAnnotation)) {
-          annotationsLit.remove();
+          annotationsListIt.remove();
 
           assert oldAnnotation.getSource().equals(newAnnotation.getSource());
 
-          annotationsLit.add(newAnnotation);
+          annotationsListIt.add(newAnnotation);
           return;
         }
       }
@@ -88,16 +76,30 @@ class ContributionAnnotationHistory {
             + oldAnnotation.getSource());
   }
 
-  protected void addNewEntry(ContributionAnnotation annotation) {
-    addNewEntry(Arrays.asList(annotation));
+  /**
+   * @param annotations annotations to add as one entry into the history
+   * @return If an entry has to be removed from the history in order to add the new one, the removed
+   *     entry is returned with the corresponding {@link IAnnotationModel}. Otherwise return <code>
+   *     null</code>.
+   */
+  Pair<IAnnotationModel, List<ContributionAnnotation>> addNewEntry(
+      final List<ContributionAnnotation> annotations) {
+
+    Pair<IAnnotationModel, List<ContributionAnnotation>> removedEntry = removeEntryIfFull();
+    queue.add(annotations);
+    return removedEntry;
   }
 
-  protected void addNewEntry(List<ContributionAnnotation> annotations) {
-    if (queue.size() >= maxHistoryLength) {
-      throw new IllegalStateException(
-          "The queue already contains the " + "allowed number of annotations.");
+  private Pair<IAnnotationModel, List<ContributionAnnotation>> removeEntryIfFull() {
+
+    Pair<IAnnotationModel, List<ContributionAnnotation>> removedEntry = null;
+
+    if (queue.size() == maxHistoryLength) {
+      final List<ContributionAnnotation> annotations = queue.remove();
+      final IAnnotationModel model = annotations.get(0).getModel();
+      removedEntry = new ImmutablePair<>(model, annotations);
     }
 
-    queue.add(annotations);
+    return removedEntry;
   }
 }
--- a/eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java
+++ b/eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java
@@ -1,0 +1,98 @@
+package saros.editor.internal;
+
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.Queue;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.log4j.Logger;
+import org.eclipse.jface.text.source.IAnnotationModel;
+import saros.editor.annotations.ContributionAnnotation;
+
+/** */
+class ContributionAnnotationHistory {
+
+  private static final Logger log = Logger.getLogger(ContributionAnnotationHistory.class);
+  Queue<List<ContributionAnnotation>> queue;
+  private int maxHistoryLength;
+
+  protected ContributionAnnotationHistory(int maxHistoryLength) {
+    this.queue = new ArrayDeque<>(maxHistoryLength);
+    this.maxHistoryLength = maxHistoryLength;
+  }
+
+  /**
+   * Removes all entries in the history and returns an aggregated list of all contribution
+   * annotations in the history.
+   *
+   * @return A flattened all history entries
+   */
+  protected List<ContributionAnnotation> clear() {
+    List<ContributionAnnotation> content = new ArrayList<>();
+    while (!queue.isEmpty()) {
+      content.addAll(queue.remove());
+    }
+    return content;
+  }
+
+  /** If the history */
+  protected Pair<IAnnotationModel, List<ContributionAnnotation>> removeHistoryEntry() {
+
+    Pair<IAnnotationModel, List<ContributionAnnotation>> removedEntry = null;
+
+    if (queue.size() == maxHistoryLength) {
+      List<ContributionAnnotation> annotations = queue.remove();
+      IAnnotationModel model = annotations.get(0).getModel();
+      for (ContributionAnnotation a : annotations) {
+        if (a.getModel() != model) throw new RuntimeException("Different Models");
+      }
+      removedEntry = new ImmutablePair<>(model, annotations);
+    }
+
+    return removedEntry;
+  }
+
+  /*
+   * Replaces the an annotation in the history with another one.
+   * E.g. if we want to replace the annotation D with D':
+   * [[A, B], [C], [D, E], [F], [G]] -> [[A, B], [C], [D', E], [F], [G]]
+   *
+   */
+  protected void replaceInHistory(
+      final ContributionAnnotation oldAnnotation, final ContributionAnnotation newAnnotation) {
+
+    for (List<ContributionAnnotation> entry : queue) {
+
+      for (final ListIterator<ContributionAnnotation> annotationsLit = entry.listIterator();
+          annotationsLit.hasNext(); ) {
+        final ContributionAnnotation annotation = annotationsLit.next();
+
+        if (annotation.equals(oldAnnotation)) {
+          annotationsLit.remove();
+
+          assert oldAnnotation.getSource().equals(newAnnotation.getSource());
+
+          annotationsLit.add(newAnnotation);
+          return;
+        }
+      }
+    }
+
+    log.warn(
+        "could not find annotation "
+            + oldAnnotation
+            + " in the current history for user: "
+            + oldAnnotation.getSource());
+  }
+
+  protected void addNewEntry(List<ContributionAnnotation> annotations) {
+    if (queue.size() >= maxHistoryLength) {
+      throw new IllegalStateException(
+          "The queue already contains the " + "allowed number of annotations.");
+    }
+
+    queue.add(annotations);
+  }
+}
--- a/eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
+++ b/eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
@@ -1,16 +1,15 @@
 package saros.editor.internal;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
-import java.util.ListIterator;
 import java.util.Map;
 import java.util.Set;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.log4j.Logger;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.jface.text.Position;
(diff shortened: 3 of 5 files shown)


saros-infrastructure:


```diff

Complexity decreasing per file
==============================
+ eclipse/src/saros/editor/internal/ContributionAnnotationManager.java  -2
         

Clones removed
==============
+ eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5134601&bid=16871785)

srossbach:
That is the big question. How should it work ? If a C&P should be fully annotated then you current solution could be a real performance nightmare. Adding and removing annotations is more CPU intensive then playing a modern computer game. Although we are already optimized it in such way that activities are executed all at once on the EDT if the CPU is not fast enough. If you CPU is fast enough CPU load goes up to 100% of one CPU die.

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -124,9 +124,17 @@ public void insertAnnotation(IAnnotationModel model, int offset, int length, Use
 
     if (!contribtionAnnotationsEnabled) return;
 
-    if (length < 0) // why is 0 len allowed ?
-    return;
+    for (int i = 0; i < length; i++) {
```
srossbach: My big guess is that if you C&P a long enough text, the receiving side will freeze.

m273d15: What do you think about checking `length > MAX_HISTORY_LENGTH` and just annotating the the last `MAX_HISTORY_LENGTH`s characters? Is a simple optimization that reduces the issue.

srossbach: @tobous How are annotations handled in IntelliJ regarding C&P, e.g one insert = 50 characters ?

And @m273d15 if you want to make mass updates in the annotation model, please calculate the added and removed annotations before you actually replace them. We have a helper function that tries to that at once, i.e only making 1 update call for all affected annotations.

eclipse/src/saros/editor/EditorManager.java
```diff
@@ -1079,12 +1071,11 @@ private void replaceText(SPath path, int offset, String replacedText, String tex
           ITextEditor textEditor = (ITextEditor) editorPart;
           IAnnotationModel model =
               textEditor.getDocumentProvider().getAnnotationModel(textEditor.getEditorInput());
+
           contributionAnnotationManager.insertAnnotation(model, offset, text.length(), source);
         }
       }
 
-      IAnnotationModel model = provider.getAnnotationModel(input);
```
tobous: Do you know why there was a separate handling beforehand? I.e. why is this no longer necessary?

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
tobous: In IntelliJ, contribution annotations are handled similarly but not exactly in the way implemented here.

The base approach is the same: All contribution annotations are split into one character wide annotations that do not auto-extend (i.e. don't extend when text is added at either end). See [AnnotationManager.addContributionAnnotation(...)](https://github.com/saros-project/saros/blob/master/intellij/src/saros/intellij/editor/annotations/AnnotationManager.java#L120).
I made this decision to avoid such cases where another user types in the middle of an annotation, thereby extending it if it were a continuously annotated block.

Where the handling seems to differ is that all text added by one activity (like a C&P) is still seen as one annotation, even though it is displayed in many different parts. See [ContributionAnnotation](https://github.com/saros-project/saros/blob/master/intellij/src/saros/intellij/editor/annotations/ContributionAnnotation.java). One annotation has a list of `AnnotationRange` objects, which each represent a highlighted section of text.
As a result, all highlightings for the action are bundled and seen as one annotation for the annotation rotation.

In IntelliJ, the performance for this is ok. (I haven't measured the impact.) The funny thing is that, even without Saros, C&P a large chunk (~2000 line) in a VERY large file (~250000 lines) freezes the local editor. Such issues, however, do not occur on the receiving side of such activities. As a result, during a running session, the remove site (which has to deal with all of the annotations but still runs without a hitch) is done far before the sending site.

srossbach: Ok so the logic what is displayed is the same as in Eclipse. I do not see a problem by just using single annotations in Eclipse too. But as I already said, use the bulk method.

eclipse/src/saros/editor/EditorManager.java
srossbach: Nobody really knows that. As I already said the logic is buggy, annotations are inserted twice (the contribution manager silently ignores it).

Guess it makes more sense either to remove this line or count the editors in the loop above and only execute this code here if you annotated 0 editors.

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
srossbach: My last 2 cents. It should not be that hard to modify the history in such way that it stores multiples annotations.

```
List annotationsToRemove = history.getAndRemoveAnnotationsForOldestEntry();
List annotationsToAdd = createAnnotationsForContributionRange(....)
history.addEntry(annotationsToAdd)
model.replaceAnnotations(annotationsToRemove, annotationsToAdd)
```

eclipse/src/saros/editor/EditorManager.java
srossbach: And I guess nobody know if annotation removal from the history is correctly working when the annotation is used in multiple editors (e.g splitview)

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
m273d15: I will modify the history handling and adopt the `AnnotationModelHelper.replaceAnnotationsInModel` method (I think this is the bulk method you talked about).

saros-infrastructure:


```diff

Issues
======
+ Solved 1
           

Complexity decreasing per file
==============================
+ eclipse/src/saros/editor/internal/ContributionAnnotationManager.java  -2
         

Clones removed
==============
+ eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5134601&bid=16871785)

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -122,73 +108,72 @@ public ContributionAnnotationManager(
...
-
-    addContributionAnnotation(
-        new ContributionAnnotation(source, model), new Position(offset, length));
+    annotationModelHelper.replaceAnnotationsInModel(model, annotationsToRemove, annotationsToAdd);
   }
 
   /**
-   * Splits the contribution annotation at given position, so that a following text change will not
-   * expand the annotation. This needs to be called before the text is changed.
+   * Creates one contribution annotations with length 1 for each char contained in the range defined
```
srossbach: Creates contribution annotations

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -122,73 +108,72 @@ public ContributionAnnotationManager(
...
+   * by {@code offset} and {@code length}. The annotation is inserted to given model if there is not
```
srossbach: This method only performs the calculation, it does not alter the model.

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -122,73 +108,72 @@ public ContributionAnnotationManager(
...
+   * inserted to given model if there is not already a contribution annotation at given position.
+   *
+   * @param model to add the annotation to.
+   * @param offset start of the annotation to add.
+   * @param source of the annotation.
+   * @returns a pair containing the annotation and its position.
+   */
+  private Pair<ContributionAnnotation, Position> createPositionedAnnotation(
+      IAnnotationModel model, int offset, User source) {
+    final int ANNOTATION_SIZE = 1;
```
srossbach: ANNOTATION_LENGTH ?

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
srossbach: The complexity of the algorithm went up from O(m) to O(m*n). m = number of annotations in the model, n = characters / annotations to add.

Furthermore is it that tragic if multiple annotations are present for a single position ? (In reality it should not happen at all, but it does see the issue in the EditorManager)

As you already know the covered range and if you want to keep the check this algorithm can be optimized to have an average runtime of O(m) again, e.g calculate the range, traverse the annotations only once and collect all that do collide, calculate the colliding annotations afterwards.

eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java
```diff
@@ -232,19 +155,49 @@ public void testRemoveAnnotationsWhenUserLeaves() {
...
 
-      assertEquals(
-          ContributionAnnotationManager.MAX_HISTORY_LENGTH * users.size(),
-          getAnnotationCount(model));
+      assertEquals(MAX_HISTORY_LENGTH * users.size(), getAnnotationCount(model));
     }
   }
 
+  @Test
+  public void testInsertAnnotationWithLengthGreaterOne() {
```
srossbach: I do not know if this is sufficient or to just introduce a test that combines multiple annotations lengths. Currently all other tests are only working with a length of 1.

eclipse/src/saros/editor/EditorManager.java
```diff
@@ -633,14 +633,6 @@ void textAboutToBeChanged(int offset, String text, int replaceLength, IDocument
...
     // inform all registered ISharedEditorListeners about this text edit
     editorListenerDispatch.textEdited(textEdit);
-
-    /*
-     * TODO Investigate if this is really needed here
-     */
-    IEditorInput input = changedEditor.getEditorInput();
-    IDocumentProvider provider = EditorAPI.getDocumentProvider(input);
-    IAnnotationModel model = provider.getAnnotationModel(input);
-    contributionAnnotationManager.splitAnnotation(model, offset);
```
srossbach: Can you please have a look at the conversation history ? I already talked about another issue in this file on line 1086.

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -389,4 +327,69 @@ private void removeFromHistoryAndAnnotationModel(final Collection<History> histo
               a instanceof ContributionAnnotation
                   && users.contains(((ContributionAnnotation) (a)).getSource()));
   }
+
+  static final class History {
```
srossbach: Any reason to not move the code to a separate class ?

saros-infrastructure:


```diff

Issues
======
+ Solved 2
- Added 2
           

Complexity increasing per file
==============================
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  2
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationHistoryTest.java  2
- eclipse/src/saros/editor/internal/ContributionAnnotationManager.java  1
- eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java  5
         

Clones removed
==============
+ eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5134601&bid=16871785)

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -122,73 +108,72 @@ public ContributionAnnotationManager(
    */
   public void insertAnnotation(IAnnotationModel model, int offset, int length, User source) {
 
-    if (!contribtionAnnotationsEnabled) return;
-
-    if (length < 0) // why is 0 len allowed ?
-    return;
+    if (!contribtionAnnotationsEnabled || length < 0) return;
```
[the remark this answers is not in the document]
m273d15: insertAnnotation is called by EditorManager#replaceText, which is also called if a deletion of a char is performed. This leads to length == 0.I would like to change the handling, but I had to stop with refactoring and rework because the PR would be too large.

I restructured the code in order to make the cases more clear.

eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java
m273d15: I added the test "testHistoryRemovelWithDifferentLengths"

m273d15:
@srossbach I was not able to directly answer these comments:

> Guess it makes more sense either to remove this line or count the editors in the loop above and only execute this code here if you annotated 0 editors.

I removed the additional insert of annotation in the EditorManager and removed the check in the ContributionAnnotationManager.

> And I guess nobody know if annotation removal from the history is correctly working when the annotation is used in multiple editors (e.g splitview)

I fixed the splitview handling. The part that removed the old annotation from the model used the wrong model.

saros-infrastructure:


```diff

Issues
======
+ Solved 2
           

Complexity increasing per file
==============================
- eclipse/src/saros/editor/internal/ContributionAnnotationManager.java  1
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationHistoryTest.java  2
- eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java  4
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  2
         

Clones removed
==============
+ eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5134601&bid=16871785)

eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java
```diff
@@ -300,4 +357,24 @@ private int getAnnotationCount(IAnnotationModel model) {
 
     return positions;
   }
+
+  private User createAliceTestUser() {
```
tobous: As the returned object is always the same, these methods could also be replaced by a static variable for each user. Or are the user objects modified by the tests?

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -112,83 +97,99 @@ public ContributionAnnotationManager(
...
-        replaceInHistory(contributionAnnotation, beforeAnnotation, afterAnnotation);
-      }
-    }
+  /**
+   * Creates a contribution annotations with length 1 at position {@code offset}.
+   *
+   * @param model to add the annotation to.
+   * @param offset start of the annotation to add.
+   * @param source of the annotation.
+   * @returns a pair containing the annotation and its position or <code>null</code> if an
```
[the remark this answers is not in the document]
srossbach: The method no longer returns null because the collision check was removed.

eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java
```diff
@@ -0,0 +1,100 @@
...
+    return removedEntry;
+  }
+
+  /*
+   * Replaces the an annotation in the history with another one.
+   * E.g. if we want to replace the annotation D with D':
+   * [[A, B], [C], [D, E], [F], [G]] -> [[A, B], [C], [D', E], [F], [G]]
+   *
+   */
+  protected void replaceInHistory(
```
srossbach: No longer used ?

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -367,18 +271,17 @@ private void removeAnnotationsForUser(final User user) {
...
 
-    for (final History history : histories) {
-      while (!history.entries.isEmpty()) {
-
-        final HistoryEntry entry = history.entries.poll();
-
-        annotationModels.add(entry.annotation.getModel());
-        users.add(entry.annotation.getSource());
+    for (final ContributionAnnotationHistory history : histories) {
+      List<ContributionAnnotation> allAnnotationsInHistory = history.clear();
```
srossbach: Just my 2 cents: Either use the final key constantly or do not use it at all. The code looks pretty weird and you can already see without using GIT Blame that it was written by at least two different developers.

saros-infrastructure:


```diff

Issues
======
+ Solved 2
           

Complexity increasing per file
==============================
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationHistoryTest.java  2
- eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java  4
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  2
         

Complexity decreasing per file
==============================
+ eclipse/src/saros/editor/internal/ContributionAnnotationManager.java  -2
         

Clones removed
==============
+ eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5134601&bid=16871785)

m273d15:
@tobous Based on your comments, I changed the "insertAnnotation" method to directly return if length <= 0. With this change it was possible to merge the "removeEntries()" and "addNewEntry()" method and I was able to refactor the history again. Furthermore, I changed the tests based on your comments.

I added all missing final keyword automatically with IntelliJ.

eclipse/test/junit/saros/editor/internal/ContributionAnnotationHistoryTest.java
```diff
@@ -0,0 +1,86 @@
...
+    for (int i = 0; i < MAX_HISTORY_LENGTH; i++) {
+      annotationsToRemove = history.addNewEntry(createDummyAnnotation(testModel));
+      assertNull(annotationsToRemove);
+    }
+
+    assertEquals(MAX_HISTORY_LENGTH, history.getSize());
+  }
+
+  @Test
+  public void testaddEntriesExceedingMaxLength() {
```
tobous:
```suggestion
  public void testAddEntriesExceedingMaxLength() {
```

eclipse/test/junit/saros/editor/internal/ContributionAnnotationHistoryTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    for (int i = 1; i < MAX_HISTORY_LENGTH; i++) {
+      annotationsToRemove = history.addNewEntry(createDummyAnnotation(testModel));
+      assertNull(annotationsToRemove);
+    }
+
+    assertEquals(MAX_HISTORY_LENGTH, history.getSize());
+
+    annotationsToRemove = history.addNewEntry(createDummyAnnotation(testModel));
+    assertThat(annotationsToRemove.getRight(), is(firstEntry));
```
tobous: Ok, this does work as `ContributionAnnotation` does not implement an equals method, meaning all comparisons are done based on object references anyways. But this could be confusing when first reading the test. If it were implementing equals, the test would not ensure the correct functionality as all dummy annotations contain the same content, meaning they would be seen as equal.

So this adjustment is not strictly necessary, but I would still suggest explicitly comparing the object references using `assertSame(...)` just to avoid such confusion (or adding a comment explaining the situation).

eclipse/test/junit/saros/editor/internal/ContributionAnnotationHistoryTest.java
```diff
@@ -0,0 +1,86 @@
...
+    for (int i = 1; i < MAX_HISTORY_LENGTH; i++) {
+      annotationsToRemove = history.addNewEntry(createDummyAnnotation(testModel));
+      assertNull(annotationsToRemove);
+    }
+
+    annotationsToRemove = history.addNewEntry(createDummyAnnotation(testModel));
+    assertEquals(testModel2, annotationsToRemove.getLeft());
+
+    annotationsToRemove = history.addNewEntry(createDummyAnnotation(testModel));
+    assertEquals(testModel, annotationsToRemove.getLeft());
```
tobous: Could also use `assertSame(...)`.

eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java
```diff
@@ -85,16 +76,35 @@ protected void replaceInHistory(
...
+   */
+  Pair<IAnnotationModel, List<ContributionAnnotation>> addNewEntry(
+      final List<ContributionAnnotation> annotations) {
 
-  protected void addNewEntry(List<ContributionAnnotation> annotations) {
-    if (queue.size() >= maxHistoryLength) {
+    if (queue.size() > maxHistoryLength) {
       throw new IllegalStateException(
-          "The queue already contains the " + "allowed number of annotations.");
+          "The queue already contains the allowed number of annotations.");
```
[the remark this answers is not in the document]
m273d15: I don't know why I included the check. I wanted to remove the if block.

saros-infrastructure:


```diff

Issues
======
+ Solved 2
           

Complexity increasing per file
==============================
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  2
- eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java  4
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationHistoryTest.java  2
         

Complexity decreasing per file
==============================
+ eclipse/src/saros/editor/internal/ContributionAnnotationManager.java  -2
         

Clones removed
==============
+ eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5134601&bid=16871785)

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java  4
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  2
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationHistoryTest.java  2
         

Complexity decreasing per file
==============================
+ eclipse/src/saros/editor/internal/ContributionAnnotationManager.java  -2
         

Clones removed
==============
+ eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5134601&bid=16871785)

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationHistoryTest.java  2
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  2
- eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java  4
         

Complexity decreasing per file
==============================
+ eclipse/src/saros/editor/internal/ContributionAnnotationManager.java  -1
         

Clones removed
==============
+ eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5134601&bid=16871785)

m273d15:
@tobous I rebased and added [this commit](https://github.com/saros-project/saros/pull/866/commits/0d13ed7b831139a2b959959fc9b9f81a01277ac7) in order to avoid the corner case:
* paste at a position where an annotation exists
* the IDocument#replace method resizes the corresponding annotations
The new commit removes the resized annotation.

eclipse/src/saros/editor/EditorManager.java
```diff
@@ -1062,6 +1054,8 @@ private void replaceText(SPath path, int offset, String replacedText, String tex
 
       // Try to replace
       try {
+        // Attention: This method also alters the annotation model if an
+        // annotation exists at the offset and the length &gt; 1
```
tobous: Use correct symbol instead of character code?
```suggestion
        // annotation exists at the offset and the length > 1
```

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -33,32 +34,20 @@
  * they belong to are deleted, the session ends, or their respective author leaves the session. To
  * avoid cluttering the editors, only the last {@value #MAX_HISTORY_LENGTH} changes are annotated.
  *
+ * <p>The invariant of the class should be that all handled annotations have a length &lt;= 1. This
```
tobous:
```suggestion
 * <p>The invariant of the class should be that all handled annotations have a length <= 1. This
```

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -112,102 +101,151 @@ public ContributionAnnotationManager(
...
 
-    if (!contribtionAnnotationsEnabled) return;
+    // Return if length == 0, (called after a deletion was performed)
+    if (!contribtionAnnotationsEnabled || length <= 0) return;
 
-    if (length < 0) // why is 0 len allowed ?
-    return;
+    if (length > 1) {
+      // Correct current model state to match with the assumption that all
+      // annotations have length &lt;= 1.
```
tobous:
```suggestion
      // annotations have length <= 1.
```

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -112,102 +101,151 @@ public ContributionAnnotationManager(
...
+    else {
+      annotationModelHelper.replaceAnnotationsInModel(
+          model, Collections.emptyList(), annotationsToAdd);
+    }
   }
 
   /**
-   * Splits the contribution annotation at given position, so that a following text change will not
-   * expand the annotation. This needs to be called before the text is changed.
+   * This method is only called if an annotation with length &gt; 1 should be inserted, because if
```
tobous: In actual javadoc, this style of using `<` and `>` technically works when viewed through a tool interpreting the javadoc but is still unnecessarily hard to read when looking at the code.
```suggestion
   * This method is only called if an annotation with length > 1 should be inserted, because if
```

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -112,102 +101,151 @@ public ContributionAnnotationManager(
...
+   * text is pasted at a position that already contains an annotation, the content change on the
+   * {@link IDocument} that is performed by the caller of {@link #insertAnnotation} leads to an
+   * internal change of the annotation model.
+   *
+   * <p>The annotation with length &gt; 1 is removed, because the new annotation is subsequently
```
tobous:
```suggestion
   * <p>The annotation with length > 1 is removed, because the new annotation is subsequently
```

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -112,102 +101,151 @@ public ContributionAnnotationManager(
...
+   * added by {@link #insertAnnotation}
    *
-   * @param model to search for annotations to split
-   * @param offset at which annotations should be split
+   * <p>Therefore, this methods works against the internal mechanisms of eclipse in order to enforce
+   * the invariant of annotations with have a length &lt;= 1.
```
tobous:
```suggestion
   * the invariant of annotations with have a length <= 1.
```

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -112,102 +101,151 @@ public ContributionAnnotationManager(
...
-      final User source = contributionAnnotation.getSource();
-
-      final Position pos = model.getPosition(contributionAnnotation);
+      ContributionAnnotation contrAnnotation = (ContributionAnnotation) annotation;
+      Position p = model.getPosition(contrAnnotation);
+      if (p.getLength() > 1) {
+        annotationsToRemove.add(contrAnnotation);
 
-      if ((offset > pos.offset) && (offset < pos.offset + pos.length)) {
+        // Expect only one annotation with length &gt; 1 created by
```
tobous:
```suggestion
        // Expect only one annotation with length > 1 created by
```

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
```diff
@@ -112,102 +101,151 @@ public ContributionAnnotationManager(
...
+   * Creates contribution annotations with length 1 for each char contained in the range defined by
+   * {@code offset} and {@code length}.
+   *
+   * @param model the model that is assigned to the created contribution annotations.
+   * @param offset start of the annotation range.
+   * @param length length of the annotation range.
+   * @param source of the annotation.
+   * @return a map containing the annotations and their positions
+   */
+  private Map<ContributionAnnotation, Position> createAnnotationsForContributionRange(
```
tobous: So the annotations are present twice in this case, meaning you can just remove the merged one?

Is there currently any logic re-adding existing annotations to the model? If so, can this become an issue in such cases as well (i.e. the internal optimization to merge the annotations is applied again)?

In general, does this have a noticeable impact on performance?

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
m273d15: Hmm, I use this, because the javadoc processor complains about using "<"/">". I would like to have a javadoc build without multiple errors in the future.

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
m273d15: I found this optimization only if the IDocument#replace method is used. Therefore, I assume that this is not an optimization of the model, but the way the replace method introduces annotations. I also tested the re-adding (e.g. if the use color changes).

With using the optimized replaceAnnotations method I noticed no impact, without the optimized version you notice an impact if you paste long text (the annotations are painted bit by bit). Therefore, we use the optimized version.

eclipse/src/saros/editor/internal/ContributionAnnotationManager.java
tobous: I guess then keep it the way it is for the javadoc. The normal comments should still be adjusted in my opinion. And in the javadoc, you could consider writing it out instead. But that is a matter of taste.

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- eclipse/src/saros/editor/internal/ContributionAnnotationHistory.java  4
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  2
- eclipse/test/junit/saros/editor/internal/ContributionAnnotationHistoryTest.java  2
         

Complexity decreasing per file
==============================
+ eclipse/src/saros/editor/internal/ContributionAnnotationManager.java  -1
         

Clones removed
==============
+ eclipse/test/junit/saros/editor/internal/ContributionAnnotationManagerTest.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5134601&bid=16871785)

PR #868, opened 2020-03-06
[FIX][CORE] Fix wrong split operation combination

tobous:
Fixes error in split operation combination logic dealing with two
insertions.

The existing logic incorrectly correlates the two insertions by the
start of the first and end of the second insertion and assumes that they
can then be combined.

When actually applying two activities matching this criteria, the
result, however, is not combinable:
```
Example: ins(5,"cd") + ins(3,"ab")
  Start ins1 = 5
  End ins2 = 5
  -> Defined result: ins(3,"abcd)

Actual result when applying ins1 and ins2:

  position  0123456789
  start:   "          "
  ins1:    "     cd   "
  ins2:    "   ab  cd "
  -> Does not match defined result
```
The actual result can not be defined through a single insertion as it is
not continuous.

Fixes the issue by instead requiring that both insertions have the same
start value.
```
Example: ins(3,"cd") + in(3,"ab")
  Start ins1=3
  Start ins2=3
  -> Defined result: ins(3,"abcd")

  position  0123456789
  start:   "          "
  ins1:    "   cd     "
  ins2:    "   abcd   "
  -> Matches defined result
```
As the adjusted optimization logic is incomplete (i.e. does not cover
all cases), I just picked a similar case to the faulty one for the
adjusted logic. Other cases would also have been possible.

### DIFF
--- a/core/test/junit/saros/concurrent/SplitOperationTest.java
+++ b/core/test/junit/saros/concurrent/SplitOperationTest.java
@@ -21,7 +21,12 @@ import saros.concurrent.jupiter.test.util.PathFake;
 import saros.filesystem.IProject;
 import saros.session.User;
 
-/** testing SplitOperation.toTextEdit() */
+/**
+ * Tests the logic that combines the contained operations if possible when creating text activities
+ * from split operations.
+ *
+ * @see SplitOperation#toTextEdit(SPath, User)
+ */
 public class SplitOperationTest {
 
   protected IProject project;
@@ -72,6 +77,7 @@ public class SplitOperationTest {
     TextEditActivity expected1 = new TextEditActivity(source, 5, "", "abcde", path);
     assertEquals(Collections.singletonList(expected1), split1.toTextEdit(path, source));
 
+    // Del(5,"cde") + Del(5,"ab") -> Del(5,"cdeab")
     Operation split2 = S(D(5, "cde"), D(5, "ab"));
     TextEditActivity expected2 = new TextEditActivity(source, 5, "", "cdeab", path);
     assertEquals(Collections.singletonList(expected2), split2.toTextEdit(path, source));
@@ -84,6 +90,7 @@ public class SplitOperationTest {
     TextEditActivity expected1 = new TextEditActivity(source, 5, "", "cd", path);
     assertEquals(Collections.singletonList(expected1), split1.toTextEdit(path, source));
 
+    // Ins(5,"abcde") + Del(5,"abcd") -> Ins(5,"e")
     Operation split2 = S(I(5, "abcde"), D(5, "abcd"));
     TextEditActivity expected2 = new TextEditActivity(source, 5, "e", "", path);
     assertEquals(Collections.singletonList(expected2), split2.toTextEdit(path, source));
@@ -110,7 +117,7 @@ public class SplitOperationTest {
     // Split(Split(A,B), Split(C,D)) with B and C can be combined
     Operation split = S(S(D(8, "uvw"), I(2, "abcde")), S(D(2, "abcd"), D(15, "xyz")));
 
-    List<TextEditActivity> expected = new LinkedList<TextEditActivity>();
+    List<TextEditActivity> expected = new LinkedList<>();
     expected.add(new TextEditActivity(source, 8, "", "uvw", path));
     expected.add(new TextEditActivity(source, 2, "e", "", path));
     expected.add(new TextEditActivity(source, 15, "", "xyz", path));
@@ -123,7 +130,7 @@ public class SplitOperationTest {
     // Split(Split(A,B), Split(C,D)) with B, C and D can be combined
     Operation split = S(S(D(8, "uvw"), I(2, "abcde")), S(D(2, "abcd"), I(3, "xyz")));
 
-    List<TextEditActivity> expected = new LinkedList<TextEditActivity>();
+    List<TextEditActivity> expected = new LinkedList<>();
     expected.add(new TextEditActivity(source, 8, "", "uvw", path));
     expected.add(new TextEditActivity(source, 2, "exyz", "", path));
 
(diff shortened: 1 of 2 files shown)


srossbach:
Can you give a reference / paper where I(5,"cd") + I(3,"ab") should result in I(3,"abcd) ?

tobous:
> Can you give a reference / paper where I(5,"cd") + I(3,"ab") should result in I(3,"abcd) ?

Ok, it seems like my commit message was not clear enough.

What I was trying to convey is that **the current/previous logic assumes** that `I(5,"cd") + I(3,"ab") -> I(3,"abcd)` holds true (it doesn't).

If you have a look at the previous logic:
```Java
      } else if (insert1.getPosition() == insert2.getPosition() + insert2.getTextLength()) {
        return new InsertOperation(insert2.getPosition(), insert2.getText() + insert1.getText());
      }
```

Applying this to our example `I(5,"cd") + I(3,"ab")`:

Insertion 1 has the start 5. Insertion 2 has the start 3 and the length 2.
-> the if statement is true
-> the resulting insertion starts at the start position of insertion 2 and contains the text of insertion 2 concatenated with the text of insertion 1.
-> ins(3,"abcd")

This assumption can easily be disproven by simply applying the example activities and looking at the result. (copied from PR message)

```
Example: ins(5,"cd") + ins(3,"ab")
  Start ins1 = 5
  End ins2 = 5
  -> Defined result: ins(3,"abcd)

Actual result when applying ins1 and ins2:

  position  0123456789
  start:   "          "
  ins1:    "     cd   "
  ins2:    "   ab  cd "
  -> Does not match defined result
```


The issue seems to be that the author of the initial logic either
- assumed that insertion 2 applies before insertion 1 or
- assumed that insertion 2 is actually a replacement (i.e. replaces the existing text of the given length with the new text instead of just inserting the new text at the given offset).

srossbach:
Just wondering why this happens. Does Jupiter shifts the offsets or is it the IDE ? Looking at the result and assuming spaces in the "line" ins(5,"cd") + ins(3,"ab") would produce I(3, "ab cd") in the IDE.

My current problem is that the original test case do pass but you have replaced it and do no longer test ``` Operation split2 = S(I(6, "0ab"), I(4, "cd"));``` . What should be the result of the two operations after you change ?

tobous:
> Just wondering why this happens. Does Jupiter shifts the offsets or is it the IDE ? Looking at the result and assuming spaces in the "line" ins(5,"cd") + ins(3,"ab") would produce I(3, "ab cd") in the IDE.

It is pretty easy to explain how this happens. The logic was completely written by us, so it is completely independent to any stuff the IDE or Jupiter algorithm does. It is an optimization that is supposed to keep the queue of text activities (generated from the queue of operations; used when transforming remote operations) as short as possible.

Somebody probably just messed up and worked with a wrong assumption. It happens to everyone. :wink:

And, as you already said and as shown in my example, "ab34cd" (where 3 and 4 are the pre-existing characters that were previously in the position where now a and b are) is the correct result.

> My current problem is that the original test case do pass

The original tests passed because they completely tested against their own assumptions. The author wrote the code so that the result is "abcd" and then tested against "abcd", so naturally the test passed.

> My current problem is that the original test case do pass but you have replaced it and do no longer test ` Operation split2 = S(I(6, "0ab"), I(4, "cd"));` . What should be the result of the two operations after you change ?

I removed the old setup as I did not think is was useful. I "implemented" it, but it wasn't very exiting. As already stated, since the resulting insertions are not continuous (they are separated by characters 3 and 4 of the original content), they can't be combined. So the result should be a list of text edit activities containing the activity equal to `new TextEditActivity(source, 6, "0ab", "", path)` and `new TextEditActivity(source, 4, "cd", "", path)`. This is also the actual return value of the method.

But I can re-add this case if you would like.

srossbach:
> It is pretty easy to explain how this happens. The logic was completely written by us, so it is completely independent to any stuff the IDE or Jupiter algorithm does. It is an optimization that is supposed to keep the queue of text activities (generated from the queue of operations; used when transforming remote operations) as short as possible.

Well that explains a lot. I thought the code was part of the Jupiter package we "imported" from the ACE project.

core/test/junit/saros/concurrent/SplitOperationTest.java
```diff
@@ -60,7 +60,7 @@ public void testInsertInsert() {
     assertEquals(Collections.singletonList(expected1), split1.toTextEdit(path, source));
 
     // vice versa
```
tobous: I should replace this comment with an actual description of the setup.

srossbach: I guess adding the current "layout"

```
  position  0 1 2 3 4 5 6 7 8 9
  ------------------------------
  start:   "0 1 2 3 4 5 6 7 8 9"
  ins1:    "0 1 2 3 4 c d 5 6 7" (8 and 9 shifted out of view)
  ins2:    "0 1 2 a b 3 4 c d 5" (6 and 7 shifted out of view)
```

Either as comment above the test or before every operation transformation and assertion could be really helpful to understand what is tested here without digging through the whole code or papers regarding Jupiter.

tobous: Yes, I could do that. But doing this for every test setup gets very tedious, especially with the later ones using multiple split operations. Furthermore, I will have to extend this test suite as part of my upcoming refactoring of the text edit position handling. The new test cases will contain more complex setups (multi-line changes, etc.), which makes this documentation style not feasible.

I will adjust the class javadoc for the test class to better indicate what is being tested. With this information, the short description of initial activities and the resulting (expected) outcome should be descriptive enough in my opinion.

Would that be an acceptable solution for you?

srossbach: As you are going to rewrite the test cases anyway I do not think that the documentation has to be included / updated with this patch.

PR #876, opened 2020-03-19
[BUILD] Ignore javadoc errors during doc generation

m273d15:
In order to create a javadoc page, we have to ignore
the errors which are thrown because of our custom tags as
"@JTourBus".

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,6 +53,10 @@ configure(projectsToConfigure) { projectToConf ->
 
   test {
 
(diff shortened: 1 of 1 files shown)


m273d15:
> I guess it doesn't hurt but does the generated javadoc provide any value with the current issues?

The result is a valid javadoc API documentation which is sometimes useful. 

>Are the JTourBus entries included in the generated doc?

No.

PR #877, opened 2020-03-19
Introduce text edit content normalization

tobous:
#### [INTERNAL][CORE] Add LineSeparatorNormalizationUtils

Adds LineSeparatorNormalizationUtils offering utility methods to
normalize and denormalize the line separators for a given text.

This utility will be used in upcoming commits to normalize content
before passing it to the core, thereby allowing for a unified line
separator handling.

#### [INTERNAL] Normalize line separators for text edit content

Adjusts the logic handling text edit activities to normalize line
separators before creating text edit activities and to revert the
normalization when applying received text edit activities.

No adjustment was necessary for Saros/I as the editors already work on
normalized Unix line separators.

This adds a new corner case that can lead to session desynchronizations
between two Saros/E instances: When two participants with different line
endings settings add content to an empty shared file, the line separator
settings will be honored. As the watchdog checks the editor content
including the line endings, this will lead to a perceived
desynchronization.
This issue will be fixed once the content handling is completely
normalized and the watchdog works on normalized content as well.

#### [INTERNAL][CORE] Introduce assumption that text edits are normalized

Adjusts the internal logic handling text edit activities under the
assumption that all text passed to them is normalized, i.e. only uses
Unix line separators.

Adjusts the logic by replacing logic checking which line separator to
use with directly using the normalization line separator.

Removes now no longer necessary utility methods guessing the used line
separator.

Extends the TextEditActivity constructor javadoc to state that passed
content must be normalized.
Also adds assertions to the text edit activity constructor ensuring that
the passed content does not contain any Windows line separators.

PR #878, opened 2020-03-19
Adjust consistency watchdog to work with normalized content

tobous:
#### [INTERNAL] Add IEditorManager.getNormalizedContent(SPath)

Adds the method IEditorManager.getNormalizedContent(SPath). The method
provides access to normalized representations of the resource content,
i.e. the content only containing Unix line separators.

It is necessary for an upcoming adjustment of the consistency watchdog
logic to work with normalized resource contents.

#### [INTERNAL][CORE] Use normalized content for checksum creation

Adjusts the consistency watchdog to use normalized file contents for
checksum creation. This is necessary as IntelliJ normalizes content line
endings internally, meaning the watchdog always works on normalized line
endings in Saros/I. For IDE cross-compatibility, other IDE
implementations must also use normalized content for the checksum
creation in order for the checksums to match correctly.

This also resolves an issue introduced with #877 where adding a newline
to an empty file would lead to a desynchronization according to the
consistency watchdog if the participants had differing settings on which
line separator to use.

### DIFF
--- a/core/src/saros/editor/IEditorManager.java
+++ b/core/src/saros/editor/IEditorManager.java
@@ -6,6 +6,7 @@ import saros.editor.text.LineRange;
 import saros.editor.text.TextSelection;
 import saros.filesystem.IProject;
 import saros.session.User;
+import saros.util.LineSeparatorNormalizationUtil;
 
 /**
  * Tracks and provides access to editors for the set of files shared in the currently running
@@ -52,6 +53,19 @@ public interface IEditorManager {
    *     with the given path exists locally
    */
   String getContent(SPath path);
+
+  /**
+   * Returns the normalized text content of the local editor associated with the specified file, or
+   * the content of the file itself if there is currently no open editor for it.
+   *
+   * <p>Normalized text only contains Unix line separators.
+   *
+   * @param path path of the file whose content should be returned
+   * @return the normalized text content of the matching local editor or file or <code>null</code>
+   *     if no file with the given path exists locally
+   * @see LineSeparatorNormalizationUtil
+   */
+  String getNormalizedContent(SPath path);
 
   /**
    * Saves the local editors of all shared files belonging to the given project. If <code>null
--- a/core/src/saros/editor/IEditorManager.java
+++ b/core/src/saros/editor/IEditorManager.java
@@ -45,24 +45,26 @@ public interface IEditorManager {
   Set<SPath> getOpenEditors();
 
   /**
-   * Returns the text content of the local editor associated with the specified file, or the content
-   * of the file itself if there is currently no open editor for it.
+   * Returns the text content of the local document associated with the specified file.
+   *
+   * <p>The document content is the content that is displayed when the file is opened in a text
+   * editor.
    *
    * @param path path of the file whose content should be returned
-   * @return the text content of the matching local editor or file, or <code>null</code> if no file
-   *     with the given path exists locally
+   * @return the text content of the matching local document or <code>null</code> if the file does
+   *     not exist or no document could be obtained for the file
    */
   String getContent(SPath path);
 
   /**
-   * Returns the normalized text content of the local editor associated with the specified file, or
-   * the content of the file itself if there is currently no open editor for it.
+   * Returns the normalized result of {@link #getContent(SPath)}.
    *
    * <p>Normalized text only contains Unix line separators.
    *
    * @param path path of the file whose content should be returned
-   * @return the normalized text content of the matching local editor or file or <code>null</code>
-   *     if no file with the given path exists locally
+   * @return the normalized text content of the matching local document or <code>null</code> if the
+   *     file does not exist or no document could be obtained for the file
+   * @see #getContent(SPath)
    * @see LineSeparatorNormalizationUtil
    */
   String getNormalizedContent(SPath path);
--- a/core/src/saros/concurrent/watchdog/ConsistencyWatchdogClient.java
+++ b/core/src/saros/concurrent/watchdog/ConsistencyWatchdogClient.java
@@ -286,26 +286,26 @@ public class ConsistencyWatchdogClient extends AbstractActivityProducer implemen
       return false;
     }
 
-    final String editorContent = editorManager.getContent(path);
+    final String normalizedEditorContent = editorManager.getNormalizedContent(path);
 
-    if (editorContent == null) {
+    if (normalizedEditorContent == null) {
       log.debug("Inconsistency detected -> no editor content found for resource: " + path);
 
       return true;
     }
 
-    if ((editorContent.length() != checksum.getLength())
-        || (editorContent.hashCode() != checksum.getHash())) {
+    if ((normalizedEditorContent.length() != checksum.getLength())
+        || (normalizedEditorContent.hashCode() != checksum.getHash())) {
 
       log.debug(
           String.format(
               "Inconsistency detected -> %s L(%d %s %d) H(%x %s %x)",
               path.toString(),
-              editorContent.length(),
-              editorContent.length() == checksum.getLength() ? "==" : "!=",
+              normalizedEditorContent.length(),
+              normalizedEditorContent.length() == checksum.getLength() ? "==" : "!=",
               checksum.getLength(),
-              editorContent.hashCode(),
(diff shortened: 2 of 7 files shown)


tobous:
This will be the last patch dealing with content normalization for now. With it, we should be able to create a session between Saros/E and Saros/I without getting issues with the editor contents.

In my initial plan, I also wanted to move the entire internal content handling to a normalized system, including content for the initial project sharing, file creations, etc. The advantage of having complete content normalization is that the local choice of line separator is no longer forced onto other participants.

This extension of the content normalization is still very much a TODO on the way towards providing good IDE cross-compatibility, but it is not essential. With the current handling, you can already share resources between IDEs with the caviat

tobous:

srossbach:
Why do we have to introduce a new method to the editormanager interface ? Isn't is easier just to define a hash algorithm in the core (e.g discard all line endings and do the checksum calculation on the result) that is used by the watchdog client and server ?

tobous:
> Why do we have to introduce a new method to the editormanager interface ? Isn't is easier just to define a hash algorithm in the core (e.g discard all line endings and do the checksum calculation on the result) that is used by the watchdog client and server ?

I added the new method because it seemed like the easiest way to go and required the least amount of changes. Furthermore, it offers performance benefits if there is no normalization necessary (i.e. for Saros/I or Saros/E using Unix line endings).

And, if we adjust the `ConsistencyWatchdogHandler`, the old method could be completely dropped, meaning this would just replace an existing method.
(The `ConsistencyWatchdogHandler` uses the editor content to create the binary content for the recovery file creation activity. This seems weird to me as the editor content does not have to match the content on disk (e.g. for Saros/I), which is what is normally used for file creation activities. So for the current handling, the file content should probably be send instead. That would eliminate the last remaining usage of `IEditorManager.getContent(...)`.
When always normalizing content, we also would no longer be interested for the editor content for the recovery content but would rather also use the normalized editor content, meaning the method `IEditorManager.getContent(...)` would again no longer be used.)

srossbach:
There is no performance gain if you use your own hash function. Using a custom hash function is also faster as you do not have to call replace but you can just use your own hashing algorithm and converting \r and \r\n to \n on the fly.

You currently have to implement this function for every IDE although a custom hash method in the core is sufficient.

And yes the file creation have to be reverted to its original form where the content of the editor was saved and then the disk content was sent.

tobous:
> There is no performance gain if you use your own hash function. Using a custom hash function is also faster as you do not have to call replace but you can just use your own hashing algorithm and converting \r and \r\n to \n on the fly.

But that approach doesn't offer optimal performance in all cases though. Since the core is IDE independent, we can't make any assumptions about the line separator used in the file and doesn't have access to IDE methods used to detect the line separator. As a result, the overhead will increase for Unix line ending separated files as we still have to check every character to look for Windows line endings (even though this can't happen for Saros/I as an example). (As a lazy estimate, if we assume that Windows and Unix users and Saros/I and Saros/E users are uniformly distributed, we would be working with an overhead in ~75% of cases: All users using Saros/I and all users using Saros/E & Unix line endings.)

A combination of both approaches would be possible by also passing the used line separator to the core, thereby transporting the information from the IDE-specific context to the core context.

But we are talking about minute performance optimizations here. As I still have a *lot* to working on IDE cross-compatibility and this decision isn't critical for the functionality (and can easily be adjusted after the fact), I would like to move on from this issue. If you still think this is *that* important, I can open an issue for it (or you could implement the new hashing logic yourself).

@srossbach Would that be acceptable to you?

> You currently have to implement this function for every IDE although a custom hash method in the core is sufficient.
> 
> And yes the file creation have to be reverted to its original form where the content of the editor was saved and then the disk content was sent.

Yes, a better handling would be nice. But I don't think such a fix belongs into this PR. I will open an issue for it. It isn't that pressing and, if I get around to implementing the second part of the content normalization, I will have to adjust that part of the logic anyways.

srossbach:
Seems you do not understood me.

```
String content = editorManager.getContent(...)
int hash = 0;
for (int i = 0; i < content.length(); i++)
{
   char c = content.charAt(i);
   if (c == '\r')
   {
      if (i + 1 < length && content.charAt(i + 1) == '\n')
         i++;

     c = '\n'
   }
   hash += hash(c)
}

**Comment by tobous** [comment]
> Seems you do not understood me.
> 
> ```
> String content = editorManager.getContent(...)
> int hash = 0;
> for (int i = 0; i < content.length(); i++)
> {
>    char c = content.charAt(i);
>    if (c == '\r')
>    {
>       if (i + 1 < length && content.charAt(i + 1) == '\n')
>          i++;
> 
>      c = '\n'
>    }
>    hash += hash(c)
> }
> ```

I do understand, but I don't see the big performance boost this is supposed to have. Both approaches have linear runtime (both O(n) where n is the length of the text if the string replacement logic is not implement naively), so they don't differ by much.

(In general, I still don't think it is sensible to support the old Mac OS line separator format (just `\r`). It even further worsens the runtime by adding an additional comparison and should hardly be used anymore. I think a better approach would be to introduce a sanity check for such corner cases (`\r` line endings, mixed line ending, etc.) at session start.) 

Using this algorithm, we go through the entire text once and check for every char whether it is a line separator or not and then compute the hash (except for the `\r`, but that is a negligible difference), meaning we have to compute two operations on each char (three if we include Mac OS line endings).

Using the current logic, we have two cases.

For a text already using Unix line endings, the normalization function immediately returns the passed text, meaning it is a NOP. As a result, we also only have to go through the entire text once and compute only a single operation on each char (the hash function), meaning the runtime will be better as we don't have to do the comparisons.

For a text using Windows line endings, we have to go through the text twice. Once for the normalization and a second time here for the hash computations. As the normalization contains string replacements reducing the text length (thereby needing a downshift of the content after the replacement if it is implemented naively), this case is *way* slower than you approach.

But, as IntelliJ always uses the Unix line separator and there also are a lot of developers using Unix line endings, the first case will probably occur way more than half of the time.

In general, the best solution in my opinion would be, as already hinted at above, to implement a mix of both approaches. Use the algorithm you suggested but also let the method used to obtain the content return the used line separator. That way, we can use any IDE-specific assumptions (like IntelliJ always using Unix line separators) to speed up the Unix cases while still improving performance for the Windows cases.

But adding the mixed solution would require the addition of an even "worse" method to the `IEditorManager` interface as it would be purpose build for this one specific case. `Pair<String,String> getContentAndLineSeparator(SPath)`, a method that returns a Pair containing the content and the line separator for the content.

You could also look for line separators in the text and get the used line separator that way. But all of these are micro-optimizations in my opinion.

**Comment by tobous** [comment]
Rebased onto current base branch, resolving some minor merge issues caused by the logger refactoring.

**Comment by tobous** [comment]
Rebased onto the current base branch without any interaction.

**Comment by tobous** [comment]

File: core/src/saros/editor/IEditorManager.java [hunk 1]
```diff
@@ -53,6 +54,19 @@
    */
   String getContent(SPath path);
 
+  /**
+   * Returns the normalized text content of the local editor associated with the specified file, or
```
srossbach, line 58: Maybe I am reading it wrong, but when I have to implement this I would just return the non normalized content if the file is not open in any editor and so this would lead to inconsistencies.

Looking at the return description it states that the file content has to be normalized as well.

tobous: I just copied it from the other description. But yes, the description (for both) does not match the functionality. It shoud state that it returns the (normalized) editor content or the file content if the file is not openable in a text editor. That is the current logic if I remember correctly (will check again). Will adjust the javadoc for both.

tobous: I should have actually looked at the logic before hastily writing a reply from memory. The current logic for Eclipse and IntelliJ **only** returns the editor/document content (even if there is no editor for the resource open). If no such editor/document content can be obtained, `null` is returned. (And the server always returns the "editor" content as it works directly on a text buffer internally.)

Will correct the javadoc.

tobous: Done. Adjusted the javadoc in additional commit. The new wording is a bit clunky and I am not completely happy with it, but I couldn't come up with anything better. Its hard to properly convey the "document" concept (i.e. the content that would be displayed in a text editor if a text editor were open for the file).

PR #883, opened 2020-03-24
[JUNIT][CORE] Add check for remote version compatibility result

tobous:


stefaus:
Would be nice if #847 is merged before ;)
https://github.com/saros-project/saros/pull/847/files#diff-de61f266c78c41b0bbfe9c90727e5106

tobous:
I didn't see that you were still adjusting the test class in the follow up PR. My bad.

Issue #620, opened 2019-08-05
Saros/E URLs are outdated


srossbach:
We have some help buttons that currently will open a browser either displaying 404 or some other stuff that is not related to would should be shown.

In general we should go through all related code in fixing the URLs.

m273d15:
I think all user related urls are:
- [ ] **eclipse/src/saros/ui/wizards/pages/CreateXMPPAccountWizardPage.java**:94: "http://www.saros-project.org/setupXMPP"
- [ ] **eclipse/src/saros/feedback/StatisticManager.java**:38: "http://www.saros-project.org/Feedback"
- [ ] **eclipse/bin/main/saros/feedback/messages.properties**:20 "https://www.saros-project.org/Feedback"
- [ ] **eclipse/bin/main/saros/feedback/messages.properties**:53: "https://www.saros-project.org/Feedback#Pseudonym"
- [ ] **eclipse/bin/main/saros/ui/messages.properties**:117: "http://www.saros-project.org/Feedback"
- [ ] **eclipse/bin/main/saros/ui/messages.properties**:254: "http://saros-project.org/faq"
- [ ] **eclipse/bin/main/saros/ui/messages.properties**:261: "https://www.saros-project.org/documentation/troubleshooting.html"
- [ ] **eclipse/src/saros/feedback/messages.properties**:20: "https://www.saros-project.org/Feedback"
- [ ] **eclipse/src/saros/feedback/messages.properties**:53: "https://www.saros-project.org/Feedback#Pseudonym"
- [ ] **eclipse/src/saros/ui/messages.properties**:117: "http://www.saros-project.org/Feedback"
- [ ] **eclipse/src/saros/ui/messages.properties**:254: "http://saros-project.org/faq"
- [X] **eclipse/src/saros/ui/messages.properties**:261: "https://www.saros-project.org/documentation/troubleshooting.html"

Should we think about some mechanism to warn us if we change a corresponding url in the docs?

srossbach:
I think it would be better to provide links that will never change and will redirect you. At least this helps you to update them without letting the user to install a new plugin version with corrected URLs.

m273d15:
Should be possible with adjusting the `_config.yml` and adding [permalinks](https://jekyllrb.com/docs/permalinks/) pointing to the important pages.

PR #885, opened 2020-03-25
[INTERNAL][E] Update saros urls

m273d15:
Update urls for next release.


m273d15:
This does not fix #620 (because there is currently no equivalent to the feedback URLs and we thought about introducing [permalinks](https://jekyllrb.com/docs/permalinks/)) but is sufficient for the next Saros/E rel.

PR #887, opened 2020-03-30
[I] Use XMPPContactsService instead of Roster

stefaus:
To remove smack classes outside the core, this patch replaces the
roster logic with XMPPContactsService.

### DIFF
--- a/intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
+++ b/intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
@@ -3,111 +3,103 @@ package saros.intellij.ui.tree;
 import com.intellij.openapi.Disposable;
 import com.intellij.openapi.util.Disposer;
 import com.intellij.util.ui.UIUtil;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import javax.swing.tree.DefaultMutableTreeNode;
 import javax.swing.tree.DefaultTreeModel;
-import javax.swing.tree.MutableTreeNode;
-import org.jivesoftware.smack.Roster;
-import org.jivesoftware.smack.RosterEntry;
-import org.jivesoftware.smack.packet.Presence;
 import saros.SarosPluginContext;
 import saros.intellij.ui.util.IconManager;
 import saros.intellij.ui.views.SarosMainPanelView;
 import saros.net.xmpp.JID;
-import saros.net.xmpp.roster.IRosterListener;
-import saros.net.xmpp.roster.RosterTracker;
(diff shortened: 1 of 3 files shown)


saros-infrastructure:


```diff

Complexity decreasing per file
==============================
+ intellij/src/saros/intellij/ui/menu/SarosFileShareGroup.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5271199&bid=17265764)

intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
```diff
@@ -3,111 +3,108 @@
...
+        public void update(Optional<XMPPContact> contact, UpdateType updateType) {
+          UIUtil.invokeLaterIfNeeded(
+              () -> {
+                if (updateType == UpdateType.ADDED) {
+                  contact.ifPresent(ContactTreeRootNode.this::addContact);
+                } else if (updateType == UpdateType.CONNECTED) {
+                  contactsService
+                      .getAllContacts()
+                      .stream()
+                      .forEach(ContactTreeRootNode.this::addContact);
```
tobous: I don't really understand this case. From reading the logic, it seems to me like it is re-adding all contacts whenever an account connects to the XMPP server. This should lead to each account entry being present multiple times as I don't see a call that cleans up the list.

But this is not the case. Every entry is always only present once.

Could you explain this to me? Where is my mistake in reading the logic? What exactly does this logic do and which case does it cover?

saros-infrastructure:


```diff

Complexity decreasing per file
==============================
+ intellij/src/saros/intellij/ui/menu/SarosFileShareGroup.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5271199&bid=17265764)

intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
stefaus: Good that you checked this :D

First I reimplemented the `IRosterListener` (rosterChanged call) here. Split it to not_connected and connected which introduced this part and than I didn't check if I could delete it.

The old impl looks like it did the same... But I removed it.

stefaus: And to answer why this doesn't duplicate contacts:
https://github.com/saros-project/saros/blob/9cba0106eaf9eb023aa4dbc0b9f0979c793052d9/intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java#L87-L89
So instead of creating a new contact the code stops here.

PR #894, opened 2020-04-01
[BUILD] Always run STF on PR

tobous:
Sets the STF workflow to run for all PRs (excluding pure documentation
PRs) and on pushes on master. This now seems sensible as the STF is
relatively stable and accidental breakages still occur.

Also runs the STF for any push on a branch matching 'test/stf/*'. This
can be used to trigger STF builds for arbitrary commits without creating
PRs, allowing the GitHub workflows to be used as a remote test runner.

### DIFF
--- a/.github/workflows/stf.yml
+++ b/.github/workflows/stf.yml
@@ -3,8 +3,11 @@ name: STF Tests
 on:
   push:
     branches:
-      - 'pr/stf/*'
       - 'master'
+      - 'test/stf/*'
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
 
 jobs:
   stf:


stefaus:
I think that the STF runs are currently not stable enough. I started them three times for #897 without code changes till they passed.
Additionally they need at least 30 minutes and produce ~90 mb of Artifacts per run.

It would be nice to run them manually.

tobous:
> I think that the STF runs are currently not stable enough. I started them three times for #897 without code changes till they passed.

Agreed. With my runs before this PR, I did not have much trouble with flickering tests. But many of the recent runs also failed for me. Maybe we merged something that reduced the test reliability? 

Either way, closed for now until the framework is more stable. We should still keep an eye on broken STF builds on the master to ensure that way that we don't break existing tests.

PR #900, opened 2020-04-10
Fix text activity offset handling

tobous:
This one seems to have slipped by me as I only tested the selection stuff when migrating the selection logic to the new positioning scheme and did not expect it to break due to the migration of the text edit activity handling (which is exactly what happened).

But this gave me a nice opportunity to fix one of my earlier design blunder – using operations instead of the `TextEditActivity` object when applying received activities – that had been annoying me for a while now. The new logic is  much easier to understand in my opinion.

#### [INTERNAL][I] Add method to replace text in documents

Adds DocumentAPI.replaceText(...), which can be used to replace a
specific range of text in a given document.

This method will replace the current usage of insertion and deletion
calls in an upcoming commit.

#### [FIX][I] Fix handling of received text edit activities

Fixes an issue where the logic to adjust the existing contribution
annotations calculated the offset for the given text positions after the
text edit activity had already been applied, leading to wrong offsets,
which in turn lead to wrong contribution annotations.
With the new logic, only the start offset is calculated from the
position. The end offsets are calculated from the new and replaced text
lengths. This is possible in this case as the text has already been
denormalized (which isn't actually necessary for IntelliJ). As we don't
have to worry about differing line separators, we can directly use the
length of the strings instead having to transform the text positions.

Simplifies the logic by calculating the needed offset once and then
reusing them where necessary.

Re-works the handling to now work with the text edit activity instead of
transforming it into operations.

Moves the logic to apply the text edit into the EditorManager.

#### [INTERNAL][I] Use DocumentAPI.replaceText(...) for content recovery

Replaces the separate invocation of DocumentAPI.deleteText(...) and
.insertText(...) with a single call to DocumentAPI.replaceText(...).

Subsequently removes the now unused method for deletion/insertion.


intellij/src/saros/intellij/editor/EditorManager.java
```diff
@@ -123,109 +121,192 @@ private void execEditorActivity(EditorActivity editorActivity) {
...
-         * ContributionAnnotation</code> is added for the inserted text and all existing annotations
-         * for the file are adjusted through {@link
-         * AnnotationManager#moveAnnotationsAfterAddition(IFile, int, int)}.
+         * @param path the path to obtain an editor for
+         * @return returns a valid editor for the given resource or <code>null</code> if no such
+         *     editor could be obtained
+         * @see BackgroundEditorPool
+         */
+        private Editor getCalculationEditor(@NotNull SPath path) {
+
```
tobous, line 182: Remove unnecessary empty line if I rebase this.

intellij/src/saros/intellij/editor/EditorManager.java
```diff
@@ -123,109 +121,192 @@ private void execEditorActivity(EditorActivity editorActivity) {
...
+            if (!replacedText.equals(documentReplacedText)) {
+              log.error(
+                  "Text to be replaced for "
+                      + path
+                      + " from offset "
+                      + start
+                      + " to "
+                      + oldEnd
+                      + " does not match the given replaced text. Should be '"
+                      + StringEscapeUtils.escapeJava(replacedText)
```
m273d15, line 247: The main idea behind using `escapeJava` is to escape non-printable characters and avoiding linebreaks in the output? (We cannot assume that the users are writing java, but it should work in most cases)

intellij/src/saros/intellij/editor/EditorManager.java
```diff
@@ -123,109 +121,192 @@ private void execEditorActivity(EditorActivity editorActivity) {
...
-                    } else if (textOperation instanceof DeleteOperation && editor == null) {
+          if (newEnd != start) {
+            if (editor == null) {
+              annotationManager.moveAnnotationsAfterAddition(file, start, newEnd);
+            }
 
-                      annotationManager.moveAnnotationsAfterDeletion(file, start, end);
-                    }
-                  });
+            annotationManager.addContributionAnnotation(user, file, start, newEnd, editor);
```
m273d15, line 308: If `editor == null` is true, the `addContributionAnnotation` is also called. I would expect that an annotation is always assigned to an editor. The editor parameter is @Nullable, but what happens with an annotation that is created for `editor == null`?

intellij/src/saros/intellij/editor/EditorManager.java
tobous: Yes, I mainly used it to escape new lines in the logged content. To be honest, I did not think about that the user might not be using java. I just used the same method already used in other parts of the core for the same purpose (escaping log messages). But, as you already pointed out, it should not matter that much. The method/class is deprecated anyways, so we should replace all of its usages with a unified solution at some point.

intellij/src/saros/intellij/editor/EditorManager.java
tobous: This is explained in the javadoc of [AbstractEditorAnnotation](https://github.com/saros-project/saros/blob/master/intellij/src/saros/intellij/editor/annotations/AbstractEditorAnnotation.java#L25).

The thought behind allowing the editor to be `null` is that there isn't always an open editor for the resource that we receive changes for that would cause annotations (e.g. I don't have an editor open for a file another session participant is currently editing). (This also covers the case that the editor for such a resource is closed and re-opened during a session.)

Such annotations should still be tracked by Saros. To do so, they are added to the annotation store without an editor. As a result, these annotations are only stored as offsets and not shown anywhere (at this point). As soon as the user open an editor for the resource, the editor will be added to all stored annotations and they will be added to that editor (which is called "adding a local representation" in the annotation manager logic). This happens in [AnnotationUpdater](https://github.com/saros-project/saros/blob/master/intellij/src/saros/intellij/eventhandler/editor/editorstate/AnnotationUpdater.java#L21).

tobous:
Rebased onto current master (without any interaction (only removing the superfluous blank line).

PR #905, opened 2020-04-13
[INTERNAL][I] Use content for file move activities

tobous:
Checks whether the file move activity contained a content to use for the
new file. If so, uses the given content. Keeps the original content
otherwise.

This split handling is necessary as file moves are handled differently
by different IDEs. For Saros/E, the adjustments to the file as part of
the move are passed as the new content for the file. For Saros/I such
adjustments are send as additional text edit activities after the file
was moved.

With this adjustment, both handlings should be compatible with each
other.


tobous:

PR #906, opened 2020-04-13
Remove IResource.move(...) method

tobous:
#### [INTERNAL][S] Replaces the usage of IResource.move(...)

Replaces the usage of IResource.move(...) with explicitly creating the
new resource and then deleting the old resource. The new implementation
returns with a warning if the old resource to move does not exist. It
does not create the new resource in such cases.

This was done to remove the last (non-test) usage of IResource.move(...)
so that the method can be removed in a subsequent PR.

#### [API] Remove IResource.move(...)

Removes the method IResource.move(...) and its implementations.
Removes all remaining usages of IResource.move(...) (only test cases).

The method was removed as it was only used by the Saros Server and is
not generally usable with the Saros resource model.

As the Server only operates on folders and the resource move
implementation is set up to handle absolute paths, it is technically
usable for the Server implementation.

But for any other IDE implementation, the method will fail for cases
where the target resource does not share the same base reference point
(currently project/module). Furthermore, the general usage of absolute
paths in the IResource logic is discouraged as the Saros resource model
always works relative to a shared base point.

In general, since the logic is only used in the IDE implementations of
Saros and not in the core, it does not seem sensible to keep it as part
of the core interface. The necessary abstractions to do so make the
logic much harder to implement/understand. Furthermore, completely
handling such move cases in the IDE logic without using the IResource
wrapper allows for much more optimized/IDE specific implementations of

### DIFF
--- a/core/src/saros/filesystem/IResource.java
+++ b/core/src/saros/filesystem/IResource.java
@@ -64,9 +64,6 @@ public interface IResource {
   /** Equivalent to the Eclipse call <code>IResource#delete(updateFlags, null)</code> */
   public void delete(int updateFlags) throws IOException;
 
-  /** Equivalent to the Eclipse call <code>IResource#delete(destination, force, null)</code> */
-  public void move(IPath destination, boolean force) throws IOException;
-
   public IPath getLocation();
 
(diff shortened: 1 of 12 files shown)


tobous:

Issue #912, opened 2020-04-15
Transfer and use file charset in project negotiation


tobous:
Currently, the file encoding is completely up to the local IDE. This can lead to issues if the local default encoding does not match the remote encoding. Furthermore, it completely relies on the IDE/project settings being synchronized as part of the session, which is no longer always the case when sharing reference points instead of projects. (Furthermore, I don't think synchronizing theses settings would work for cross-IDE sessions.)

To avoid these issue, the encoding for each file should be send and set for the local resources as part of the project negotiation. (To completely solve the encoding issue, we also have to actually always transfer the encoding for file activities if content is send and use the encoding when creating resources during a session. I have already implemented these changes and will create a PR for them soon.)

While implementing this behavior, we should also keep the possible issues mentioned in #774 in mind. The implementation of this behavior could resolve #774.

srossbach:
Just as a reminder as you have already made changes to the encoding stuff. If you explicitly set an encoding in Eclipse for a file it is written to the meta file (and IRC it does not matter if this encoding is the same as the project/container encoding). So in the end you could end up with a very big file containing thousands of unneccessary entries.

tobous:
> Just as a reminder as you have already made changes to the encoding stuff. If you explicitly set an encoding in Eclipse for a file it is written to the meta file (and IRC it does not matter if this encoding is the same as the project/container encoding). So in the end you could end up with a very big file containing thousands of unneccessary entries.

When introducing the `IFile` method to set the encoding, I kept the old behavior for setting Eclipse encoding that only set a specific value if it does not match the project default. See [`EclipseFileImpl.updateFileEncoding()`](https://github.com/saros-project/saros/blob/master/eclipse/src/saros/filesystem/EclipseFileImpl.java#L58). As a result, only the necessary entries will be added to the encoding file.

srossbach:
I guess that this issues has to be fixed soon. Depending on what you share (in Eclipse) the file responsible for managing the encoding is either synchronized or not (along with other meta files like `.project`, `.classpath`) etc.

PR #915, opened 2020-04-16
Remove IContainer.getDefaultCharset()

tobous:
~~Might still need to be adjusted. Currently, merging this PR would break the server as the project encoding is used as the default for file encodings. This issue would be resolved by implementing #912. Another option would be to add a default handling to the server file implementation until #912 is resolved. (As a result of the change, this would also mean that the server would have to persist the encodings or would no longer be able to add a project to the session.)~~ Added temporary workaround until #912 is resolved.

~~This PR also removes the addition of the project default encoding from the file list. This might still be useful for compatibility checks during before executing the project negotiation on the remote side. So this PR might be adjusted to just move the method `getDefaultCharset()` to `IProject`.~~
Any charset actually used in the project should be contained in the compatibility list as the charset for each file on the list is added as part of the list creation.

#### [INTERNAL] Remove usage of IContainer.getDefaultCharset()

Removes the usage of IContainer.getDefaultCharset() to prepare for its
upcoming removal.

Introduces UTF-8 as the default charset for the ServerFileImpl that is
used if no charset has been specified. This is needed until #912 is
resolved as resources created as part of the project negotiation won't
set their charset with the current logic. This workaround does not
change the current behavior of the server as this default was previously
set in the server project implementation instead and was requested by
the file implementation if no charset was available.

#### [API] Remove IContainer.getDefaultCharset()

Removes the method IContainer.getDefaultCharset() as not all containers
provide the concept of a default charset.

The method was only used in the FileListFactory to add to the list of
encodings and in the Server file implementation if the file does not
exist.

Furthermore, such a method to access the default charset for a container
should be unnecessary. When creating a file locally, the charset
assigned to the newly created resource should be used instead. When
creating files for a received file activity, the charset contained in
the activity should be used.

### DIFF
--- a/core/src/saros/filesystem/IContainer.java
+++ b/core/src/saros/filesystem/IContainer.java
@@ -33,8 +33,6 @@ public interface IContainer extends IResource {
 
   public IResource[] members() throws IOException;
 
-  public String getDefaultCharset() throws IOException;
-
   /**
    * Returns a handle for the file with the given relative path to this resource.
    *
--- a/core/src/saros/negotiation/FileListFactory.java
+++ b/core/src/saros/negotiation/FileListFactory.java
@@ -77,8 +77,6 @@ public class FileListFactory {
 
     FileList list = new FileList();
 
-    list.addEncoding(project.getDefaultCharset());
-
     List<IFile> files = calculateMembers(list, project);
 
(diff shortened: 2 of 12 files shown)


tobous:

tobous:

PR #916, opened 2020-04-16
Remove IResource.adaptTo(Class)

tobous:
#### [INTERNAL] Remove usage of IResource.adaptTo(Class)

Removes all (non-test) usages of `IResource.adaptTo(Class)` as preparation
for the removal of the method. Replaces the usage with a simple cast.
Also adds a resource type check beforehand if needed.

#### [API] Remove IResource.adaptTo(Class)

Removes the method `IResource.adaptTo(Class)` and its implementations.

The method introduced an unnecessary additional step for "unpacking" an
`IResource` variable to access the contained object through one of its
implementing interfaces. Most of the time, the actual type of the object
was already known when calling this method, making the additional type
check unnecessary. (Furthermore, the result was never checked for null,
making the additional type check completely pointless.)

The Saros/E implementation of the interface was the only one providing
any additional, more complicated logic. This additional logic is not
necessary in my opinion as we are usually only interested in accessing
the same object through a different interface. Furthermore, the
`adaptTo()` method was never used in the Saros/E code.

Removes all tests for the method or its implementations.


tobous:

tobous:

PR #920, opened 2020-04-17
[DOC] Create CNAME file in gh-pages

m273d15:
Storing the CNAME file in the master branch
and deploying it into the gh-pages branch
led to the situation that the landing page was missing.
In order to avoid that we create the file only in the gh-pages.

### DIFF
--- a/.github/workflows/build_doc.yml
+++ b/.github/workflows/build_doc.yml
@@ -50,10 +50,12 @@ jobs:
       # therefore we want to avoid that GitHub tries to interpret
       # the gh-pages content as Jekyll project
       # https://github.blog/2009-12-29-bypassing-jekyll-on-github-pages/
-      - name: Create .nojekyll file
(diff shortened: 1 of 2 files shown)


tobous:
I know its unlikely, but in case we ever were to change the URL of our homepage, it would be nice to have the CNAME store somewhere more visible than hidden in a build config.

How about creating a separate config file for it? Or can't we simply always keep it in the docs directory instead of adding it on build? If it is not automatically copied to the _site directory, we could instead just copy it there as part of the build instead of creating it.

m273d15:
> How about creating a separate config file for it? Or can't we simply always keep it in the docs directory instead of adding it on build? If it is not automatically copied to the _site directory, we could instead just copy it there as part of the build instead of creating it.

I created a file CNAME_template that is simply renamed to CNAME in the build.

PR #931, opened 2020-04-20
[FIX][I] Save all documents on session start

tobous:
Adjusts the logic to save all files for a given IProject to simply save
all modified documents.

This adjustment is necessary as the previous implementation relied on
there being an open editor for modified documents. In IntelliJ, however,
documents are not automatically saved on closing the editor, meaning
there can be unsaved documents without an open editor. As a result, the
session negotiation could sometimes work on outdated file states,
leading to the wrong initial state on the receiving side.

Simply saving all documents is the lazy approach implementing this. It
would also have been possible to go through all unsaved documents and
check whether they belong to the given project. This is not necessary in
my opinion as there shouldn't be that many unsaved documents around,
especially since, with the default settings, IntelliJ saves each
document as soon as its editor looses focus.

### DIFF
--- a/intellij/src/saros/intellij/editor/EditorManager.java
+++ b/intellij/src/saros/intellij/editor/EditorManager.java
@@ -795,17 +795,6 @@ public class EditorManager extends AbstractActivityProducer implements IEditorMa
     editorListenerDispatch.remove(listener);
   }
 
-  /**
-   * Saves the document for the given file, thereby flushing its contents to disk.
-   *
-   * @param file the file for the document to save
-   * @see Document
-   * @see LocalEditorHandler#saveDocument(IFile)
-   */
-  private void saveDocument(IFile file) {
-    localEditorHandler.saveDocument(file);
-  }
-
   public void removeAllEditorsForFile(IFile file) {
     editorPool.removeEditor(file);
   }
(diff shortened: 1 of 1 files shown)


tobous:
As mentioned in the commit message, a more complex approach actually only saving the necessary files is also possible. This would be the logic for this approach (I would also have to introduce the method `DocumentAPI.getUnsavedDocuments()`):
```Java
  @Override
  public void saveEditors(final IProject project) {
    Project currentProject = ((IntelliJProjectImpl) project).getModule().getProject();

    for (Document document : DocumentAPI.getUnsavedDocuments()) {
      VirtualFile virtualFile = DocumentAPI.getVirtualFile(document);

      if (virtualFile == null) {
        continue;
      }

      IResource resource = VirtualFileConverter.convertToResource(currentProject, virtualFile);

      if (resource == null) {
        continue;
      }

      if (resource.getProject().equals(project)) {
        DocumentAPI.saveDocument(document);
      }
    }
  }
```

I don't think this approach is necessary, but I still wanted to mention it.

What are your thoughts on the matter?

intellij/src/saros/intellij/editor/EditorManager.java
```diff
@@ -1023,21 +1012,7 @@ private void executeInUIThreadSynchronous(Runnable runnable) {
 
   @Override
```
srossbach, line 1013: What is the contract in the interface ?

tobous: For [`EditorManager.saveEditors(IProject)`](https://github.com/saros-project/saros/blob/master/core/src/saros/editor/IEditorManager.java#L79):
Save open editors of given project or all editors if given project is `null`. But this is based on the notion that only open editors can have modified content, which is not the case for IntelliJ. As mentioned in the commit message, documents (i.e. the editor content) can remain unsaved (i.e. are not written to disk) even if the editor is closed. The cached state will again be used when a new editor is opened for the document.

For [`DocumentAPI.saveAllDocuments()`](https://github.com/saros-project/saros/blob/master/intellij/src/saros/intellij/editor/DocumentAPI.java#L89):
Calls [`FileDocumentManager.saveAllDocuments()`](https://github.com/JetBrains/intellij-community/blob/master/platform/core-api/src/com/intellij/openapi/fileEditor/FileDocumentManager.java#L70) on the EDT, de-normalizing and then flushing all editor/document content to disk.

tobous:
Rebased onto current master resolving minor merge issues.

PR #932, opened 2020-04-20
Bump minimal version support to 2019.2.3 and address deprecations

tobous:
#### [BUILD][I] Build against IntelliJ Idea 2019.3.4

Sets the build target to IntelliJ Idea 2019.3.4.

Smoketest did not reveal any compatibility issues.
List of incompatible changes does not suggest any changes concerning
logic used by Saros/I.

I did not set the build target to 2020.1 as the update seems to have
introduced some changes to the filesystem handlers, the effects of which
have to be investigated further.

In particular, VirtualFileListener.beforeContentsChange(...) is now
often triggered without being caused by a save or a refresh. Such cases
are currently ignored by Saros but could be important events that we are
interested in.

#### [I] Increase oldest supported IntelliJ version to 2019.2.3

Sets the new oldest supported IntelliJ IDEA version to 2019.2.3
(build 192.6817.14). This was done in order to not fall too far behind
the current release.

2019.2.3 was chosen as it is the earliest 2019.2.X release that is
still available on the IntelliJ IDEA download page.

Dropping older releases enables us to replace logic that has since been
deprecated, ensuring compatibility with future releases.

#### [FIX][I] Address deprecations resolvable with minimal version 2019.2

Addresses deprecation issues that can now be resolved after the
backwards compatibility was bumped up to 2019.2.

Replaces the usage of ListCellRendererWrapper with
SimpleListCellRenderer in ModuleTab.

Adds parent disposable objects for document and selection listener
registration calls.

Adjusts LocalDocumentModificationHandlerTest to work with new
registration method.

The same should be done for viewport listener
registration calls once the backwards compatibility has been bumped up
to at least 2019.3. Adds corresponding comment to the call.

Adds a missing NotNull annotation to the listener implementations in
LocalClosedEditorModificationHandler and
LocalDocumentModificationHandler.

Uses the module object contained in the content entry root model for
logging in ModuleConfigurationInitializer.

Adjusts the comment regarding VirtualFilePropertyEvent.isRename() to
defer the adjustment until the method is no longer marked as
experimental. Using experimental API methods should be avoided as their
removal/adjustment is not listed in the list of
changes/incompatibilities, meaning their usage can more easily lead to
breakage with IntelliJ version updated.

### DIFF
--- a/intellij/build.gradle
+++ b/intellij/build.gradle
@@ -3,7 +3,7 @@ apply plugin: 'saros.gradle.intellij.plugin'
 sarosIntellij {
   sandboxBaseDir = intellijSandboxDir as File
   localIntellijHome = intellijHome as File
-  intellijVersion = 'IC-2019.3'
+  intellijVersion = 'IC-2019.3.4'
 }
 
 intellij {
--- a/intellij/resources/META-INF/plugin.xml
+++ b/intellij/resources/META-INF/plugin.xml
@@ -30,7 +30,7 @@
     </vendor>
 
     <!-- please see https://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/build_number_ranges.html for description -->
-    <idea-version since-build="182.5107.16"/>
+    <idea-version since-build="192.6817.14"/>
 
     <!-- please see https://www.jetbrains.org/intellij/sdk/docs/basics/getting_started/plugin_compatibility.html on how to target different products -->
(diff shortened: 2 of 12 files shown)


tobous:
The main thing to discuss in this PR is whether we want to bump the minimal supported version from `2018.2` to `2019.2`.

As mentioned in the commit message, bumping the compatibility allows us to stay at least somewhat current by removing some of the older deprecated methods/classes.

The release `2019.2`, which was first released in August of 2019 and whose latest still available release `2019.2.3` was released in September of 2019, still offers a large-enough window of supported releases to (hopefully) cover most users.

tobous:

PR #939, opened 2020-04-24
[STF] Retry stf tests

m273d15:
Use the gradle retry plugin for running the stf tests.
The current configuration lead to the behavior that:
* The test run fails if >= 10 tests failed in the first attempt (no retries, because
we assume that there is a real problem if so many tests are failing) 
* Tests that failed are repeated up to 3 times

### DIFF
--- a/stf.test/build.gradle
+++ b/stf.test/build.gradle
@@ -1,3 +1,7 @@
+plugins {
+    id 'org.gradle.test-retry' version '1.1.5'
+}
+
 sarosEclipse {
   manifest = file('META-INF/MANIFEST.MF')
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -8,6 +8,10 @@ buildscript
   }
 }
 
+plugins {
+  id 'org.gradle.test-retry' version '1.1.5' apply false
```
tobous, line 12: Out of curiosity: Why is it necessary to declare this in the root build script? Can't you just import the plugin in the stf build script?

m273d15: It is not necessary as long as we use the plugin only in one subproject. (Define the dependency once in the root project and using it by other subprojects).
IIRC it was necessary for a previous gradle version, but not anymore.

PR #948, opened 2020-04-28
[STF] Update to container tag 0.6

tobous:

### DIFF
--- a/travis/script/stf/shared_vars.sh
+++ b/travis/script/stf/shared_vars.sh
@@ -1,8 +1,8 @@
 #!/bin/bash -e
 
-stf_master_image="saros/ci_build:0.5"
-stf_slave_image="saros/stf_test_slave:0.5"
-stf_xmpp_image="saros/stf_xmpp_server:0.5"
+stf_master_image="saros/ci_build:0.6"
+stf_slave_image="saros/stf_test_slave:0.6"
+stf_xmpp_image="saros/stf_xmpp_server:0.6"
 
 stf_master_name="stf_master"
 stf_network_name="stf_test_network"


tobous:
It fails running `buildSearchableOptions`, I am not quite sure why though. It builds locally using `2019.2.3`. The only reason I can think of is that the local build uses the JBR version of IntelliJ while the docker file uses the non-JBR version.

@m273d15 You had some problems with the task `buildSearchableOptions` in the past. Any idea? Also, is there a particular reason why the docker container uses the non-JBR version of IntelliJ (besides it being ~100mb smaller)?

tobous:
Ok, using the release including the JBR does not work either. Also, adjusting the logic to pull the IntelliJ installation through gradle to ensure that the build does exactly the same thing as the local build also doesn't work.

The issue might be that IntelliJ started building with Java 11 starting with `2019.X` and there is no Java 11 installation available in the container. Furthermore, for whatever reason, when downloading the installation through gradle, it tries to download the JBR 11 but fails as it can't reach the download sources.

tobous:
Ok, building with java 12 (I couldn't find an alpine image for java 11) does not seem to help.

I still have no idea what the problem is.

tobous:
I tried building it locally using release `2019.2.3 - no JBR`, `2019.2.3 - JBR 11`, `2019.2.3 - JBR 8`, `2019.2.4 - no JBR`, `2019.3.4 - no JBR`, and using the gradle configuration determine and download the version (`2019.3.4 - JBR`). None of these worked.

m273d15:
> @m273d15 You had some problems with the task buildSearchableOptions in the past. Any idea? Also, is there a particular reason why the docker container uses the non-JBR version of IntelliJ (besides it being ~100mb smaller)?

I assume the "~100mb smaller" is the reason. I prepare an image without IntelliJ installation and test the result.

PR #951, opened 2020-04-29
[STF] Use openjdk img instead of custom img

m273d15:
* Use tag 0.6

* Use sh in master scripts
  The openjdk alpine image does not contain
  a bash installation. Therefore we have to
  switch to sh

* Move build scripts in master dir
  The build scripts are only used by stf
  (master container). Therefore, the script
  is simplified and moved into the master dir

## Subsequent TODOs
* Remove build image in the "dockerfiles" repository


tobous:
Do you know what the issue with the previous build setup was?

m273d15:
> Do you know what the issue with the previous build setup was?

No. If you like we can analyze the setup.

PR #959, opened 2020-05-04
Fix minor annotation manager issues

tobous:
#### [FIX][I] Remove annotation from mapping when rotated out

Adjusts the call AnnotationQueue.removeIfFull() to also remove the
dequeued annotation from the actual annotation mapping.

This was not causing any issues with the annotation logic as these
annotations no longer contain valid range highlighters, meaning they
would be removed from the annotation store when the corresponding editor
is closed. But this ensures that we only store annotations as long as
necessary to avoid any other potential issues with working on disposed
annotations and avoid unnecessary memory consumption.

#### [INTERNAL][I] Ensure that highlighter is valid in annotation range CTOR


saros-infrastructure:


```diff

Complexity increasing per file
==============================
- intellij/src/saros/intellij/editor/annotations/AnnotationRange.java  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5465648&bid=17914527)

PR #975, opened 2020-05-14
Fix minor issues

tobous:
#### [DOC] Make restriction headings actual headings

Adjusts the "Restrictions" section in the guide for starting a session
in Saros/I to feature headings instead of bold text. This was done so
that they also support smart anchors, making it easier to link to them.

#### [DOC] Fix typos

### DIFF
--- a/docs/documentation/getting-started.md
+++ b/docs/documentation/getting-started.md
@@ -168,11 +168,11 @@ in the Saros instances of your invited contacts (with synchronizing files, copyi
 {% alert info %}
 ##### Current restrictions
 
-**Number of Participants**
+###### **Number of Participants**
 
 Currently, Saros/I is restricted to two-participant sessions, meaning you can only create session containing the host and a single client.
 
-**Share only a single module**
+###### **Share only a single module**
 
 You can currently only share a single module. A module has to adhere to the following restrictions to be shareable through Saros:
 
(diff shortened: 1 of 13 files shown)


docs/documentation/getting-started.md
```diff
@@ -168,11 +168,11 @@ in the Saros instances of your invited contacts (with synchronizing files, copyi
 {% alert info %}
 ##### Current restrictions
 
-**Number of Participants**
+###### **Number of Participants**
```
m273d15, line 171: Are the `**` still necessary?

tobous: `h6` headings are no longer bold, making it very hard to distinguish between the text and the heading.

m273d15: In order to format all `h6` headings bold you could add
```css
h6 {
  font-weight: bold;
}
```
to `_sass/style.scss`

tobous: This is a general style question. If we set it for this heading, we should also consider setting it for other headings as it would look weird in comparison otherwise. As we haven't used h6 anywhere else, I didn't think it was worth it to implement a general solution for it.

Here is a screenshot with all headings in non-bold and bold with normal text in between for comparison.

headings-comparison

tobous:

Issue #797, opened 2020-01-06
Replace IPath with the usage of java.nio.file.Path


tobous:
`IPath` seems like an unnecessary implementation for a feature that is already provided by the default java libraries. For Saros/E, this isn't an issue as it is mainly a wrapper for the Eclipse `IPath` implementation, but for other IDEs this requires a separate implementation. This could theoretically also be "solved" by implementing the `IPath` implementations using `Path` for the other IDEs, but using Path directly in all cases seems more sensible to me as it removes an unnecessary level of abstraction and reduces unnecessary code overhead.

Most methods defined in [`IPath`](https://github.com/saros-project/saros/blob/master/core/src/saros/filesystem/IPath.java) can be directly replaced by a corresponding method in `Path`, the only outliers requiring some (minor) work being `toPortableString()` and `toOSString()` (from what I can tell).

Since this is a central part of the implementation, it should also be tested along the way that the move to using `Path` does not introduce any unintended side effects or changes the behavior of the logic. The [STF](https://www.saros-project.org/contribute/saros-testing-framework.html) should be a good starting point for this.

srossbach:
I do not know the Java Path API internals but looking at the Server code (and I guess the IntelliJ one is just a copy) then the toPortableString implementation is already broken.

Assuming we are using absolute paths (that we do not use at all)

Sending from Windows ```C:\FOO\BAR``` will translate to ```C:/FOO/BAR``` on Unix
While it is a valid path name on Unix the correct result would be to throw an exception as Unix does not have device letters.

Sending from Unix ```C:/FOO\BAR" (2 segments, C: , FOO\BAR" will translate to ```C:\FOO\BAR``` on Windows
The first segment is a drive letter on the Windows side which is wrong
The second segment splits into two segments foo and bar.

Source of the Eclipse Path implementation which is about 1k LOCs mainly handling these issues.

https://github.com/eclipse/rt.equinox.bundles/blob/master/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/Path.java

tobous:
Yeah, but on the other hand, creating "portable" versions of absolute paths does not make sense for our use case anyways as it is *very* unlikely that they will be usable for other clients.

So the easiest solution in my opinion would be to require the path to be relative for the `getPortablePath()` method to function and return `null` or throw an exception otherwise.

And edge cases such as segmented paths should be cought by the CTOR/factory method. As we have full control over the usage, such restrictions should not be too big of an issue.

[...]

PR #977, opened 2020-05-14
Add javadoc to resource interfaces

tobous:
#### [DOC][CORE] Add javadoc for resource interfaces

Adds javadoc for the resource interfaces describing their desired
behavior.

Some implementations might not match the described behavior. This will
be corrected in the upcoming migration to using reference points as the
resource implementations will have to be reworked anyways.

#### [NOP][CORE] Do minor cleanup in IPath and IPathFactory

Removes unnecessary declarations from IPath and IPathFactory.

Adds a TODOs to remove IPath and IPathFactory, linking to issue #797.

#### [NOP][CORE] Remove README in filesystem

Removes the README in core filesystem as it is no longer up-to-date.
Furthermore, with the rework of the resource interfaces, it should no
longer be necessary.


tobous:
I was a bit uncertain on what to keep in this PR, but decided to move the newer commits to a separate PR. So this PR should be ready to be reviewed.

tobous:

PR #978, opened 2020-05-14
Add default resouce methods

tobous:
#### [API][CORE] Add default implementations for specific IProject methods

Overwrites specific methods of IContainer in IProject to specify the
differing behavior for projects.

Adds default implementations for specific methods in IProjects that have
a constant behavior.

#### [API][CORE] Add default implementations for IResource.getType()

Adds default implementations of getType() for IProject, IFile, and
IFolder returning the corresponding type.


tobous:
Adding explanations for the default values/behaviors would probably be nice.

tobous:

PR #979, opened 2020-05-15
[I] Adjust documentation for release 0.3.0

tobous:
#### [I] Adjust documentation for release 0.3.0

Updates all mentions of the current Saros/I release to 0.3.0.

Adds the release notes page for release 0.3.0.

Bumps the release version badge displayed on the README.md page to
0.3.0.

#### [DOC][I] Adjust doc to fit support for all JetBrains IDEs

Adjusts the documentation to reflect that 'Saros for IntelliJ' is now
compatible with all JetBrains IDEs. (We might have to think about
changing the name to make this easier.)

Tries to replace mentions of 'IntelliJ' with 'JetBrains' where sensible
or specify that other JetBrains IDEs are supported as well.

Replaces mentions of 'IntelliJ IDEA' with just 'the IDE' when talking
about interacting with the IDE the plugin is running on.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -25,9 +25,9 @@ A clear and concise description of what you expected to happen.
 Please provide the corresponding log files, but be sure to **remove or obfuscate sensitive data** you don't want to publish. Please also make sure only to include the log files concerning the session you encountered the bug in.
 
 You can find the log files in the following locations:
-* IntelliJ IDEA:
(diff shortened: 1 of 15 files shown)


m273d15:
AFAIK The plugin will be installable in all "[IntelliJ Platform IDEs](https://www.jetbrains.org/intellij/sdk/docs/intro/intellij_platform.html#ides-based-on-the-intellij-platform)", which are not necessarily JetBrains products (e.g. AndroidStudio)

tobous:
I might have overdone it a bit with linking to the page describing "IDEs based on the IntelliJ platform". This could become cumbersome to update if the link ever were to break. But I wanted to make sure to avoid confusion on the user's side.

PR #999, opened 2020-05-23
Migrate UI classes to reference points (ref-point migration 3/5)

tobous:
Migrates the main UI logic to use the new reference point based resource implementations. Largely reworks the logic to choose how to represent a shared reference point locally.

### Commits

<details><summary><b>[INTERNAL][I] Migrate ContactPopMenu to new resource implementation</b></summary>
<br>

The contact pop menu still displays the list of shareable modules as
they still represent sensible units of resources to share. In a future
update, the option to open a real sharing menu to better select the
resources to share should be added.

Still keeps the lazy workaround only supporting modules with one content
root as it can't be guaranteed that modules with multiple content roots
can be represented as a unified tree.

</details>

<details><summary><b>[INTERNAL][I] Add method to check whether file is in project content</b></summary>
<br>

Adds ProjectAPI.isInProjectContent(Project,VirtualFile). The method can
be used to check whether a given virtual file is part of the content of
the given project, i.e whether it is located below the content root of
one of the modules of the project and is not excluded or ignored.

The method will be used to ensure that only resources belonging to the
selected project are used to represent a local reference point. This
handling is useful to avoid potential issues caused allowing unrelated
resources to be used as a reference point, e.g. them not being covered
by the filesystem listener or the current VFS snapshot.

This handling also adds a new restriction that only non-excluded
resources can be used as a reference point. As excluded resources are
currently ignored anyways, this matches the current behavior anyways.
When moving to a more complex exclusion behavior no longer dependent on
the IDE exclusion system, this might have to be adjusted.

</details>

<details><summary><b>[INTERNAL][I] Migrate AddProjectToSessionWizard</b></summary>
<br>

Migrates AddProjectToSessionWizard to use the new resource
implementation.

Adjusts the logs and messages shown to the user related to the reference
point handling.

Breaks the logic for using existing reference points. This was necessary
as the module selection page has not been updated yet, meaning we can't
obtain virtual file objects from it yet. This will be fixed once the
selection page has been migrated.

The rest of the cleanup of the class will be done in a separate commit.

Deletes ModuleConfiguration, ModuleConfigurationProvider, and
ModuleConfigurationInitializer as they were no longer being used.

Adds a todo to potentially remove AdditionalProjectDataFactory as the
module configuration logic was the only logic using it. This has to be
discussed further.

</details>

<details><summary><b>[INTERNAL][I] Migrate ModuleTab to use new resource implementation</b></summary>
<br>

Updates ModuleTab, ModuleSelectionResult, and LocalRepresentationOption
to use the new resource implementation.

Explicitly does not adjust the javadoc to keep the patch as small as
possible. The existing javadoc, class names, method names, variable
names, etc. will be adjusted in a separate commit.

Adjusts ModuleTab to return VirtualFile objects for the selected base
directory or the selected existing directory to share. The selected
virtual files must be part of the project content and must not be
ignored.

Replaces the combobox used to choose an existing module with a directory
chooser allowing the user to choose and existing directory instead.

Subsequently adjusts the text fields that are used to specify the paths
for the virtual files to only be adjustable through the directory
chooser dialog by setting the fields to not be editable. This was done
as the logic to obtain a virtual file for a path is potentially long
running (as it requires a VFS snapshot lookup), so running it on every
modification of the text field would cause too large of an overhead. By
using the directory chooser, the text field is only updated once the
wanted path is chosen. Furthermore, the directory chooser dialog ensures
that only existing resources that are directories are chosen,
simplifying the checking logic on our side.
To not reduce the usability, a mouse click listener was added to the
text fields which opens the directory chooser when the text fields is
clicked. The directory chooser still allows the user to enter a path of
their choosing instead of navigating the files system tree, thereby
replacing all functionality previously offered by the text field.

Removes some of the logic to pre-select specific option in certain cases
(i.e.  if a module with the same name is found locally) as they are no
longer valid. This logic will be re-added in a followup commit.

Adds the check whether there is a project selected to the validity
requirements. This was previously always done implicitly by other
fields.

Adjusts the messages shown to the user related to the adjusted
functionality to match the new functionality. The rest of the messages
will be updated separately.

Adjusts AddProjectToSessionWizard to use the new return values of
ModuleSelectionResult.

</details>

<details><summary><b>[INTERNAL][I] Rework pre-selection logic of ModuleTab</b></summary>
<br>

Updates the pre-selection logic of ModuleTab to check the name of the
project base directory and all content roots of all contained modules to
find a directory matching the name of the shared reference point.

This heuristic to guess a valid directory to represent the shared
reference point was chosen based on the assumption that user will most
often share the complete workspace or a complete module.

If the user instead chose to, for example, share a random inner
directory of a module, there is no good way of making an educated guess
of which directory could be used to represent it in the local project.
Names might exist multiple times and going through all contained
directories produces too high of an overhead to be feasible.

An alternative would be to pass the workspace-relative path as part of
the project negotiation. This would enable the logic to check whether a
similar resource exists in the local workspace. This would offer a
better heuristic for suggesting local directories to use but also won't
always work, especially in cross-IDE sessions.
But this approach does not fit into the scope of this change.

</details>


saros-infrastructure:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- intellij/src/saros/intellij/ui/wizards/pages/moduleselection/ModuleTab.java  2
         

Complexity decreasing per file
==============================
+ intellij/src/saros/intellij/context/SarosIntellijSessionContextFactory.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5574772&bid=18276451)

core/src/saros/negotiation/AdditionalProjectDataFactory.java
```diff
@@ -22,6 +22,7 @@
  * providers are disjoint. If multiple providers use the same key, the mapping of the provider added
  * last will be used.
  */
+// TODO remove as no longer used?
```
tobous, line 25: Extend explanation and/or link to PR.

intellij/src/saros/intellij/ui/messages.properties
```diff
@@ -183,9 +183,11 @@ ModuleTab_module_base_path_file_chooser_title=Choose Module Base Path
 ModuleTab_module_base_path_file_chooser_description=Choose the base bath used for the new module.
 ModuleTab_use_existing_module=Use existing module
 ModuleTab_use_existing_module_local_module=Local module
-ModuleTab_create_new_module_name_invalid_tooltip=The module name must not contain separator characters and must not match any existing module name for the chosen project.
-ModuleTab_create_new_module_base_path_invalid_tooltip=The given module base path does not exist or does not point to a directory.
-ModuleTab_use_existing_module_local_module_invalid_tooltip=An existing module must be chosen.
+ModuleTab_create_new_module_name_invalid_tooltip=The directory name must not be empty and must not contain separator characters.
+ModuleTab_create_new_module_base_path_invalid_tooltip=An existing directory that is part of the project content must be chosen.
+ModuleTab_existing_directory_path_file_chooser_title=Choose Existing Directory
+ModuleTab_existing_directory_path_file_chooser_description=Choose an existing directory to represent the shared reference point.
```
m273d15, line 189: Do we want to communicate the concept "Reference Point" to users?
I think this is a general question. I would expect that the user has the perception "This is the directory I want to share" and I think we should talk with the users in their language.

intellij/src/saros/intellij/ui/wizards/AddProjectToSessionWizard.java
```diff
@@ -144,202 +135,199 @@ public void next() {
...
-
-          ISarosSession session = sessionManager.getSession();
-
-          if (session == null) {
-            log.error("Encountered project negotiation without running session");
-
-            NotificationPanel.showError(
-                Messages.AddProjectToSessionWizard_no_session_message,
-                Messages.AddProjectToSessionWizard_no_session_title);
+        private void doExistingDirectory(
```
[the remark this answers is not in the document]
tobous: Yes, I introduced these methods. They are based on the two chases the user has: use an existing directory or create a new directory as part of the negotiation. There wasn't any other thought behind the naming of the methods. They were just introduced to encapsulate the logic representing the different choices.

intellij/src/saros/intellij/ui/messages.properties
tobous: But we still need a way to communicate to the user that certain directories are sharing roots and others aren't. For example, deleting a shared inner directory is fine with Saros. Deleting a reference point/sharing root directory, however, can not be handled by Saros and will lead to the session breaking down.

We could change the wording to "root directory" or "sharing root directory", but a similar naming scheme/concept is already used in a by IntelliJ IDEA with its content/source/resource roots, so we would have to be careful not to cause confusion.

m273d15: True, I think "root directory" is more intuitive.

tobous: Resolved in #1008.

tobous:
Rebased onto current base branch resolving minor merge conflicts.

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- intellij/src/saros/intellij/ui/wizards/pages/moduleselection/ModuleTab.java  2
         

Complexity decreasing per file
==============================
+ intellij/src/saros/intellij/context/SarosIntellijSessionContextFactory.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5574772&bid=18276451)

core/src/saros/negotiation/AdditionalProjectDataFactory.java
tobous: I decided to remove this logic in a separate PR as it is a change to the core. Furthermore, changing it here would cause merge conflicts down the line when rebasing the core cleanup (#1012).

tobous:

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- intellij/src/saros/intellij/ui/wizards/pages/moduleselection/ModuleTab.java  2
         

Complexity decreasing per file
==============================
+ intellij/src/saros/intellij/context/SarosIntellijSessionContextFactory.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5574772&bid=18276451)

PR #1002, opened 2020-05-25
Update eclipse & core version

m273d15:
Cherry-picked the version updates of the rc-16.0.0 branch.

### DIFF
--- a/eclipse/update/site.xml
+++ b/eclipse/update/site.xml
@@ -3,10 +3,8 @@
    <description url="http://dpp.sf.net/update">
       Site to install the Saros Feature for Distributed Party Programming
    </description>
-   <feature url="http://dpp.sourceforge.net/update/saros.feature_14.10.31.jar" id="saros.feature" version="14.10.31">
+   <feature url="features/saros.feature_16.0.0.jar" id="saros.feature" version="16.0.0">
       <category name="DPP"/>
    </feature>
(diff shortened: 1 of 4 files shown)


eclipse/update/site.xml
```diff
@@ -3,10 +3,8 @@
    <description url="http://dpp.sf.net/update">
       Site to install the Saros Feature for Distributed Party Programming
    </description>
-   <feature url="http://dpp.sourceforge.net/update/saros.feature_14.10.31.jar" id="saros.feature" version="14.10.31">
+   <feature url="features/saros.feature_16.0.0.jar" id="saros.feature" version="16.0.0">
```
tobous, line 6: Why does the relative link work in this case? Is the feature jar contained in the update site but not bundled with the other jars?

m273d15: The `url` attribute was only used by the XSL transformation that was used to generate the HTML-representation of the update-site (which used the relative path to the feature jar). However, we don't use this XSLT anymore. We create our HTML-representation via Jekyll (I introduced it with v15.0.0) and I removed the XSLT in the last [PR](https://github.com/saros-project/update-site-artifacts/pull/2).

Eclipse marks the attribute `url` as required, but the update-site build works without the argument. I could substitute the relative path with empty content `""`  and a comment that describes why we keep the empty entry.

Furthermore, we could remove all other files in the `update` subproject except `site.xml`
which are used to generate the HTML-representation (If you like I can added these changes to the PR).

tobous: I'm indifferent to whether this is done as part of this PR or in a separate PR. Do as you like.

PR #1005, opened 2020-05-25
[API][CORE] Return list for IContainer.members()

tobous:
Adjusts IContainer.members() to return a list instead of an array. This
simplifies the implementation and usage of the method which previously
required multiple conversions into an array and back into a list.

### DIFF
--- a/core/src/saros/negotiation/FileListFactory.java
+++ b/core/src/saros/negotiation/FileListFactory.java
@@ -1,7 +1,6 @@
 package saros.negotiation;
 
 import java.io.IOException;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.LinkedList;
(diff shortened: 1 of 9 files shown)


core/src/saros/negotiation/FileListFactory.java
```diff
@@ -100,7 +99,7 @@ public static FileList createFileList(
   private static List<IFile> calculateMembers(final FileList list, final IProject project)
       throws IOException {
 
-    List<IResource> resources = Arrays.asList(project.members());
+    List<IResource> resources = project.members();
 
     if (resources.size() == 0) return Collections.emptyList();
```
m273d15, line 104: Could be `resources.isEmpty()`

tobous: Done.

tobous:
Rebased onto current master. As the rest of the changes that this PR was based on still need longer to be reviewed, I decided to change the PR order and merge this first.

PR #1009, opened 2020-05-25
[BUILD] Add workflow running PMD on changes to the ruleset

tobous:
Adds the GitHub workflow build_pmd which is run on any push or pull
request matching the branch name 'pr/pmd/**' or modifying the pmd
ruleset file.

This action ensures that the PMD ruleset applies without any issues
before merging onto master. This is necessary as Codacy, which runs our
PMD checks, only uses the ruleset file of the master branch.

### DIFF
--- a/.github/workflows/build_pmd.yml
+++ b/.github/workflows/build_pmd.yml
@@ -1,0 +1,32 @@
+# Has to be the same name as in build in order to require the same check in GitHub
+name: Build
+
+on:
+  push:
+    paths:
+      - 'ruleset.xml'
+  pull_request:
+    paths:
+      - 'ruleset.xml'
+
+jobs:
+
+  run-pmd:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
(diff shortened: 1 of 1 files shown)


tobous:

.github/workflows/build_pmd.yml
```diff
@@ -0,0 +1,36 @@
+# Has to be the same name as in build in order to require the same check in GitHub
+name: Build
+
+on:
+  push:
+    branches:
```
m273d15, line 6: I never combined two types of filters (branches and paths). Do you know that they are evaluated as conjunction or disjunktion? I thought they use a conjunction: The change has to match the branch and the path.

tobous: I could not find any specific documentation for this case, but my guess would be that you are right. Otherwise, the PMD job should have run for this PR as it is on a branch matching `pr/pmd/**`. I will remove the branch name patterns as looking for changes to the pmd ruleset file should be good enough.

tobous: Works now for changes to `ruleset.xml`. See [test run](https://github.com/saros-project/saros/runs/709805667).

PR #1018, opened 2020-05-28
[INTERNAL][E] Move "Create Account" button in page

m273d15:
Move the "Create Account..." button logic into the corresponding wizard page.
The previous solution integrated the button in the WizardDialog and only
showed the button if the `EnterXMPPAccountPage` is selected.

I think the code is more consistent, but I am not really happy with the listener's boilerplate code.

-------
New Layout:
image


eclipse/src/saros/ui/widgets/wizard/EnterXMPPAccountComposite.java
```diff
@@ -90,6 +97,20 @@ public void expansionStateChanged(ExpansionEvent e) {
           }
         });
 
+    createServerExpandableCompositeContent();
+
+    if (CreateXMPPAccountWizard.CREATE_DIALOG_ENABLED) {
```
srossbach, line 102: I do not understand the whole checks. I mean you cannot click on a button that is either disabled or not visible. So there should be only one check in the whole class or none at all.

m273d15: Ahhh sure. Thank you :+1:
I just need
```java
createAccountButton.setVisible(CreateXMPPAccountWizard.CREATE_DIALOG_ENABLED)
```

PR #1019, opened 2020-05-28
[INTERNAL][CORE] Remove AdditionalResourceData

tobous:
Removes the field 'additionalResourceData' from ResourceNegotiationData
as it is no longer used.

Removes the logic used to inject this additional data.

This change breaks compatibility as it adjusts the xStream fields of
ResourceNegotiationData.


saros-infrastructure:


```diff

Complexity decreasing per file
==============================
+ core/src/saros/negotiation/ResourceNegotiationData.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5601506&bid=18350397)

srossbach:
I would not remove this if you plan to not transmit the .project file of an Eclipse project.

tobous:
> I would not remove this if you plan to not transmit the .project file of an Eclipse project.

That was not part of the planned logic. As we want to support IDE cross-compatibility, we don't necessarily always have access to the `.project` files for the shared content. (Even if the projects are set up for Eclipse. See Saros for example. The `.project` files are generated locally and are not under versioning.)

Furthermore, the shared scope does not always have to include a `.project` file, even if we are sharing Eclipse projects. So I decided to drop the IDE-specific handling on the sender side. Instead, the receiving user decides where to map the shared resources.

This works great if the user already has the project setup. But If they don't have any projects in the local workspace, this has some drawbacks. The idea was to create a default (language independent) project that we then create the resources in. This severely reduces the functionality of the IDE in such session, but I don't see another way of implementing a universal handling for incoming resource negotiations with an empty workspace. Furthermore, if `.project` files were shared, the user still has the option to correctly import the project after the fact. (This might require a session restart. I am not sure how Eclipse handles such updates of the resource mode.)

The creation of a default project is technically only necessary if we don't receive any `.project` files as we can still create the shared resources using the `IWorkspaceRoot` as the base resource handle. But, as Eclipse only displays resources covered by an imported project (as far as I know), we still have to create default projects for all non-covered reference points.

In general, this aspect of the new resource handling for Eclipse still needs some work. I will have to do some more in-depth testing on how exactly Eclipse handles new resources (with and without a project) and what is possible when importing existing or creating new projects.

But I am still convinced that relying on transferring the `.project` file is not a viable solution in all cases, so I would like to keep the sending logic as IDE-agnostic as possible and instead let the receiving side handle how to represent the shared resources.

srossbach:
I just mentioned it because I had play around with the factory to include project natures and other stuff which are normally part of the .project file.

I guess which a more flexible sharing there are multiple use cases. But one of them is to simple share a project that should be equal as most as possible on the receiving side with to much additional setup. Adding natures to a project in Eclipse is very cumbersome especially if you do not know which natures are needed.

tobous:
> I guess with a more flexible sharing there are multiple use cases. But one of them is to simple share a project that should be equal as most as possible on the receiving side with to much additional setup. Adding natures to a project in Eclipse is very cumbersome especially if you do not know which natures are needed.

My assumption was that I could check whether the created resource contained a `.project` file and just tell Eclipse to load the project. Or is there any additional setup necessary?

srossbach:
The Eclipse Wizard creates a new project (no natures anything) and then when the IPN will write the .project file (which have to be included in this process) Eclipse will "update" the project.

tobous:

saros-infrastructure:


```diff

Complexity decreasing per file
==============================
+ core/src/saros/negotiation/ResourceNegotiationData.java  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5601506&bid=18350397)

srossbach:
Can this be discarded regarding #1051 ?

tobous:
> Can this be discarded regarding #1051 ?

Yes.

PR #1031, opened 2020-06-22
[BUILD] Include grade sources when downloading gradlew

tobous:
Adjusts the gradle wrapper configuration to also download the gradle
sources. Downloading the sources as well allows IDEs to automatically
link the documentation for gradle methods used in the build.gradle
files.


m273d15:
Sorry, I should update gradle via:
`gradle wrapper --gradle-version <version> --distribution-type all` (I used the call without the `--distribution-type` argument.)

PR #1032, opened 2020-06-22
Migrate internal classes to use new resource implementation (ref-point migration 1/10)

tobous:
Migrates the internal logic of Saros/E to use the new resource model introduced in #1023. The UI and filesystem handler logic is migrated separately in follow-up PRs. Outdated javadoc and method names that are not directly related to the changed logic will be updated in a separate clean-up PR at the end.

Most of the changes made in this PR are relatively simple as they just replace the usage of the old resource implementation with the new one and usage of the `ResourceAdapterFactory` with `ResourceConverter`.

### Reviewing this PR

### Commits

<b>[INTERNAL][E] Migrate FolderActivityConsumer to new resource impl</b>

<details><summary><b>[INTERNAL][E] Partially migrate CollaborationUtils to new resource impls</b></summary>
<br>

Migrates the internal logic of CollaborationUtils to use the new
resource implementations. The logic dealing with the UI and old sharing
assumptions (e.g. that only projects are shared) will be migrated in a
followup commit.

</details>

<b>[REFACTOR][E] Replace manual unwrapping with calls to ResourceConverter</b>

<b>[INTERNAL][E] Migrate FileActivityConsumer to new resource impls</b>

<b>[INTERNAL][E] Migrate EditorPool to new resource impls</b>

<details><summary><b>[FIX][E] Correctly tear down Editor pool mapping</b></summary>
<br>

Adjusts EditorPool.removeAllEditors() to correctly tear down the inner
data structures, ensuring that no mapping is held after the call.

</details>

<b>[INTERNAL][E] Migrate SharedDocumentProvider to new resource impls</b>

<b>[INTERNAL][E] Use ResourceConverter in EclipsePathFactory</b>

<b>[INTERNAL][E] Migrate EclipseWorkspaceImpl to new resource impls</b>

<details><summary><b>[INTERNAL][E] Migrate FileContentNotifierBridge to new resource impls</b></summary>
<br>

This migration uncovered further issues in the separation between the
plugin and session context. The checksum cache is instantiated inside
the plugin context as it is used by the resource negotiations, whose
factories are also part of the plugin context.

With the new logic, the notifier only works if there is a running
session that has a valid reference point mapping as we otherwise can't
compute Saros file objects for the modified Eclipse files. This is due
to Saros file objects being relative to their reference point.

In a future patch, the separation between the plugin in session context
should be improved by moving the resource negotiation factories to a
separate class located in the session context. This would also allow the
EditorManager to be moved into the session context.

As the checksum cache implementation also can't be used before a valid
reference point mapping is established, it we could also consider
splitting the resource negotiation into a resource mapping negotiation
and then a resource content negotiation to better differentiate between
the two phases and avoid illegal usage of the checksum cache.

To avoid cyclic dependency issues with the Saros session manager, it has
to be injected later on.

</details>

<b>[INTERNAL][E] Migrate EclipseCollaborationUtilsImpl</b>

<details><summary><b>[INTERNAL][STF] Migrate Eclipse PackageExplorerView</b></summary>
<br>

Adds a nullcheck for the obtained Eclipse resource object as the new
converter no longer accepts null values.

</details>

<details><summary><b>[INTERNAL][E] Remove EditorAPI.getEditorFile(IEditorPart)</b></summary>
<br>

Removes EditorAPI.getEditorFile(IEditorPart). The method only offers
additional logic to convert from Eclipse resource objects to Saros
resource objects. This is not the responsibility of the editor API.

Adds convenience method convertToFile(IFile) to EditorManager. The
method can be used to convert Eclipse file objects into Saros file
objects using the current set of shared reference points and
ResourceConverter.convertToFile(Set, IFile).

Replaces the usage of EditorAPI.getEditorFile(IEditorPart) with calling
EditorAPI.getEditorResource(IEditorPart) and subsequently converting it
using convertToFile(IFile).

</details>

<b>[INTERNAL][E] Migrate EditorManager to new resource impls</b>

### DIFF
--- a/eclipse/src/saros/editor/EditorManager.java
+++ b/eclipse/src/saros/editor/EditorManager.java
@@ -54,9 +54,9 @@ import saros.editor.remote.UserEditorStateManager;
 import saros.editor.text.LineRange;
 import saros.editor.text.TextPosition;
 import saros.editor.text.TextSelection;
-import saros.filesystem.EclipseFileImpl;
 import saros.filesystem.IReferencePoint;
 import saros.filesystem.ResourceAdapterFactory;
+import saros.filesystem.ResourceConverter;
 import saros.observables.FileReplacementInProgressObservable;
 import saros.repackaged.picocontainer.annotations.Inject;
 import saros.session.AbstractActivityConsumer;
@@ -441,7 +441,7 @@ public class EditorManager implements IEditorManager {
   }
 
   private String doGetContent(saros.filesystem.IFile wrappedFile) {
-    IFile file = ((EclipseFileImpl) wrappedFile).getDelegate();
+    IFile file = ResourceConverter.getDelegate(wrappedFile);
     FileEditorInput input = new FileEditorInput(file);
 
     IDocumentProvider provider = EditorAPI.connect(input);
@@ -467,7 +467,7 @@ public class EditorManager implements IEditorManager {
       return null;
     }
 
-    IFile file = ((EclipseFileImpl) wrappedFile).getDelegate();
+    IFile file = ResourceConverter.getDelegate(wrappedFile);
 
     String lineSeparator = FileUtil.getLineSeparator(file);
 
@@ -703,7 +703,7 @@ public class EditorManager implements IEditorManager {
     log.trace(".execTextEdit invoked");
 
     saros.filesystem.IFile fileWrapper = textEdit.getResource();
-    IFile file = ((EclipseFileImpl) fileWrapper).getDelegate();
+    IFile file = ResourceConverter.getDelegate(fileWrapper);
 
     if (!file.exists()) {
       log.error("TextEditActivity refers to file which" + " is not available locally: " + textEdit);
@@ -1169,7 +1169,7 @@ public class EditorManager implements IEditorManager {
    */
   private boolean isDirty(saros.filesystem.IFile wrappedFile) throws FileNotFoundException {
 
-    IFile file = ((EclipseFileImpl) wrappedFile).getDelegate();
+    IFile file = ResourceConverter.getDelegate(wrappedFile);
 
     if (file == null || !file.exists()) {
       throw new FileNotFoundException("File not found: " + wrappedFile);
@@ -1197,7 +1197,7 @@ public class EditorManager implements IEditorManager {
   public void saveEditor(saros.filesystem.IFile wrappedFile) {
     checkThreadAccess();
 
-    IFile file = ((EclipseFileImpl) wrappedFile).getDelegate();
+    IFile file = ResourceConverter.getDelegate(wrappedFile);
 
     log.trace(".saveEditor (" + file.getName() + ") invoked");
 
--- a/eclipse/src/saros/editor/EditorManager.java
+++ b/eclipse/src/saros/editor/EditorManager.java
@@ -55,7 +55,6 @@ import saros.editor.text.LineRange;
 import saros.editor.text.TextPosition;
 import saros.editor.text.TextSelection;
 import saros.filesystem.IReferencePoint;
-import saros.filesystem.ResourceAdapterFactory;
 import saros.filesystem.ResourceConverter;
 import saros.observables.FileReplacementInProgressObservable;
 import saros.repackaged.picocontainer.annotations.Inject;
@@ -633,7 +632,7 @@ public class EditorManager implements IEditorManager {
     String normalizedReplacedText =
         LineSeparatorNormalizationUtil.normalize(replacedText, lineSeparator);
 
-    saros.filesystem.IFile wrappedFile = ResourceAdapterFactory.create(file);
+    saros.filesystem.IFile wrappedFile = editorPool.getFile(changedEditor);
 
     TextEditActivity textEdit =
         TextEditActivity.buildTextEditActivity(
@@ -1066,11 +1065,13 @@ public class EditorManager implements IEditorManager {
 
     if (EditorAPI.getViewer(editorPart) == null) return false;
 
-    final IResource resource = EditorAPI.getEditorResource(editorPart);
+    final IFile file = EditorAPI.getEditorResource(editorPart).getAdapter(IFile.class);
 
-    if (resource == null) return false;
+    if (file == null) return false;
 
-    return session.isShared(ResourceAdapterFactory.create(resource));
+    final saros.filesystem.IFile fileWrapper = convertToFile(file);
+
+    return fileWrapper != null && session.isShared(fileWrapper);
   }
 
   /**
@@ -1268,12 +1269,19 @@ public class EditorManager implements IEditorManager {
   /**
    * Sends an Activity for clients to save the editor of given file.
    *
+   * <p>Does nothing if the file does not represent a shared resource.
+   *
    * @param file the saved file
    */
   void sendEditorActivitySaved(IFile file) {
+    saros.filesystem.IFile fileWrapper = convertToFile(file);
+
+    if (fileWrapper == null || !session.isShared(fileWrapper)) {
+      return;
+    }
+
     activityDelayer.fireActivity(
-        new EditorActivity(
-            session.getLocalUser(), Type.SAVED, ResourceAdapterFactory.create(file)));
+        new EditorActivity(session.getLocalUser(), Type.SAVED, fileWrapper));
   }
 
   /**
@@ -1455,7 +1463,7 @@ public class EditorManager implements IEditorManager {
 
       if (resource == null) continue;
 
-      if (ResourceAdapterFactory.create(resource).equals(file)) return true;
+      if (ResourceConverter.getDelegate(file).equals(resource)) return true;
     }
 
     return false;
--- a/eclipse/src/saros/editor/EditorManager.java
+++ b/eclipse/src/saros/editor/EditorManager.java
@@ -375,7 +375,7 @@ public class EditorManager implements IEditorManager {
 
     this.preferenceStore = preferenceStore;
 
-    editorPool = new EditorPool(this);
+    editorPool = new EditorPool(this, sessionManager);
     partListener = new SafePartListener2(log, new EditorPartListener(this));
 
     registerCustomAnnotations();
@@ -997,7 +997,7 @@ public class EditorManager implements IEditorManager {
   }
 
   private void partClosedOfFile(IEditorPart editorPart, saros.filesystem.IFile file) {
-    editorPool.remove(editorPart);
+    editorPool.remove(editorPart, file);
     openEditorFiles.remove(file);
 
     ITextViewer viewer = EditorAPI.getViewer(editorPart);
--- a/eclipse/src/saros/editor/EditorManager.java
+++ b/eclipse/src/saros/editor/EditorManager.java
@@ -526,7 +526,9 @@ public class EditorManager implements IEditorManager {
       return;
     }
 
-    saros.filesystem.IFile wrappedFile = EditorAPI.getEditorFile(part);
+    IFile file = EditorAPI.getEditorResource(part).getAdapter(IFile.class);
+    saros.filesystem.IFile wrappedFile = convertToFile(file);
+
     if (wrappedFile == null) {
       log.warn("Could not find file for editor " + part.getTitle());
       return;
@@ -553,17 +555,18 @@ public class EditorManager implements IEditorManager {
    *     selected text in editor.
    */
   void generateSelection(IEditorPart part, TextSelection newSelection) {
+    IFile file = EditorAPI.getEditorResource(part).getAdapter(IFile.class);
+    saros.filesystem.IFile wrappedFile = convertToFile(file);
 
-    saros.filesystem.IFile file = EditorAPI.getEditorFile(part);
-    if (file == null) {
+    if (wrappedFile == null) {
       log.warn("Could not find file for editor " + part.getTitle());
       return;
     }
 
-    if (file.equals(locallyActiveEditor)) localSelection = newSelection;
+    if (wrappedFile.equals(locallyActiveEditor)) localSelection = newSelection;
 
     activityDelayer.fireActivity(
-        new TextSelectionActivity(session.getLocalUser(), newSelection, file));
+        new TextSelectionActivity(session.getLocalUser(), newSelection, wrappedFile));
   }
 
   /**
@@ -913,7 +916,8 @@ public class EditorManager implements IEditorManager {
       return;
     }
 
-    saros.filesystem.IFile editorFile = EditorAPI.getEditorFile(editorPart);
+    IFile file = EditorAPI.getEditorResource(editorPart).getAdapter(IFile.class);
+    saros.filesystem.IFile editorFile = convertToFile(file);
 
     if (Objects.equals(editorFile, locallyActiveEditor)) {
       log.debug(
@@ -991,9 +995,12 @@ public class EditorManager implements IEditorManager {
       return;
     }
 
-    saros.filesystem.IFile file = EditorAPI.getEditorFile(editorPart);
+    IFile file = EditorAPI.getEditorResource(editorPart).getAdapter(IFile.class);
+    saros.filesystem.IFile wrappedFile = convertToFile(file);
 
-    partClosedOfFile(editorPart, file);
+    if (wrappedFile != null) {
+      partClosedOfFile(editorPart, wrappedFile);
+    }
   }
 
   private void partClosedOfFile(IEditorPart editorPart, saros.filesystem.IFile file) {
@@ -1326,7 +1333,9 @@ public class EditorManager implements IEditorManager {
     IEditorPart activeEditor = EditorAPI.getActiveEditor();
 
     if (activeEditor != null) {
-      locallyActiveEditor = EditorAPI.getEditorFile(activeEditor);
+      IFile file = EditorAPI.getEditorResource(activeEditor).getAdapter(IFile.class);
+      locallyActiveEditor = convertToFile(file);
+
       partActivated(activeEditor);
     }
   }
@@ -1345,8 +1354,10 @@ public class EditorManager implements IEditorManager {
    */
   private void refreshAnnotations(IEditorPart editorPart) {
 
-    saros.filesystem.IFile file = EditorAPI.getEditorFile(editorPart);
-    if (file == null) {
+    IFile file = EditorAPI.getEditorResource(editorPart).getAdapter(IFile.class);
+    saros.filesystem.IFile wrappedFile = convertToFile(file);
+
+    if (wrappedFile == null) {
       log.warn("Could not find file for editor " + editorPart.getTitle());
       return;
     }
@@ -1377,7 +1388,7 @@ public class EditorManager implements IEditorManager {
         continue;
       }
 
-      EditorState remoteEditor = userEditorStateManager.getState(user).getEditorState(file);
+      EditorState remoteEditor = userEditorStateManager.getState(user).getEditorState(wrappedFile);
 
       if (remoteEditor == null) continue;
 
@@ -1788,5 +1799,26 @@ public class EditorManager implements IEditorManager {
    */
   public IActivityProducer getActivityProducer() {
     return activityDelayer;
+  }
+
+  /**
+   * Converts the given file delegate to a Saros file object.
+   *
+   * @param fileDelegate the file delegate to convert
+   * @return a Saros file object representing the given file delegate or <code>null</code> if there
+   *     is not running session or the delegate does not belong to a shared reference point
+   * @throws NullPointerException if the given file delegate is <code>null</code>
+   * @see ResourceConverter#convertToFile(Set, IFile)
+   */
+  private saros.filesystem.IFile convertToFile(IFile fileDelegate) {
+    ISarosSession currentSession = session;
+
+    if (currentSession == null) {
+      return null;
+    }
+
+    Set<IReferencePoint> referencePoints = session.getReferencePoints();
+
+    return ResourceConverter.convertToFile(referencePoints, fileDelegate);
   }
 }
--- a/eclipse/src/saros/editor/EditorManager.java
+++ b/eclipse/src/saros/editor/EditorManager.java
@@ -374,7 +374,7 @@ public class EditorManager implements IEditorManager {
 
     this.preferenceStore = preferenceStore;
 
-    editorPool = new EditorPool(this, sessionManager);
+    editorPool = new EditorPool(this);
     partListener = new SafePartListener2(log, new EditorPartListener(this));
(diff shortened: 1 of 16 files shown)


saros-infrastructure:


```diff

Complexity increasing per file
==============================
- eclipse/src/saros/util/EclipseCollaborationUtilsImpl.java  1
- eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java  7
- eclipse/src/saros/editor/EditorPool.java  2
- stf/src/saros/stf/server/rmi/superbot/component/view/eclipse/impl/PackageExplorerView.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5736297&bid=18764265)

eclipse/src/saros/editor/EditorPool.java
```diff
@@ -237,7 +249,30 @@ public void remove(final IEditorPart editorPart) {
 
     editorManager.disconnect(file);
 
-    final saros.filesystem.IFile wrappedFile = ResourceAdapterFactory.create(file);
+    final saros.filesystem.IFile wrappedFile = convert(file);
```
srossbach, line 252: I do not understand why specific Core logic must be implemented here. I also discovered that the hashset uses core IFile, The whole logic here is Eclipse related, i.e it operates on editor parts and Eclipse IFiles.

It not the duty of this class to check any sharing etc.

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
```diff
@@ -51,8 +74,28 @@ public void resourceChanged(IResourceChangeEvent event) {
 
         final IFile file = delta.getResource().getAdapter(IFile.class);
 
-        for (IFileContentChangedListener listener : fileContentChangedListeners)
-          listener.fileContentChanged(ResourceAdapterFactory.create(file));
+        final ISarosSession sarosSession = sarosSessionManager.getSession();
```
srossbach, line 77: So you are basically breaking the cache implementation ... congratulation !

Does the interface states it operates only on SHARED files ? I do not think so.

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
```diff
@@ -29,10 +40,22 @@ public FileContentNotifierBridge() {
...
+  private void initialize() {
+    SarosPluginContext.initComponent(this);
+  }
+
   @Override
   public void resourceChanged(IResourceChangeEvent event) {
     if (event.getDelta() == null) return;
 
+    if (sarosSessionManager == null) {
+      initialize();
```
srossbach, line 52: Nice hack ... circular dependency problem ?

srossbach (dismissed):

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
tobous: Yeah. Its not pretty, but it is already used in other places that had similar issues. (I know, not a good excuse.)

The problem is that the caching logic is required by the negotiation factory, which in turn is required by the session manager, which is required here for the caching logic.

This could be resolved by moving the caching logic into the session context and splitting the negotiation factory into a session negotiation factory (plugin context) and a resource negotiation factory (plugin context).

But, again, I don't think this should be done as part of this migration.

eclipse/src/saros/editor/EditorPool.java
tobous: I introduced the usage of Saros `IFile` as part of the `SPath` removal, but it was using `SPath` objects beforehand, so that didn't really change a whole lot.

I don't see a problem in using Saros objects here. We are only interested in the editors of shared files anyways.

Furthermore, converting between Eclipse and Saros resource implementations is not as easy as it used to be. With the reference point implementations, we have to check every shared reference point by comparing its base path against the resource path. This creates a huge overhead compared to just wrapping the Eclipse object. So I would actually prefer directly using Saros objects here and unwrapping them if necessary instead of doing it the other way around.

This is also the reason why this class is "checking for sharing". With the new implementation, we can only ever create meaningful Saros resource objects for shared resources. Otherwise, we have no way of determining which directory to use as the base reference point to use.

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
srossbach: Start a session. Stop it. Modify some files that were part of the former sharing. Start a session again. You now have dirty checksums.. And if the dirty checksums match the remote checksums (which is likely) you will produce inconsistencies.

eclipse/src/saros/editor/EditorPool.java
srossbach: The check if this editor is shared is already done in the EditorManager. Why do you not just change the signature of the add method to require an IFile object beside the IEditorPart ?

tobous: Will do.

tobous:

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- eclipse/src/saros/editor/EditorPool.java  2
- eclipse/src/saros/util/EclipseCollaborationUtilsImpl.java  1
- eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java  7
- stf/src/saros/stf/server/rmi/superbot/component/view/eclipse/impl/PackageExplorerView.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5736297&bid=18764265)

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
```diff
@@ -51,8 +74,28 @@ public void resourceChanged(IResourceChangeEvent event) {
...
+
+        if (fileWrapper == null) {
+          if (log.isTraceEnabled()) {
+            log.trace("Ignoring resource change for non-shared file " + file);
+          }
+          return;
+        }
+
+        for (IFileContentChangedListener listener : fileContentChangedListeners) {
+          listener.fileContentChanged(fileWrapper);
```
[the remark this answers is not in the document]
srossbach: In case some listener throws an exception every listener afterwards would not be notified.

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
srossbach: I guess moving it into session scope can be pretty annoying if a session crashes for what reason. You should not assume that everyone is using SSDs or 16 Gigabytes of system memory (OS cache).

tobous: Both solutions are possible and offer their own advantages.

Separating the negotiation factories removes the dirty circular dependency workaround (and removes another one currently necessary for the [`NegotiationFactory`](https://github.com/saros-project/saros/blob/master/core/src/saros/negotiation/NegotiationFactory.java#L58)). But, as you mentioned, it only provides the caching benefits during the running session (i.e. for additional resource negotiations with new participants).

Making the caching logic generic and still using Eclipse `IFile` for the Saros/E cache removes the reference point lookup from the notification bridge logic and still offers the caching benefit for starting new sessions.
But, the cache being session-independent puts us in the weird position that Saros could potentially still use up memory and computing power even though there is currently no Saros session running. Even if the user is not planning on starting another session, the current caching model from the last session is still kept in memory and updated constantly for every change in the workspace.

I implemented a rough draft for both approaches:
[Move checksum cache to session context](https://github.com/saros-project/saros/commit/3fbc9e55447baad4b71a546cdf417525161c1baf)
[Make filesystem checksum cache generic](https://github.com/saros-project/saros/commit/0162fb579738a07b853391b5cfadbe657c265dc3)

I personally would still tend towards making the caching logic session-bound. (Also, I don't plan on fixing this as part of this PR but rather as a separate PR.)

@srossbach @m273d15 @stefaus What are your opinions?

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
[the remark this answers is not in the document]
tobous: Ok, and what "exception" should I catch in particular? A specific exceptions type? All checked exceptions? Runtime exceptions? Errors? Everything?

eclipse/src/saros/editor/EditorPool.java
tobous: Done. Also applied to `EditorPool.add(...)` and removed the conversion logic from `EditorPool`.

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- eclipse/src/saros/util/EclipseCollaborationUtilsImpl.java  1
- stf/src/saros/stf/server/rmi/superbot/component/view/eclipse/impl/PackageExplorerView.java  1
- eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5736297&bid=18764265)

tobous:
Adjusted `EditorPool` according to review comments.

eclipse/src/saros/editor/EditorManager.java
```diff
@@ -1326,7 +1343,9 @@ private void updateEditorPool() {
     IEditorPart activeEditor = EditorAPI.getActiveEditor();
 
     if (activeEditor != null) {
-      locallyActiveEditor = EditorAPI.getEditorFile(activeEditor);
+      IFile file = EditorAPI.getEditorResource(activeEditor).getAdapter(IFile.class);
+      locallyActiveEditor = convertToFile(file);
```
srossbach, line 1347: As convertToFile can return null I guess the logic may act slightly different as before (have not looked at the null checks, to be more specific isShared(locallyActiveEditor) lines).

tobous: `SarosSession.isShared(IResource)` delegates the call to [`SharedReferencePointMapper.isShared(IResource)`](https://github.com/saros-project/saros/blob/pr/eclipse/ref-point-migration/1/inner-classes/core/src/saros/session/internal/SharedReferencePointMapper.java#L140), which checks whether the input is `null` and just returns `false` in such cases.

I am not really sure what you mean by "slightly different behavior". This still sets the `locallyActiveEditor` to the currently active shared editor (if present). The only difference is that it no longer also holds the reference to non-shared active editors. But such editors should not be of interest to Saros. Furthermore, in the only corner case that I can think of where this could matter – the currently active, non-shared editor belonging to a resource that is added to the session – the variable is updated as part of the [`resourcesAdded(...)` listener in `EditorManager`](https://github.com/saros-project/saros/blob/pr/eclipse/ref-point-migration/1/inner-classes/eclipse/src/saros/editor/EditorManager.java#L326), calling [`updateEditorPool()`](https://github.com/saros-project/saros/blob/pr/eclipse/ref-point-migration/1/inner-classes/eclipse/src/saros/editor/EditorManager.java#L1327)

srossbach: The state of a non shared editor is also broadcasted. But I guess as we do not distinguish between no opened editor and a non shared editor I guess it will not matter.

Looking at the code and if I am not mistaken it was possible to send an SPath pointing to a non shared editor which should be normally null.

tobous: > Looking at the code and if I am not mistaken it was possible to send an SPath pointing to a non shared editor which should be normally null.

Even if it was possible, it is no longer possible with the new `ResourceTransportWrapper` logic that replaced the `SPath`. The new logic now always requires Saros `IResource` objects which can only be created for shared resources.

It is, however, still possible to send `null` as the resource for the selected editor to tell the other clients that no shared editor is currently active.

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
m273d15: I think moving the cache in the session context reduces the benefit and leads to the question "Do we need this cache?".

@srossbach Do you know the history of this cache (who requested the optimization and why)?

If we want to/should keep the cache I think we should use the second alternative (make the cache generic; even if I would like to remove the circular dependency). Otherwise, the benefit seems to be too small.

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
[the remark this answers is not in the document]
tobous: I don't think this is something that should be resolved as part of this PR. As far as I can tell, this issue applies to most (if not all) of our listener implementations.

In general, the question is how we want to handle unexpected exceptions. Is it worth it still informing other listeners, i.e. do we think that the plugin is still in an operable state where informing the other listeners servers any purpose?

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
tobous: I would also question the usefulness of the cache in general. If the user were to be working on huge projects, their hardware would have to match anyways as they otherwise could not load the project into the IDE or work on it in any sensible/timely manner.
But I am still a bit torn on this as I do see the utility during a session. Saros already hogs a lot of resources during a session, so actually using up more resources by creating checksums for all shared resources might push it over the edge (i.e. make the system noticeably slow down). On the other hand, the added cache still produces additional overhead during the session by keeping the checksums up-to-date, even if no further resource negotiation is planned, so removing it might benefit the general Saros performance (but not by a whole lot).

This would be a whole different discussion if the cache could also be used by the watchdog during the session as it would then serve a far broader purpose in this case. But this is currently not possible/not an option as the negotiation cache works on the file content on disc and the watchdog works on the editor content. Working on editor content is important for the watchdog to always work on the most current information, so using the content on disc is not an option. And using the editor content for the negotiation cache is not an option either (for now) as the negotiation could contain binary files which can't be opened in an editor. These caches could, however, be unified if we were to introduce the unified content handling (i.e. adding a binary check and always using the editor content for non-binary files, even for the negotiations).

If we want to keep the cache, I still disagree that moving it to the session context removes most of its benefits. I think it is much more likely that more participants are invited during the session than that the user is starting a new session on the same resources. (Yes, the session crashing/timing out is an exception to this, but I don't think this can be categorized as a default/common case.)

Also, as already mentioned, I am unhappy with the cache still running and using up system resources even though there is no Saros session running. In general, this is a very unexpected behavior from the users perspective and, if this were to negatively impact the system performance, could lead to lots of complaints/uninstalls.
Even if this is an extreme case that might not be that common (or even very uncommon), I would still question the benefits of the cache compared to its drawbacks, especially outside of the session. The overhead gets worse with increases in project size and decreases in hardware resources/power, so the users that would most benefit from this cache would also be impacted most negatively by its overhead.

srossbach:
I agree. If you move the cache into the session context you can discard the whole logic.

I added this cache to increase the speedup when working over time, e.g you invite someone later at a given point or just stop the session and continue somewhere in the future.

I do not know how bad the speed lost would be as there are many factors (beside from CPU load which will be high if you remove the cache) like are you using an SSD or and HDD (notebook hdd are very bad) and how much memory the OS will use for file caching.  From my experiences developer PC of a company are always to old and always lack memory as the IDE isn't the only program that is currently running, so expect a memory usage of about 80-90 unless you are blessed with 16 GByte and more. And last but not least do not forget the famous Antivirus-Realtime protection.

I guess you have to play around with realistic project sizes and "normal developer PCs" to decide if the cache is still useful or not.

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
stefaus: As mentioned in the standups I think it makes sense to move the cache into the session context. It is a tradeoff and not suitable for all cases, but as @tobous mentioned I don't think its good that Saros still updates checksums outside of sessions, burning resources without use. But keeping it inside the session context is good for multiple invites especially for Saros Server.

srossbach: The cache do not update checksums. It just invalidates existing stored checksums for files that were changed.

tobous: > The cache do not update checksums. It just invalidates existing stored checksums for files that were changed.

You are right. I misread the code. But it still creates overhead with calculating the path hashes, but it isn't as bad as I thought.

I still think it is at its most useful during the session as I would assume it is more likely to start a session with multiple participants or add participants to a running session (where the cache would still work for all non-modified shared files) than starting another session on the same resources in the same application lifecycle.

Furthermore, starting a new session requires a break in the workflow anyways, so additional time to re-populate the cache should not be that much of an issue in my opinion.

tobous: Resolved in #1072.

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
tobous: Resolved in #1072.

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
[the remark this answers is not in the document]
m273d15: @tobous This is still unresolved. Otherwise, I would approve the PR. What is your opinion?

tobous:

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- eclipse/src/saros/util/EclipseCollaborationUtilsImpl.java  1
- eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java  7
- stf/src/saros/stf/server/rmi/superbot/component/view/eclipse/impl/PackageExplorerView.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5736297&bid=18764265)

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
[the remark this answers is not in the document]
tobous: Done.

tobous:
Added try catch block for `FileContentNotificationBridge` listener notification.

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java  8
- eclipse/src/saros/util/EclipseCollaborationUtilsImpl.java  1
- stf/src/saros/stf/server/rmi/superbot/component/view/eclipse/impl/PackageExplorerView.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5736297&bid=18764265)

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
```diff
@@ -51,8 +74,32 @@ public void resourceChanged(IResourceChangeEvent event) {
...
+          if (log.isTraceEnabled()) {
+            log.trace("Ignoring resource change for non-shared file " + file);
+          }
+          return;
+        }
+
+        for (IFileContentChangedListener listener : fileContentChangedListeners) {
+          try {
+            listener.fileContentChanged(fileWrapper);
+          } catch (Exception e) {
```
srossbach, line 99: Just RuntimeException

eclipse/src/saros/editor/EditorManager.java
```diff
@@ -1789,4 +1810,25 @@ private void checkThreadAccess() {
...
+  /**
+   * Converts the given file delegate to a Saros file object.
+   *
+   * @param fileDelegate the file delegate to convert
+   * @return a Saros file object representing the given file delegate or <code>null</code> if there
+   *     is not running session or the delegate does not belong to a shared reference point
+   * @throws NullPointerException if the given file delegate is <code>null</code>
+   * @see ResourceConverter#convertToFile(Set, IFile)
+   */
+  private saros.filesystem.IFile convertToFile(IFile fileDelegate) {
```
srossbach, line 1823: I have seen the logic 3 times in the whole patch set. Any reason not to move the logic to the ResourceConverter and adding (a) method(s) that accepts a session object and an IResource (IFile, IFolder) object ?

eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java
tobous: Done.

eclipse/src/saros/editor/EditorManager.java
tobous: All this method does is check whether the given session object is `null`, request the currently shared reference points and call `ResourceConverter.convertToFile(...)` with it.

In most other cases, the check whether the session object is valid is already done beforehand by different logic. In general, I don't think it is necessary to introduce a separate method for this in the `ResourceConverter`.

Regardless, this is a very minor change that can also be done after the fact, so I don't think it should hold back this PR.

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- stf/src/saros/stf/server/rmi/superbot/component/view/eclipse/impl/PackageExplorerView.java  1
- eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java  8
- eclipse/src/saros/util/EclipseCollaborationUtilsImpl.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5736297&bid=18764265)

tobous:

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- eclipse/src/saros/util/EclipseCollaborationUtilsImpl.java  1
- stf/src/saros/stf/server/rmi/superbot/component/view/eclipse/impl/PackageExplorerView.java  1
- eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java  8
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5736297&bid=18764265)

m273d15 (dismissed):

stefaus (approved):

tobous:

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- stf/src/saros/stf/server/rmi/superbot/component/view/eclipse/impl/PackageExplorerView.java  1
- eclipse/src/saros/filesystem/checksum/FileContentNotifierBridge.java  8
- eclipse/src/saros/util/EclipseCollaborationUtilsImpl.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5736297&bid=18764265)

PR #1036, opened 2020-06-22
Migrating ui internals to new resource implementation 2/3 (ref-point migration 5/10)

tobous:
Migrates parts of the internal UI logic of Saros/E to use the new resource model introduced in #1023. Outdated javadoc and method names that are not directly related to the changed logic will be updated in a separate clean-up PR at the end.

This is the second part of the internal UI migration.

### Reviewing this PR

### Commits

<details><summary><b>[INTERNAL][E] Migrate CollaborationUtils.startSession(...)</b></summary>
<br>

Migrates CollaborationUtils.startSession(...) to work on a set of
reference points. Adjusts the usages of the method to correctly supply a
set of reference points.

</details>

<details><summary><b>[INTERNAL][E] Migrate CollaborationUtils.addResoucesToSession(...)</b></summary>
<br>

Renames the method to addReferencePointsToSession(...) to make it more
clear what the method actually does.

Adjusts the method to get passed the set of reference points to add.

Adjusts the usage of the method to match the new implementation.

Uses WizardUtils.openAddResourcesToSessionWizard(...) instead of calling
addReferencePointsToSession(...) directly in
SessionAddSelectedProjectsHandler. This was done to avoid having to add
additional logic to filter the resource selection and determine the
minimal set of base resources to actually use as the reference points.
As this logic is already present in the wizard, using it was the easiest
workaround. Furthermore, with the new behavior, the option fits much
better into the rest of the usage flow of the UI as all other option to
add reference points to the session go through the wizard.

</details>

### DIFF
--- a/eclipse/src/saros/ui/command_handlers/SessionAddSelectedProjectsHandler.java
+++ b/eclipse/src/saros/ui/command_handlers/SessionAddSelectedProjectsHandler.java
@@ -1,15 +1,24 @@
 package saros.ui.command_handlers;
 
+import java.util.HashSet;
 import java.util.List;
 import org.eclipse.core.commands.AbstractHandler;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
 import org.eclipse.core.resources.IResource;
 import saros.session.ISarosSession;
-import saros.ui.util.CollaborationUtils;
+import saros.ui.util.WizardUtils;
 import saros.ui.util.selection.retriever.SelectionRetrieverFactory;
 
-/** Handles the addition of selected {@link IResource}s to the running {@link ISarosSession}. */
+/**
+ * Handles the addition of selected {@link IResource}s to the running {@link ISarosSession}.
+ *
+ * <p>An {@link saros.ui.wizards.AddResourcesToSessionWizard} is opened for this purpose with the
+ * currently selected resources preselected.
+ *
+ * <p>This class is used to define the behavior of the project view context menu entry to add
+ * reference points to a running session.
+ */
 public class SessionAddSelectedProjectsHandler extends AbstractHandler {
 
   @Override
@@ -18,7 +27,8 @@ public class SessionAddSelectedProjectsHandler extends AbstractHandler {
     List<IResource> selectedResources =
         SelectionRetrieverFactory.getSelectionRetriever(IResource.class).getSelection();
 
-    CollaborationUtils.addResourcesToSession(selectedResources);
+    WizardUtils.openAddResourcesToSessionWizard(new HashSet<>(selectedResources));
+
     return null;
   }
 }
--- a/eclipse/src/saros/ui/util/CollaborationUtils.java
+++ b/eclipse/src/saros/ui/util/CollaborationUtils.java
@@ -58,33 +58,19 @@ public class CollaborationUtils {
   }
 
   /**
-   * Starts a new session and shares the given resources with given contacts.<br>
-   * Does nothing if a {@link ISarosSession session} is already running.
+   * Starts a new session and shares the given reference points with the given contacts.
    *
-   * @param resources
-   * @param contacts
+   * <p>Does nothing if a {@link ISarosSession session} is already running.
+   *
+   * @param referencePointContainers the containers to use for the reference points to share
+   * @param contacts the contacts to share the reference points with
    * @nonBlocking
    */
-  public static void startSession(final List<IResource> resources, final List<JID> contacts) {
-    // TODO remove assertion that only projects are shared (resolved in followup PR)
-    assert resources.stream().allMatch(resource -> resource instanceof IProject)
-        : "Encountered non-project resource to share";
+  public static void startSession(
+      final Set<IContainer> referencePointContainers, final List<JID> contacts) {
 
-    final Set<IProject> projects =
-        resources.stream().map(resource -> (IProject) resource).collect(Collectors.toSet());
+    Set<IReferencePoint> referencePoints = getReferencePoints(referencePointContainers);
 
-    startSession(projects, contacts);
-  }
-
-  /**
-   * Starts a new session and shares the given projects with given contacts.<br>
-   * Does nothing if a {@link ISarosSession session} is already running.
-   *
-   * @param projects the projects share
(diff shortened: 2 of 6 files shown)


eclipse/src/saros/ui/util/CollaborationUtils.java
```diff
@@ -58,44 +56,37 @@ private CollaborationUtils() {
...
    *
-   * @param projects the projects share
-   * @param contacts the contacts to share the projects with
+   * @param referencePoints the reference points to share
+   * @param contacts the contacts to share the reference points with
    * @nonBlocking
    */
-  public static void startSession(final Set<IProject> projects, final List<JID> contacts) {
+  public static void startSession(
+      final Set<IReferencePoint> referencePoints, final List<JID> contacts) {
```
srossbach, line 68: As this is mainly (only?) called by Eclipse what is the problem to chance this into a set of IContainer instead ?

eclipse/src/saros/ui/util/CollaborationUtils.java
```diff
@@ -207,38 +177,33 @@ public static void addResourcesToSession(Set<IProject> projectsToAdd) {
...
+          if (!session.hasWriteAccess()) {
+            DialogUtils.popUpFailureMessage(
+                Messages.CollaborationUtils_insufficient_privileges,
+                Messages.CollaborationUtils_insufficient_privileges_text,
+                false);
+            return;
+          }
 
-            sessionManager.addReferencePointsToSession(convert(projectsToAdd));
+          Set<IProject> projects =
```
srossbach, line 189: Duplicated code ? See line 76

tobous: Will extract it into a separate method.

eclipse/src/saros/ui/util/CollaborationUtils.java
tobous: I see your point. It would avoid the to unwrap the object to access the delegate. I will change it.

I implemented it this way to match the method in `ICollaborationUtils`. But since I removed that interface and its implementations (as it was only used by the HTML UI way back), that is no longer applicable.

tobous:

eclipse/src/saros/ui/util/CollaborationUtils.java
tobous: Done.

eclipse/src/saros/ui/util/CollaborationUtils.java
tobous: Done.

eclipse/src/saros/ui/command_handlers/SessionAddSelectedProjectsHandler.java
```diff
@@ -18,7 +27,8 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
     List<IResource> selectedResources =
         SelectionRetrieverFactory.getSelectionRetriever(IResource.class).getSelection();
 
-    CollaborationUtils.addResourcesToSession(selectedResources);
+    WizardUtils.openAddResourcesToSessionWizard(new HashSet<>(selectedResources));
```
srossbach, line 30: This is very inconvenient now. The correct rework would be to add the possibility to simply right-click on a folder or project and the click share.

eclipse/src/saros/ui/util/CollaborationUtils.java
```diff
@@ -276,6 +237,32 @@ public void run() {
...
+        .collect(Collectors.toSet());
+  }
+
+  /**
+   * Returns all projects that contain at least one of the given reference points.
+   *
+   * @param referencePoints the currently shared reference points
+   * @return all projects that contain at least one of the given reference points
+   */
+  private static Set<IProject> getReferencePointProjects(Set<IReferencePoint> referencePoints) {
```
srossbach, line 259: getProjects(FromReferencePoints) or convertToProjects ?

eclipse/src/saros/ui/command_handlers/SessionAddSelectedProjectsHandler.java
tobous: Yes, the usability is worse, but not by a lot in my opinion. Instead of directly adding the resource to the session, it now opens a dialog where the chosen resources are pre-selected. It is one additional click.

The issue with keeping the old logic is that the user can select arbitrary sets of resources, parts of which could already be shared or could be children of other selected resources. To process this, it would require applying all of the filtering logic done in the wizard. Furthermore, due to the filtering, the actually shared resources might not match the users expectations. But, as all of this happens in the background, we would not have any way of displaying this to the user. With the wizard, the user sees the actual result that the addition of the resources would have/which resources would actually be added to the session.

eclipse/src/saros/ui/util/CollaborationUtils.java
tobous: Will rename to `getProjects(...)`.

tobous:

eclipse/src/saros/ui/util/CollaborationUtils.java
tobous: Done.

srossbach (dismissed):

eclipse/src/saros/ui/util/CollaborationUtils.java
```diff
@@ -276,6 +266,32 @@ public void run() {
...
+  /**
+   * Returns all projects that contain at least one of the given reference points.
+   *
+   * @param referencePoints the currently shared reference points
+   * @return all projects that contain at least one of the given reference points
+   */
+  private static Set<IProject> getProjects(Set<IReferencePoint> referencePoints) {
+    return referencePoints
+        .stream()
+        .map(referencePoint -> ResourceConverter.getDelegate(referencePoint).getProject())
```
m273d15, line 262: Is it save to call `getProject` or should we check for `null`?

srossbach: `null` could only be returned if the reference point would be the workspace root. If you would like to add a sanity check you have to check that no passed in IContainer is the workspace root.

tobous:

tobous:

tobous:

tobous:

PR #1047, opened 2020-06-24
[BUILD] Workaround failing STF Tests

m273d15:
Eclipse dropped the GTK2 support [with 4.10](https://www.eclipse.org/eclipse/news/4.10/platform.php#gtk2-removal).
Our build process uses the latest compatible
bundle version (and the corresponding [new jar](https://mvnrepository.com/artifact/org.eclipse.platform/org.eclipse.swt/3.114.100) was released at 17.06.20).
Therefore, the new bundle version is used and we have to install GTK3.

However, this installation has to be moved into
a Dockerfile file in the corresponding repository.
I will open a follow-up PR.

### DIFF
--- a/travis/script/stf/master_setup_utils.sh
+++ b/travis/script/stf/master_setup_utils.sh
@@ -17,6 +17,8 @@ function start_container_master()
 
 function setup_container_master()
 {
+  # Workaround: Package installation has to be moved into a Dockerfile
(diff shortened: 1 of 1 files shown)


m273d15:
Another workaround was to pin the swt bundle version, but I prefer to keep the environment up-to-date.

PR #1048, opened 2020-06-25
Move XMPP Exception Parsing into Core

stefaus:
Moves XMPP exception parsing into Core and `IConnectionStateListener` now returns an `ErrorType` enum instead of an `Exception`.

### DIFF
--- a/eclipse/src/saros/ui/widgets/ConnectionStateComposite.java
+++ b/eclipse/src/saros/ui/widgets/ConnectionStateComposite.java
@@ -7,15 +7,13 @@ import org.eclipse.swt.custom.CLabel;
 import org.eclipse.swt.events.DisposeEvent;
 import org.eclipse.swt.events.DisposeListener;
 import org.eclipse.swt.widgets.Composite;
-import org.jivesoftware.smack.XMPPException;
-import org.jivesoftware.smack.packet.StreamError;
-import org.jivesoftware.smack.packet.XMPPError;
 import saros.SarosPluginContext;
 import saros.account.IAccountStoreListener;
 import saros.account.XMPPAccount;
 import saros.account.XMPPAccountStore;
 import saros.communication.connection.ConnectionHandler;
 import saros.communication.connection.IConnectionStateListener;
+import saros.communication.connection.IConnectionStateListener.ErrorType;
 import saros.context.IContextKeyBindings.SarosVersion;
 import saros.net.ConnectionState;
 import saros.repackaged.picocontainer.annotations.Inject;
@@ -49,7 +47,7 @@ public class ConnectionStateComposite extends Composite {
   private final CLabel stateLabel;
 
   private ConnectionState lastConnectionState;
-  private Exception lastError;
+  private ErrorType lastError;
 
   private final IAccountStoreListener accountStoreListener =
       new IAccountStoreListener() {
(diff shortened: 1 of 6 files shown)


saros-infrastructure:


```diff

Complexity decreasing per file
==============================
+ eclipse/src/saros/ui/widgets/ConnectionStateComposite.java  -6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5756916&bid=18844766)

eclipse/src/saros/ui/widgets/ConnectionStateComposite.java
```diff
@@ -122,7 +100,8 @@ public void widgetDisposed(DisposeEvent e) {
    * @param state the current connection state or <code>null</code>
    * @param error additional error information or <code>null</code>
    */
-  private void updateLabel(ConnectionState state, Exception error) {
+  private void updateLabel(ConnectionState state, ErrorType error) {
+    if (ConnectionStateComposite.this.isDisposed()) return;
```
srossbach, line 104: just isDisposed()

It is also questionable if we should do such checks in code that is normally executed when the widget is not disposed.

See ChatRoomsComposite where you did not moved the isDisposed check.

stefaus: Changed it.

Looks a bit superfluous but in some edge cases...

Yes, but didn't wanted to change the other listeners as well.

saros-infrastructure:


```diff

Complexity decreasing per file
==============================
+ eclipse/src/saros/ui/widgets/ConnectionStateComposite.java  -6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5756916&bid=18844766)

eclipse/src/saros/ui/widgets/ConnectionStateComposite.java
```diff
@@ -122,7 +100,8 @@ public void widgetDisposed(DisposeEvent e) {
    * @param state the current connection state or <code>null</code>
    * @param error additional error information or <code>null</code>
    */
-  private void updateLabel(ConnectionState state, Exception error) {
+  private void updateLabel(ConnectionState state, ErrorType error) {
+    if (isDisposed()) return;
```
srossbach, line 104: Thanks for violating guidelines.

See https://www.ibm.com/support/knowledgecenter/en/SS8PJ7_9.5.0/org.eclipse.platform.doc.isv/reference/api/org/eclipse/swt/widgets/Widget.html#checkWidget()

Basically this means that you should call this method in every public method of a widget.

It is just bad coding to check if the widget itself has disposed and silently drop any further action (even in private methods).

There is a reason the check is always done in the runable and not in the called method.

Imagine you introduce a public method that internally calls updateLabel and forget the checkWidget call. Nothing will happen if the widget is already disposed and you will silently swallow the wrong access to the widget.

Even if this might not be a big deal the raised exception could be more helpful to figure out what is going wrong instead of watching the UI and pondering why it is not correctly updating.

stefaus: > Thanks for violating guidelines.

I find your tone passively aggressive and would appreciate it if you would change it.

I moved the check from the listeners to this position. And the check was introduced a while ago by yourself:
https://github.com/saros-project/saros/commit/6c6faca6d871d3f7c224961db95aa2e04b0d47b2#diff-66f12eebcb2e00c1fb289219b89fead8R57-R58 So I'm not sure why your argue like that. What is now added is that it is also checked in the constructor execution (which I would say doesn't change something).

---

From your linked doc:
>Currently, it is an error to invoke any method (other than isDisposed()) on a widget that has had its dispose() method called.

What I read is use checkWidget if you would like to have a exception, use isDisposed if you just want to check. Do you see something that should be done in the disposed state?
The check is like before still done in the context of the runnable thread.
Arguing about the visibility of the check in cases of for example additions to the listeners, Ok this is maybe a thing but a reason for this reviews.
Same for errors in new methods and I don't see how it helps in these cases to do the check in the listener. When the widget is disposed the doc says to not call methods on it and there is nothing which is displayed, so what would you update? On dispose code should be handled in a dispose listener.

PR #1058, opened 2020-07-02
Add sanity check to prevent nested reference points

tobous:
Adds a sanity check ensuring the no nested reference points can be created. As this is only a sanity check, it does not provide the best user experience. Nested reference points should be prevented by the IDE-specific UI not allowing such resource choices when starting or receiving resource negotiations.

### Reviewing this PR

### Commits

<details><summary><b>[INTERNAL] Add method to check for nested reference points</b></summary>
<br>

Adds IReferencePoint.isNested(IReferencePoint), providing the logic to
check whether the given reference point would create a nested
configuration with the called-on reference point.

Adds the IDE implementations for the new method.

</details>

<details><summary><b>[INTERNAL][CORE] Add sanity check preventing nested reference points</b></summary>
<br>

Adds a sanity check to the logic adding reference points to the
SharedReferencePointMapper to prevent nested reference points from being
created. If adding a reference point would create such nested reference
points, an IllegalStateException is thrown instead.

This is only meant as a sanity. Normally, this should already be
prevented by the IDE-specific UI implementation.

</details>

<details><summary><b>[REFACTOR][CORE] Clean up SharedReferencePointMapper</b></summary>
<br>

Removes unnecessary explicit type arguments.

Moves the instantiation of the contained maps to the constructor.

Replaces an explicit addition of a range of elements to a newly created
list with using the corresponding ArrayList constructor.

Casts a resource object to IReferencePoint before checking whether it is
contained in the mapping to ensure type safety.

</details>

### DIFF
--- a/core/src/saros/filesystem/IReferencePoint.java
+++ b/core/src/saros/filesystem/IReferencePoint.java
@@ -20,6 +20,17 @@ public interface IReferencePoint extends IContainer {
   IPath getReferencePointRelativePath();
 
   /**
+   * Returns whether this reference point in combination with the other reference point represents a
+   * nested configuration, either because this reference point is contained in the other reference
+   * point or the other reference point is contained in this reference point.
+   *
+   * @param otherReferencePoint the other reference point with which to check
+   * @return whether this reference point in combination with the other reference point represents a
+   *     nested configuration
+   */
+  boolean isNested(IReferencePoint otherReferencePoint);
+
+  /**
    * Always throws an IO exception.
    *
    * <p>Deleting a reference point resource is not supported.
--- a/core/src/saros/session/internal/SharedReferencePointMapper.java
+++ b/core/src/saros/session/internal/SharedReferencePointMapper.java
@@ -83,10 +83,32 @@ class SharedReferencePointMapper {
               + currentReferencePoin);
     }
 
+    checkForNestedReferencePoints(referencePoint);
+
     idToReferencePointMapping.put(id, referencePoint);
     referencePointToIDMapping.put(referencePoint, id);
 
     log.debug("added reference point " + referencePoint + " with ID " + id);
+  }
+
+  /**
+   * Checks whether adding the given reference point to the mapping would create shared nested
+   * reference points.
+   *
+   * @param addedReferencePoint the reference point to add to the mapping
+   * @throws IllegalStateException if nested reference points are detected
+   */
+  private void checkForNestedReferencePoints(IReferencePoint addedReferencePoint) {
+    for (IReferencePoint sharedReferencePoint : idToReferencePointMapping.values()) {
+      if (addedReferencePoint.isNested(sharedReferencePoint)) {
+        throw new IllegalStateException(
+            "Reference point "
+                + addedReferencePoint
+                + " can not be added to the mapping as this would create nested reference points "
+                + "in combination with the already shared reference point "
+                + sharedReferencePoint);
+      }
+    }
   }
 
   /**
--- a/core/src/saros/session/internal/SharedReferencePointMapper.java
+++ b/core/src/saros/session/internal/SharedReferencePointMapper.java
@@ -27,19 +27,23 @@ class SharedReferencePointMapper {
   private static final Logger log = Logger.getLogger(SharedReferencePointMapper.class);
 
   /** Mapping from reference point IDs to currently registered shared reference points. */
-  private final Map<String, IReferencePoint> idToReferencePointMapping =
-      new HashMap<String, IReferencePoint>();
+  private final Map<String, IReferencePoint> idToReferencePointMapping;
 
   /** Mapping from currently registered shared reference points to their id's. */
-  private final Map<IReferencePoint, String> referencePointToIDMapping =
-      new HashMap<IReferencePoint, String>();
+  private final Map<IReferencePoint, String> referencePointToIDMapping;
 
   /**
    * Map for storing which clients have which reference points. Used by the host to determine who
    * can currently process an activity related to a particular reference point. (Non-hosts don't
    * maintain this map.)
    */
-  private final Map<User, List<String>> referencePointsOfUsers = new HashMap<User, List<String>>();
+  private final Map<User, List<String>> referencePointsOfUsers;
+
+  SharedReferencePointMapper() {
+    this.idToReferencePointMapping = new HashMap<>();
+    this.referencePointToIDMapping = new HashMap<>();
+    this.referencePointsOfUsers = new HashMap<>();
+  }
 
   /**
    * Adds a reference point to the set of currently shared reference points.
@@ -162,8 +166,10 @@ class SharedReferencePointMapper {
   public synchronized boolean isShared(IResource resource) {
     if (resource == null) return false;
 
-    if (resource.getType() == REFERENCE_POINT)
-      return idToReferencePointMapping.containsValue(resource);
+    if (resource.getType() == REFERENCE_POINT) {
+      IReferencePoint referencePoint = (IReferencePoint) resource;
+      return idToReferencePointMapping.containsValue(referencePoint);
+    }
 
     IReferencePoint referencePoint = resource.getReferencePoint();
 
@@ -178,7 +184,7 @@ class SharedReferencePointMapper {
    * @return a newly created {@link Set} with the shared reference points
    */
   public synchronized Set<IReferencePoint> getReferencePoints() {
-    return new HashSet<IReferencePoint>(idToReferencePointMapping.values());
+    return new HashSet<>(idToReferencePointMapping.values());
   }
 
   /**
@@ -217,10 +223,7 @@ class SharedReferencePointMapper {
    * @see #userHasReferencePoint(User, IReferencePoint)
    */
   public synchronized void addMissingReferencePointsToUser(User user) {
-    List<String> referencePointIds = new ArrayList<String>();
-    for (String referencePointId : idToReferencePointMapping.keySet()) {
-      referencePointIds.add(referencePointId);
-    }
+    List<String> referencePointIds = new ArrayList<>(idToReferencePointMapping.keySet());
 
     this.referencePointsOfUsers.put(user, referencePointIds);
   }
--- a/eclipse/src/saros/filesystem/EclipseReferencePoint.java
+++ b/eclipse/src/saros/filesystem/EclipseReferencePoint.java
@@ -5,6 +5,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import org.apache.log4j.Logger;
+import org.eclipse.core.resources.IContainer;
 import org.eclipse.core.runtime.CoreException;
 
 /** Eclipse implementation of the Saros reference point interface. */
@@ -43,6 +44,16 @@ public class EclipseReferencePoint implements IReferencePoint {
   @Override
   public EclipsePathImpl getReferencePointRelativePath() {
     return new EclipsePathImpl(org.eclipse.core.runtime.Path.EMPTY);
+  }
+
+  @Override
+  public boolean isNested(IReferencePoint otherReferencePoint) {
+    org.eclipse.core.runtime.IPath p1 = delegate.getFullPath();
+
+    IContainer d2 = ResourceConverter.getDelegate(otherReferencePoint);
+    org.eclipse.core.runtime.IPath p2 = d2.getFullPath();
+
+    return p1.equals(p2) || p1.isPrefixOf(p2) || p2.isPrefixOf(p1);
   }
 
   @Override
--- a/intellij/src/saros/intellij/filesystem/IntellijReferencePoint.java
+++ b/intellij/src/saros/intellij/filesystem/IntellijReferencePoint.java
@@ -110,6 +110,16 @@ public class IntellijReferencePoint implements IReferencePoint {
     return IntellijPath.EMPTY;
   }
 
+  @Override
+  public boolean isNested(IReferencePoint otherReferencePoint) {
+    Path p1 = Paths.get(virtualFile.getPath());
+
+    IntellijReferencePoint i2 = (IntellijReferencePoint) otherReferencePoint;
+    Path p2 = Paths.get(i2.getVirtualFile().getPath());
+
+    return p1.equals(p2) || p1.startsWith(p2) || p2.startsWith(p1);
+  }
+
   /**
    * Returns the path to the given file relative to the virtual file represented by this reference
    * point.
--- a/server/src/saros/server/filesystem/ServerProjectImpl.java
+++ b/server/src/saros/server/filesystem/ServerProjectImpl.java
@@ -4,6 +4,8 @@ import static saros.filesystem.IResource.Type.REFERENCE_POINT;
 
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import saros.filesystem.IReferencePoint;
 
 /** Server implementation of the {@link IReferencePoint} interface. */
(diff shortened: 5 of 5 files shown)


saros-infrastructure:


```diff

Complexity increasing per file
==============================
- eclipse/src/saros/filesystem/EclipseReferencePointComparator.java  1
- intellij/src/saros/intellij/filesystem/IntellijReferencePointComparator.java  1
- core/src/saros/filesystem/ReferencePointComparator.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5796923&bid=18973332)

tobous:
Adding a new interface and classes to check whether reference points are nested is not the nicest solution, but I could not think of a nicer way of comparing the relation between them that is accessible to the core. This would previously have been done by using the "workspace relative path", but since this is not a concept that works for IntelliJ, this is not an option in my opinion.

But, as the interface is relatively general, it can be extended later on to also offer other comparison methods for reference points that might be necessary/useful later on.

core/src/saros/session/internal/SarosSession.java
```diff
@@ -861,4 +877,42 @@ boolean hasActivityProducers() {
...
+   */
+  // TODO show user error message
+  private void leaveSessionDueToNestedReferencePoint(IReferencePoint addedReferencePoint) {
+    log.error(
+        "adding reference point "
+            + addedReferencePoint
+            + " would lead to nested reference points; ending session - shared reference points: "
+            + getReferencePoints());
+
+    sessionContainer.getComponent(ISarosSessionManager.class).stopSession(LOCAL_USER_LEFT);
```
srossbach, line 916: Beside the fact that aborting the session is very bad as it should not happen (in former version the project was simply not shared, ok not even better but ...) I highly recommend to not call stopSession inside the Session class.

core/src/saros/session/internal/SarosSession.java
```diff
@@ -724,6 +734,12 @@ public IReferencePoint getReferencePoint(String referencePointID) {
   @Override
   public void addReferencePointMapping(String referencePointId, IReferencePoint referencePoint) {
     if (referencePointMapper.getReferencePoint(referencePointId) == null) {
+      if (wouldCreateNestedReferencePoint(referencePoint)) {
+        leaveSessionDueToNestedReferencePoint(referencePoint);
```
srossbach, line 738: Just thrown an Exception.

core/src/saros/filesystem/ReferencePointComparator.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Utility class providing static access to the {@link IReferencePointComparator} injected by the
+ * IDE-specific implementation.
+ */
+public class ReferencePointComparator {
+
+  @Inject private static IReferencePointComparator referencePointComparator;
+
+  static {
+    SarosPluginContext.initComponent(new ReferencePointComparator());
```
srossbach, line 15: I know there are several classes which are full of such hacks, but the way you use it among the SarosSession / ReferencePointMapper is evil. Basically it is spaghetti / static method code that is very hard to test.

As you have the container when the session is started you can retrieve it there instead of calling a static method.

tobous: This class is ideal to offer static functionality. It is not dependent on any instance variables and offers general utility methods.

The issue is that the functionality has to be provided by the IDE-specific implementations, meaning the methods providing the functionality can't be static themselves. But having them be normal class and requesting the from the context if needed only further bloats up the constructors of our classes. A prime example for this is the path factory (or, as an extension, its Eclipse implementation). There is no need for it to be non-static besides it being injected by the IDE implementations. Its usage becomes so much more of a hassle just because it has to be requested from the context whenever it is needed. (I know that these classes aren't even used that much anymore. This is more meant as a general example.)

And while this class specifically can't be tested, that should not be an issue as it is only a surrogate to make accessing the actual implementations easier. This does not change anything about the testing behavior of the actual implementations. It slightly changes the testing behavior for testing classes that use the logic, but this can either be resolved by adding a mock to the context if it is used (which also needs to happen when testing the context without the static logic) or by mocking the static call using powermock.

Also, this setup would be a lot cleaner and far less "dirty"/"evil" if we were to add a getter for the plugin context (just like for the session context). This would allow us to avoid the reflection stuff and would replace the static block with just initializing the static variable using the getter. As this getter would also easily be mockable using powermock, stubbing the whole utility could be done using a test utility method.

Also, I don't get what the ReferencePointMapper has to do with this.

core/src/saros/session/internal/SarosSession.java
tobous: See other comment. In what state does this leave the session and the running resource negotiation?

core/src/saros/filesystem/ReferencePointComparator.java
srossbach: The correct implementation regarding Unit testing of the core would be to implement a static util class in Eclipse, IntelliJ, and the Server. Afterwards you add just a IReferencePointComparer to the session context which just delegates to the static calls.

With your current implementation it is impossible to do a minimalistic instantiation of the SarosSession class with because of the static call to (I should have deleted this shit method) SarosPluginContext.initComponent class. It is simply not how DI works.

core/src/saros/session/internal/SarosSession.java
srossbach: In the same broken state if you just press cancel during the resource negotiation. Still not fixed since years. Maybe this is a good start before thinking about other issues like this cornercase.

core/src/saros/session/internal/SarosSession.java
srossbach: If you want to cancel the session do it async. Fork a thread.

srossbach: IIRC stopSession will wait on existing resource negotiation etc. However this will be called inside a resource negotiation. Guess you will run into a nasty timeout with will continue afterwards in an unstable state.

tobous: Moved to new thread. Works when starting a session or adding resources to the session. I tested this using Saros/E for the sending and receiving side.

> IIRC stopSession will wait on existing resource negotiation etc.

[`SarosSessionManager.stop()` closes all running negotiations.](https://github.com/saros-project/saros/blob/master/core/src/saros/session/SarosSessionManager.java#L327).

core/src/saros/filesystem/ReferencePointComparator.java
srossbach: When it comes to testing assume that mocking static methods is just a "bonus" and not the way to go.

In order to test this this I have to set a context. When I use the default context it starts creating an XMPP store etc. If I want to use another implementation I have to write my own initComponent method (which it btw marked as deprecated).

Can you please tell me in which places in the Core Annotation injection is used. We had an agreement to not use them in the Core at all, only CTOR injection.

tobous: > When it comes to testing assume that mocking static methods is just a "bonus" and not the way to go.
> 
> In order to test this this I have to set a context. When I use the default context it starts creating an XMPP store etc. If I want to use another implementation I have to write my own initComponent method

Which is exactly what happens for `SarosSessionTest`. A general solution for this problem would be to create a test context factory that allows registering classes to mock (and optionally also which objects to use) and then creates a functioning mock context, including the injection logic.

> (which it btw marked as deprecated).

Yes, I noticed that as well. But the marking lacks any indication as to why and what to use instead. So I just assumed that this is something someone set at some point and then forgot about. Also, only the context method is marked as deprecated and not the static wrapper that is used to normally access it.

> Can you please tell me in which places in the Core Annotation injection is used.

As far as I can tell, the only class using annotation injection in the core (prior to this change) is `SarosSession` (see [variables](https://github.com/saros-project/saros/blob/master/core/src/saros/session/internal/SarosSession.java#L73) and [initialization](https://github.com/saros-project/saros/blob/master/core/src/saros/session/internal/SarosSession.java#L790)).

> We had an agreement to not use them in the Core at all, only CTOR injection.

I am not quite sure who "we" is in this context. I was not aware of any such agreement/rule and could not find any mention of it in the [new](https://www.saros-project.org/contribute/best-practices.html) or [old](https://saros-project.github.io/legacy_docs/coderules.1.html#dependency_injection) documentation.

If this is indeed something we want to avoid in the core, this is something we should definitely mention in the documentation. But, before we make such a decision, there should be a clear reasoning as to why.

In this discussion, it still has not become clear to me what the issue with annotation injection or its current usage is. Yes, testing gets harder, but not by that much in my opinion (but maybe I am already "numbed" by all the testing difficulties that arise with the usage of static IDE methods or other parts of the IDE context :sweat_smile:). Furthermore, the difficulties can be largely mitigated by introducing test helper classes.

Also, as [the old documentation already stated](https://saros-project.github.io/legacy_docs/coderules.1.html#dependency_injection), static access to the context is not always avoidable, e.g. when dealing with UI classes that are not initialized by Saros but by the IDE, so I don't think we can get rid of annotation injection (at least not without replacing it with another way of statically accessing the context).

srossbach: It all boils down regarding testing. Of course you can just use one argument. Every class will just get an IContext object and then calls getComponent to until the world collides. Just add a small taste of static massive calls that require even more setup, either through more static calls before calling the static call or even a more complex setup.

If you strictly follow this approach you will end up with classes that you simply do not want / or simply cannot test (and I have encountered that in my last job, you could not simply test one class in isolation, static method calls all over the place and there was no free of charge tool for C# that was able to mock static methods).

If you look at the SarosSessionTest, it is just a smoke test with an massive amount of setup. In reality it only checks if the forked PicoContainer is correctly detached and garbage collected.

We currently do not have any JUnit integration test(s) but only the STF. Of course in the STF everything is available but regarding JUnit integration tests and unit test every static method call that relies on other object that has to be properly initialized is just annoying.

As far as I remember the SarosPluginContext class was moved to the Core to avoid code duplication.

There was not really a discussion if X small line LOC classes in every plugin are more harmful than adding  this "EierlegendeWollmilchsau" to the Core and made it available to everything, including the Core.

It also do not really has something to do with Annotation Injection. That is just a matter of taste. In a real DI setup you throw everything into the DI framework and the classes are properly initialized.

However if you look at the implementation of the IContext#initComponent method it simply violates these principles.

To sum it up. You have the SarosSession class. It has full access to the the DI container. It creates a child container and get the factory for this child container from the DI context. Now after everything should be initialized you still have classes which will access static method calls which will start DI injection (lazy) again. To be more specific, if you call IContext#initComponent you are not part of the DI container. If such code is called from code that is part of a DI container then this is just bad design.

tobous: Removed.

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- intellij/src/saros/intellij/filesystem/IntellijReferencePointComparator.java  1
- eclipse/src/saros/filesystem/EclipseReferencePointComparator.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5796923&bid=18973332)

core/src/saros/session/internal/SarosSession.java
```diff
@@ -861,4 +881,62 @@ boolean hasActivityProducers() {
...
+   * <p>Stops the session using a different thread to prevent lockups.
+   *
+   * <p>This method should be called to end the running session if a user tried to add a new
+   * reference point that would lead to nested reference points being shared.
+   *
+   * @param addedReferencePoint the newly added reference point
+   * @see #wouldCreateNestedReferencePoint(IReferencePoint)
+   */
+  // TODO show user error message
+  private void leaveSessionDueToNestedReferencePoint(IReferencePoint addedReferencePoint) {
```
srossbach, line 915: Looking at the negotiation code, throwing just an exception will:

1. stop the session on the remote side (this is what you want)
1.1 this is not true if the receiving side is the host, see https://github.com/saros-project/saros/blob/master/core/src/saros/negotiation/AbstractIncomingResourceNegotiation.java#L303
2. presents an unknown error containing the message of the exception (not the best option but better than this here)

I do not looked what will happen if the sharing side will trigger this, but I think if this happens on the sharing side it will do no harm at all.

I think a possible "fix" is to modify the IPN in such way that it kicks out the remote user when the receiving side is the host.

tobous: But I don't get what the issue with the current approach is. The session teardown seems to work just fine. I haven't tried non-host project sharing though.

(Also, I would like to avoid putting too much additional work into this as I am moving on to writing the actual thesis part of the thesis.)

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- server/src/saros/server/filesystem/ServerReferencePointComparator.java  1
- eclipse/src/saros/filesystem/EclipseReferencePointComparator.java  1
- intellij/src/saros/intellij/filesystem/IntellijReferencePointComparator.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5796923&bid=18973332)

tobous:
Rebased onto current base branch only resolving minor merge conflicts.

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- intellij/src/saros/intellij/filesystem/IntellijReferencePointComparator.java  1
- eclipse/src/saros/filesystem/EclipseReferencePointComparator.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5796923&bid=18973332)

tobous:

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- intellij/src/saros/intellij/filesystem/IntellijReferencePointComparator.java  1
- eclipse/src/saros/filesystem/EclipseReferencePointComparator.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5796923&bid=18973332)

tobous:

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- eclipse/src/saros/filesystem/EclipseReferencePointComparator.java  1
- intellij/src/saros/intellij/filesystem/IntellijReferencePointComparator.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5796923&bid=18973332)

tobous:
Rebased onto current base branch only resolving minor merge conflicts.

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- intellij/src/saros/intellij/filesystem/IntellijReferencePointComparator.java  1
- eclipse/src/saros/filesystem/EclipseReferencePointComparator.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5796923&bid=18973332)

tobous:
@srossbach Do you still have issues with how the logic leaves the session (or other parts of the changes)?

core/src/saros/session/internal/SarosSession.java
srossbach: I still suggest to move the introduced logic here to the ReferencePointMapper. And as I already stated just throw an exception for now. We already have an issue open for normal/abnormal cancellation of a current project negotiation. If you still want to abort the session do it in the IPN. But be aware that the IPN is executed on the server side in the Saros server project.

tobous:
Rebased onto current master without any changes or interactions.

saros-infrastructure:


```diff

Complexity increasing per file
==============================
- intellij/src/saros/intellij/filesystem/IntellijReferencePointComparator.java  1
- eclipse/src/saros/filesystem/EclipseReferencePointComparator.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/Saros/saros/pullRequest?prid=5796923&bid=18973332)

core/src/saros/session/internal/SarosSession.java
tobous: Moving this check to the `ReferencePointMapper` is kind of messy. The check requires the `IReferencePointComparator` instance retrieved from the session context, but the mapper is instantiated before the context.

So I had to move the instantiation of the reference point mapper below the session context initialization.

srossbach: Another solution would be to move the method into the IReferencePoint interface.

tobous: Yeah, that is a much better/cleaner approach. Will do. :+1:

tobous: Moved the check to `SharedReferencePointMapper`. Also removes the logic leaving the session and just throws an exception.

srossbach (approved):

PR #1061, opened 2020-07-10
[INTERNAL][CORE] Pass JID of local user to session CTOR

tobous:
Avoids the dependency on the connection service to initialize the
session context by passing the JID of the local user to the session
constructor.

Removes the method to resolve plugin context dependencies from
SarosSession as it was no longer necessary. Moved the error handling
in-line for the last usage.

Updates the session tests for the new CTOR argument.

### DIFF
--- a/core/src/saros/session/internal/SarosSession.java
+++ b/core/src/saros/session/internal/SarosSession.java
@@ -42,7 +42,6 @@ import saros.filesystem.IResource;
 import saros.net.IConnectionManager;
 import saros.net.ITransmitter;
 import saros.net.xmpp.JID;
-import saros.net.xmpp.XMPPConnectionService;
 import saros.preferences.IPreferenceStore;
 import saros.repackaged.picocontainer.MutablePicoContainer;
 import saros.repackaged.picocontainer.PicoContainer;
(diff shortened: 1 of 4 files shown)


core/src/saros/session/internal/SarosSession.java
```diff
@@ -787,19 +788,15 @@ private SarosSession(
...
-
     this.sessionID = id;
     this.referencePointMapper = new SharedReferencePointMapper();
     this.activityQueuer = new ActivityQueuer();
     this.containerContext = context;
 
-    // FIXME that should be passed in !
-    JID localUserJID = connectionService.getJID();
-
     assert localUserJID != null;
```
srossbach, line 800: I think the assertion can be removed here. Either move it to the SessionManager or let the SessionManager correctly handle null values (which is not likely to happen but can theoretically occur if the network goes down etc.)

tobous: That would duplicate the null check. I did not really see a reason to change the current check as the setup did not really change. The only difference is that the local JID is requested a bit earlier.

srossbach: I cannot find the null check. As long as I remember we simply assumed that a connection is established. Hence the assertion. Even if it does not halt true all the time.

tobous: > I cannot find the null check.

This is the "null check". I just meant checking before calling requires to check everywhere where the method is called (i.e. in two different places with this method) while having it inside the method unifies the checks. But that just leads us back to the basic question of who should check the parameters, the caller or the called method.

I can make this an `IllegalArgumentException` if you would prefer.

PR #1063, opened 2020-07-13
[BUILD] Use picocontainer jar

m273d15:
Use the jar created by the repackaged project
saros-project/repackaged-picocontainer.


m273d15:
The core is used to export the dependencies to the eclipse OSGi bundle (and also the stf bundle). Therefore, multiple dependencies are located in the core project, but these dependencies are also used by other Saros components and could be moved in the root project.

PR #1067, opened 2020-07-20
[INTERNAL][CORE] Adds security check for received activities

srossbach:
It was possible to impersonate yourself as another session user
(required a custom modification).

### DIFF
--- a/core/src/saros/session/internal/ActivitySequencer.java
+++ b/core/src/saros/session/internal/ActivitySequencer.java
@@ -339,7 +339,7 @@ public class ActivitySequencer implements Startable {
       return;
     }
 
-    sarosSession.exec(activities);
+    sarosSession.exec(sender, activities);
   }
 
   /** Sends an activity to the given recipients. */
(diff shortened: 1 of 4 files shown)


core/src/saros/session/internal/ActivitySequencer.java
```diff
@@ -359,7 +359,7 @@ public void sendActivity(List<User> recipients, final IActivity activity) {
           new Runnable() {
             @Override
             public void run() {
-              sarosSession.exec(Collections.singletonList(activity));
+              sarosSession.exec(user.getJID(), Collections.singletonList(activity));
```
m273d15, line 362: `getJID` is deprecated. However, I found no alternative method (as e.g. `getXMPPContact`/`getContact`).
@stefaus How would you get the Contact/JID?

stefaus: At the moment there is no proper replacement for everything where you need the contact plus the correct resource endpoint. `XMPPContact` can return the current best available `JID`, but this could change at any time.

the `user.getJID()` deprecation is really old, but until we have something better working this PR looks good to me

stefaus (approved):

PR #1068, opened 2020-07-23
[DOC] Add info where to get a JDK8

m273d15:
Oracle does not provide a public JDK 8 version.
Therefore, we provide a link to alternative sources.


srossbach:
My Java JDK 8 is from Oracle. Did I missed something ?

tobous:
> Oracle does not provide a public JDK 8 version.
> Therefore, we provide a link to alternative sources.

From what I can tell, Oracle still provides a JKD 8: https://www.oracle.com/java/technologies/javase/javase-jdk8-downloads.html

stefaus:
For the Oracle JDK download page you need to an Oracle account... that's why it's not public.

PR #1075, opened 2020-07-31
[NOP] Rename master->coordinator, slave->worker

m273d15:
Rename performed automatically (via `sed`). The only exception is `stf_worker_image="saros/stf_test_slave:0.6"`, because
this lines requires to rename the image in `saros-project/dockerfiles`.

I used "coordinator" because the STF tests executed in a coordinator container coordinate the
interactions of different eclipse instances/workers. Probably, "runner" is also a good name.


m273d15:
> Are there issues with renaming the docker image? Otherwise, it would be nice to keep it consistent.

I am currently preparing the commit.

PR #1084, opened 2020-09-11
[STF] Adds direct access to open an editor

srossbach:
Adds support for bypassing the GUI when opening an editor for a specific
file.

### DIFF
--- a/stf/META-INF/MANIFEST.MF
+++ b/stf/META-INF/MANIFEST.MF
@@ -16,7 +16,8 @@ Require-Bundle: org.eclipse.ui,
  org.eclipse.jdt.launching;resolution:=optional,
  saros.core,
  saros.eclipse,
- org.junit
+ org.junit,
+ org.eclipse.ui.ide
 Bundle-ClassPath: .,
  lib/org.eclipse.swtbot.ant.junit_2.7.0.201806111355.jar,
  lib/org.eclipse.swtbot.e4.finder_2.7.0.201806111355.jar,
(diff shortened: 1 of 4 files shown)


tobous:
These changes are no longer necessary in my opinion. It was not an issue with the upstream API but rather with the setup of the projects on the invitee's side. I described this in more detail in #1040.

srossbach:
Thanks for spending so much time to find the issue 👍 . I am still confused why user input on the IDE is necessary to detect project nature changes. Looking at the documentation this should be done on the fly when you modify the .project file. I looked at the modifications but in my opinion it still hard to understand what is going on, especially for new developers. You read the code and ask yourself, why does I have to create an empty stub project ? This should be hidden somewhere in the Util / Superbot methods in my opinion. Still wondering if a simply project refresh could do the trick although t his will increase testing time again.

tobous:
> Thanks for spending so much time to find the issue  :+1:. I am still confused why user input on the IDE is necessary to detect project nature changes. Looking at the documentation this should be done on the fly when you modify the .project file.

I think it does happen in the background occasionally, which is why some of the tests succeeded some of the time. But this background-refresh probably only happens periodically, e.g. every 10 seconds or so. At least that would be my guess.

> I looked at the modifications but in my opinion it still hard to understand what is going on, especially for new developers. You read the code and ask yourself, why does I have to create an empty stub project? This should be hidden somewhere in the Util / Superbot methods in my opinion.

This is mostly hidden in the Util method. The only case where this is not hidden is when a single participants participant is added to a running session using the super bot directly instead of utility methods. I decided to adjust the code to create the project and use it instead of re-writing the tests.

> Still wondering if a simply project refresh could do the trick although t his will increase testing time again.

I tried fixing this by refreshing the project on the sending and receiving side but it was very inconsistent. I am not sure why, but my guess would be that the timings during the test are quite tight and the refresh happened while the resources were already being processed.

tobous:
@srossbach I don't see a reason to keep this PR around. Do you still have a use for it? Otherwise I would close it.

tobous:
Superseded by #1040.

PR #1095, opened 2020-11-03
Replace option to create projects as part of resource negotiation

### LINKED ISSUE #1094: Using "Create New Project" option creates wrong directory structure
Inviter folder _src_ of project _A_

On the receiving side select **Create New Project**

First issue: The suggested project name is _src_ although it should be _A_ .

After the project is created the directory layout does not match.

On the receiving side the _src_ folder is not present, instead the whole content of _src_ is directly placed under the project path. I am aware that this should actually be the case but regarding the logic so create a new **Project** this might not be very helpful.

tobous:
Removes the option to create a project to use for the incoming resource negotiation to represent the shared reference point. Instead adds a button to the wizard allowing the user to manually create and configure a new project. This guarantees that the wizard is still completable, even if the workspace is initially empty.

Adjusts the STF to match these changes. Adds new utility methods to set up sessions by creating stub projects on the receiving side. This matches the behavior for the setups providing projects with language support. Subsequently adjusts the existing tests still using the option `NEW_PROJECT`. Removes all methods and constants corresponding to this option.

Also removes the option `EXISTING_PROJECT_WITH_COPY` from the STF methods. It was only used in ignored tests and is no longer supported by the Saros/E UI. Removes the ignored tests as well.

Resolves #1094.

#### Reviewing This PR

#### Commits

<details><summary><b>[UI][E] Remove option to create project as part of resource negotiation</b></summary>
<br>

Removes the option to create a project as part of the incoming resource
negotiation. This option will be replaced with a button offering the
user the functionality to manually create a new project that can then be
used for the incoming resource negotiation.

This avoids potential issues caused by incorrectly set up projects as
part of the negotiation. Since Saros no longer supports correctly
configuring the project, it is best to completely avoid creating it and
rather pass this responsibility to the user.

</details>

<details><summary><b>[UI][E] Add project creation button to resource negotiation wizard</b></summary>
<br>

Adds a button to create a new project to the wizard used to accept
incoming resource negotiations. This button can be used by the user to
create and configure new projects that can then subsequently be used to
represent the shared reference points in the local workspace.

This option is necessary to ensure that the wizard is always
completable, even in cases where the workspace is empty when the wizard
is started.

As the button is part of the external wizard and not the selection tab
itself, creating a new project through the button does not update the
currently selected tab. This has to be done manually by the user.

</details>

<details><summary><b>[STF] Add modality option to Util.setUpSessionWithProjectAndFile()</b></summary>
<br>

Adds the option to pass the used session invitation modality state to
`Util.setUpSessionWithProjectAndFile(...)`. This was done in preparation
for an upcoming change adding more usages of the method requiring
different modality states.

All existing usages of the method were adjusted to use the "concurrent"
session invitation modality. This was the previous default for the
method.

</details>

<details><summary><b>[STF] Add utility methods for setups using existing non-java projects</b></summary>
<br>

Adds `Util.setUpSessionWithProject(...)`, providing the functionality to
set up a session without Java language support with the given project.
Project stubs are created on the receiving side, allowing the
negotiation to use `EXISTING_PROJECTS`.

This method was added to replace the usage of the mode `NEW_PROJECT`,
which is no longer available.

Amends `Util.setUpSessionWithProjectAndFile(...)` to also create and use
project stubs on the receiving side.

</details>

<details><summary><b>[STF] Adjust tests to no longer use option "NEW_PROJECT"</b></summary>
<br>

Adjusts the existing STF tests relying on the resource negotiation to
create the shared projects. This option was removed from the UI.

Instead, such tests now create a stub project on the receiving side and
use the option `EXISTING_PROJECT` instead.

Adjusted some test setups to use the available utility methods to
simplify the session setup if possible.

</details>

<details><summary><b>[STF] Remove methods supporting "NEW_PROJECT" negotiation mode</b></summary>
<br>

Removes the methods providing access to the resource negotiation mode
creating a new project on the receiving side as this mode is no longer
available.

Subsequently removes constants only necessary for these methods.

</details>

<details><summary><b>[STF] Remove methods supporting "EXISTING_PROJECT_WITH_COPY" mode</b></summary>
<br>

Removes methods supporting the creation of a project copy on the
receiving side as part of the resource negotiation as it is no longer
available.

Removes the tests for this mode. All of them were ignored anyway.

Removes the constants only used in the removed methods.

</details>

### DIFF
--- a/eclipse/src/saros/ui/wizards/pages/LocalRepresentationSelectionPage.java
+++ b/eclipse/src/saros/ui/wizards/pages/LocalRepresentationSelectionPage.java
@@ -1,11 +1,7 @@
 package saros.ui.wizards.pages;
 
-import static saros.ui.widgets.wizard.ReferencePointOptionComposite.LocalRepresentationOption.NEW_PROJECT;
-
-import java.io.File;
 import java.nio.charset.Charset;
 import java.text.MessageFormat;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -353,19 +349,16 @@ public class LocalRepresentationSelectionPage extends WizardPage {
 
     setErrorMessage(null);
 
-    String warningMessage = findAndReportClashingProjectArtifacts();
+    String warningMessage;
 
     if (!unsupportedCharsets.isEmpty()) {
-      if (warningMessage == null) {
-        warningMessage = "";
-      } else {
-        warningMessage += "\n";
-      }
-
-      warningMessage +=
+      warningMessage =
           MessageFormat.format(
               Messages.LocalRepresentationSelectionPage_warning_unsupported_encoding_found,
               StringUtils.join(unsupportedCharsets, ", "));
+
+    } else {
+      warningMessage = null;
     }
 
     setMessage(warningMessage, WARNING);
@@ -430,53 +423,6 @@ public class LocalRepresentationSelectionPage extends WizardPage {
     } else {
       currentErrors.remove(referencePointID);
     }
-  }
-
-  /**
-   * Scans the current Eclipse Workspace for existing project artifacts that would clash with the
-   * projects created as part of the resource negotiation with the current selections.
-   *
-   * @return a warning message if such artifacts are found, or <code>null</code> otherwise
-   */
-  private String findAndReportClashingProjectArtifacts() {
-    IPath workspacePath = ResourcesPlugin.getWorkspace().getRoot().getLocation();
-
-    if (workspacePath == null) {
-      return null;
-    }
-
-    File workspaceDirectory = workspacePath.toFile();
-
-    List<String> dirtyProjectNames = new ArrayList<>();
-
-    for (ReferencePointOptionComposite composite : referencePointOptionComposites.values()) {
-      ReferencePointOptionResult referencePointOptionResult = composite.getResult();
-
-      if (referencePointOptionResult.getLocalRepresentationOption() != NEW_PROJECT) {
-        continue;
-      }
-
-      String projectName = referencePointOptionResult.getNewProjectName();
-
-      if (projectName == null || projectName.isEmpty()) {
-        continue;
-      }
-
-      if (new File(workspaceDirectory, projectName).exists()) {
-        dirtyProjectNames.add(projectName);
-      }
-    }
-
-    String warningMessage = null;
-
-    if (!dirtyProjectNames.isEmpty()) {
-      warningMessage =
-          MessageFormat.format(
-              Messages.LocalRepresentationSelectionPage_warning_project_artifacts_found,
-              StringUtils.join(dirtyProjectNames, ", "));
-    }
-
-    return warningMessage;
   }
 
   /**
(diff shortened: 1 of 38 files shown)


eclipse/src/saros/ui/wizards/pages/LocalRepresentationSelectionPage.java
```diff
@@ -219,6 +220,18 @@ private void showNewTabErrorMessage(SelectionEvent e) {
...
 
+    Button newProjectButton = new Button(composite, SWT.PUSH);
+    newProjectButton.setImage(ImageManager.ETOOL_NEW_PROJECT);
+    newProjectButton.setLayoutData(new GridData(SWT.END, SWT.FILL, false, false));
+    newProjectButton.setText(Messages.LocalRepresentationSelectionPage_new_project_button);
+    newProjectButton.addSelectionListener(
+        new SelectionAdapter() {
+          @Override
+          public void widgetSelected(SelectionEvent e) {
+            WizardUtils.openNewProjectWizard();
```
srossbach, line 231: Can you apply a TODO to correctly configure the project (project nature etc., of course nothing of the required data is currently transmitted) ?

stf/src/saros/stf/shared/Constants.java
```diff
@@ -10,19 +10,9 @@
     CONCURRENTLY
```
srossbach, line 10: If the _enum_ is not used on the server part it should not be placed here.

tobous: But these constants aren't STF client/server-specific, are they? The enum is basically only used in combination with `TypeOfCreateProject`, so it only seemed natural to define it in the same class. Where else should I put it?

tobous:

tobous:

stf/src/saros/stf/shared/Constants.java
srossbach: Guess it should be moved to the client. But this also includes many other stuff and so this is not the best idea to do this in this patch.

eclipse/src/saros/ui/wizards/pages/LocalRepresentationSelectionPage.java
srossbach: It is not that easy to change the project nature afterwards. Most of the times you have to convert the project to a facetted project which is not the same as, e.g a pure Java project.

srossbach (approved):

eclipse/src/saros/ui/wizards/pages/LocalRepresentationSelectionPage.java
tobous: > It is not that easy to change the project nature afterwards. Most of the times you have to convert the project to a facetted project which is not the same as, e.g a pure Java project.

I am not that familiar with Eclipse project natures, so I can't speak to that. But I would assume that the default java project configuration (or any other configuration that can be created through the normal Eclipse project creation wizard) is usable in most cases (even if it reduces the available feature set).

Either way, this is not the place to put a TODO for this in my opinion. When creating the project, we have no idea which shared reference point it will actually be used for. So the logic would have to be added to the resource negotiation. The only option to even potentially offer correctly setting up the project is to send the necessary data as additional resource negotiation data in my opinion. With any other approach, we would be adding IDE-specific logic/concepts to core logic.
(Additionally, changing the project configuration might be confusing to the user in cases where only inner resources (i.e. not a project root) are shared. In such a setup, adjusting the project configuration would modify resources that are not shared.)

But I don't think that this should be part of this patch. If you are interested in pursuing such an approach (or want to create a reminder for a later date), I would suggest opening a new issue for it.

PR #1097, opened 2020-11-05
[NOP][CORE] Remove the ROOT resource type

tobous:
Removes the ROOT resource type as it is no longer being used.


tobous:
Forgot to remove it as part of the rework.

PR #1104, opened 2020-11-16
Check for qualifier as part of version compatibility check

tobous:
Adjusts the Saros version compatibility logic. Adds a separate check ensuring that the entire version string matches (including micro and qualifier) when a qualifier is given as part of the version. This ensures that users can't start a session between different development builds. Subsequently cleans up the version compatibility logic.

Adds a specific page describing the Saros version compatibility to the homepage. Also adds a section mentioning the development builds to the releases page.

### Commits

<details><summary><b>[INTERNAL][CORE] Add compatibility result for differing dev builds</b></summary>
<br>

Adds a compatibility result indicating that at least one of the compared
versions contained a qualifier and the version strings did not match
exactly. As the qualifier is the indicator that a development build is
used, this result can be used to ensures that only matching development
builds are compatible.

</details>

<details><summary><b>[INTERNAL][CORE] Include qualifier in compatibility check</b></summary>
<br>

Adds a separate case to the compatibility check, ensuring that the
entire version string matches in case one of the compared versions
contains a qualifier. This check ensures that clients running a
development build can only start a session with clients running the same
development build. This should avoid issues where users download
different, potentially incompatible development builds and start a
session with them.

Simplifies the logic comparing two versions.

Adjusts Version.equals(...) to include the micro version and qualifier
in the check.

Extends VersionManagerTest to cover all possible comparison cases.

</details>

<details><summary><b>[NOP][CORE] Remove unused capabilities of Compatibility enum</b></summary>
<br>

Removes the additional capabilities of the Compatibility enum (like
specifying the integer represented by an entry and inverting an entry).
They were no longer being used.

</details>

<b>[REFACTOR][CORE] Clean up version compatibility logic</b>

<details><summary><b>[DOC] Add page for Saros version compatibility</b></summary>
<br>

Adds a dedicated page to the user documentation explaining the Saros
version compatibility. Also explains the special case when using a Saros
version containing the 'qualifier' element (e.g. development builds).

Adds the new compatibility page to the user documentation sidebar. As
this is kind of fringe knowledge, I added as the second-to-last entry
(only followed by the FAQ). Adding this information to the FAQ instead
of a separate section seemed too inaccessible/hard to find to me.

Adds a section on development builds to the release page. Explicitly
mentions that such builds may be unstable.

</details>

### DIFF
--- a/core/src/saros/versioning/VersionManager.java
+++ b/core/src/saros/versioning/VersionManager.java
@@ -55,17 +55,7 @@ public class VersionManager {
       log.warn("contact: " + contact + ", remote version string is invalid: " + versionString);
     }
 
-    Compatibility compatibility = determineCompatibility(localVersion, remoteVersion);
+    Compatibility compatibility = localVersion.determineCompatibilityWith(remoteVersion);
     return new VersionCompatibilityResult(compatibility, localVersion, remoteVersion);
-  }
-
-  /**
-   * Compares the two given versions for compatibility. The result indicates whether the local
-   * version is compatible with the remote version.
-   */
-  private Compatibility determineCompatibility(Version localVersion, Version remoteVersion) {
-    Compatibility compatibility = Compatibility.valueOf(localVersion.compareTo(remoteVersion));
-
-    return compatibility;
   }
 }
--- a/core/src/saros/versioning/VersionManager.java
+++ b/core/src/saros/versioning/VersionManager.java
@@ -10,7 +10,7 @@ import saros.net.xmpp.contact.XMPPContact;
 /**
  * Component for figuring out whether two Saros plug-in instances with known Version are compatible.
  *
- * <p>This class compares if local and remote version (not checking qualifier) are the same.
+ * @see Version#determineCompatibilityWith(Version)
  */
 @Component(module = "core")
 public class VersionManager {
@@ -28,8 +28,10 @@ public class VersionManager {
 
   public VersionManager(@SarosVersion String version, InfoManager infoManager) {
     this.localVersion = Version.parseVersion(version);
-    if (this.localVersion == Version.INVALID)
+
+    if (this.localVersion == Version.INVALID) {
       throw new IllegalArgumentException("version string is malformed: " + version);
+    }
 
     this.infoManager = infoManager;
     infoManager.setLocalInfo(VERSION_KEY, localVersion.toString());
@@ -45,12 +47,14 @@ public class VersionManager {
     Objects.requireNonNull(contact, "contact is needed");
 
     String versionString = infoManager.getRemoteInfo(contact, VERSION_KEY).orElse(null);
+
     if (versionString == null) {
       log.warn("remote version not found for: " + contact);
       return new VersionCompatibilityResult(Compatibility.UNKNOWN, localVersion, Version.INVALID);
     }
 
     Version remoteVersion = Version.parseVersion(versionString);
+
     if (remoteVersion == Version.INVALID) {
       log.warn("contact: " + contact + ", remote version string is invalid: " + versionString);
     }
--- a/core/src/saros/versioning/Compatibility.java
+++ b/core/src/saros/versioning/Compatibility.java
@@ -9,7 +9,7 @@ public enum Compatibility {
   /**
    * The local version is (probably) too old to work with the remote version.
    *
-   * <p>The user should be told to upgrade
+   * <p>The user should be told to upgrade.
    */
   OLDER,
 
@@ -20,12 +20,12 @@ public enum Compatibility {
    */
   NEWER,
 
-  /** The compatibility could not be determined. */
-  UNKNOWN,
-
   /**
-   * At least one of the versions contains a qualifiers and the complete version string does not
+   * At least one of the versions contains a qualifier and the complete version string does not
    * match.
    */
-  QUALIFIER_MISMATCH
+  QUALIFIER_MISMATCH,
+
+  /** The compatibility could not be determined. */
+  UNKNOWN
 }
--- a/core/src/saros/versioning/Compatibility.java
+++ b/core/src/saros/versioning/Compatibility.java
@@ -4,76 +4,28 @@ package saros.versioning;
 public enum Compatibility {
 
   /** Versions are (probably) compatible */
-  OK(0) {
-    @Override
-    public Compatibility invert() {
-      return OK;
-    }
-  },
+  OK,
+
   /**
    * The local version is (probably) too old to work with the remote version.
    *
    * <p>The user should be told to upgrade
    */
-  OLDER(1) {
-    @Override
-    public Compatibility invert() {
-      return NEWER;
-    }
-  },
+  OLDER,
+
   /**
    * The local version is (probably) too new to work with the remote version.
    *
    * <p>The user should be told to tell the peer to update.
    */
-  NEWER(2) {
-    @Override
-    public Compatibility invert() {
-      return OLDER;
-    }
-  },
+  NEWER,
 
   /** The compatibility could not be determined. */
-  UNKNOWN(3) {
-    @Override
-    public Compatibility invert() {
-      return UNKNOWN;
-    }
-  },
+  UNKNOWN,
 
   /**
    * At least one of the versions contains a qualifiers and the complete version string does not
    * match.
    */
-  QUALIFIER_MISMATCH(4) {
-    @Override
-    public Compatibility invert() {
-      return QUALIFIER_MISMATCH;
-    }
-  };
-
-  private final int code;
-
-  Compatibility(final int code) {
-    this.code = code;
-  }
-
-  /**
-   * @return <code>TOO_OLD</code> if the initial compatibility was <code>TOO_NEW</code>, <code>
-   *     TOO_NEW</code> if the initial compatibility was <code>TOO_OLD</code>, <code>OK</code>
-   *     otherwise
-   */
-  public abstract Compatibility invert();
-
-  public int getCode() {
-    return code;
-  }
-
-  public static Compatibility fromCode(int code) {
-    for (Compatibility compatibility : Compatibility.values()) {
-      if (compatibility.getCode() == code) return compatibility;
-    }
-
-    return UNKNOWN;
-  }
+  QUALIFIER_MISMATCH
 }
--- a/core/src/saros/versioning/Compatibility.java
+++ b/core/src/saros/versioning/Compatibility.java
@@ -41,6 +41,17 @@ public enum Compatibility {
     public Compatibility invert() {
       return UNKNOWN;
     }
+  },
+
+  /**
+   * At least one of the versions contains a qualifiers and the complete version string does not
+   * match.
+   */
+  QUALIFIER_MISMATCH(4) {
+    @Override
+    public Compatibility invert() {
+      return QUALIFIER_MISMATCH;
+    }
   };
 
   private final int code;
--- a/core/src/saros/versioning/Compatibility.java
+++ b/core/src/saros/versioning/Compatibility.java
@@ -1,7 +1,5 @@
 package saros.versioning;
 
-import java.util.Comparator;
-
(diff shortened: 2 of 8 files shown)


srossbach:
The change roughly requires 2 additional lines of code. Why did you add a DEV variable into the core ? It is not used anywhere beside in the test cases. What is wrong of returning OLDER or NEWER instead ?

tobous:
> The change roughly requires 2 additional lines of code. Why did you add a DEV variable into the core ? It is not used anywhere beside in the test cases. What is wrong of returning OLDER or NEWER instead ?

Well, none of the values are used outside of the test case. The specific logic displaying error messages according to the exact version mismatch result was removed a while ago. I added it as a separate option in case we want to better differentiate between the different cases in the future.

More importantly, newer/older can't be used to describe the version difference in such cases. If one version is `1.0.0.abc` and the other is `1.0.0.def`, asking which one is older/newer does not really make sense. But I want such versions to be incompatible. If a qualifier is given, I want the whole version string to have to match.

I added this to avoid users mixing development builds (i.e. the builds release as part of the [GithHub workflow on the master](https://github.com/saros-project/saros/actions?query=workflow%3ABuild+branch%3Amaster+is%3Asuccess+event%3Apush)). It already happened that a used mixed different builds (or a development build with an official release) and reported an issue cause by this. In general, development builds are a snapshot of the current top commit on the master at that point, meaning small incompatibilities can arise frequently during the development process.

This new kind of incompatibility is also why I changed the version comparison logic. It breaks the approach solely relying on the integer comparison, meaning using `compareTo(...)` was no longer an option.

srossbach:
I really do not know if you are changing the plugin version via tooling but looking at the master and latest release tag the MANIFEST.MF contains the wrong version number.

For the current master it is still 16.0.0 which should be corrected to either a version number that will never be seen in production or to the next expected version that should be released. Saros/E 16.0.1 internal version is also 16.0.0 .

So what do you expect to fix when someone checkout the current master and build the plugin by himself ?

tobous:
> I really do not know if you are changing the plugin version via tooling but looking at the master and latest release tag the MANIFEST.MF contains the wrong version number.
> 
> For the current master it is still 16.0.0 which should be corrected to either a version number that will never be seen in production or to the next expected version that should be released. Saros/E 16.0.1 internal version is also 16.0.0 .

I did not create the release `16.0.1`, but, if I remember correctly, it was created shortly after `16.0.0` to add a quick bugfix. Somebody probably forgot to adjust the version number. While this is not optimal, it should not really matter that much in this case as the micro identifier is ignored for the version check anyway.

We should at least adjust it to `16.0.1`. (The current convention is that the master remains on the last release version until a new release is being prepared. But it might be sensible to change that.)

> So what do you expect to fix when someone checkout the current master and build the plugin by himself ?

This change is mainly aimed at people using our prebuilt development builds. When people are building the plugin themselves, there isn't really anything we can do about compatibility as they can change the version number to whatever they want anyway. We technically have the capabilities of adding a qualifier to all local builds by default. But, if we were to change our mind, we can easily add a constant dev qualifier by adding it as the `versionQualifier` in the [`gradle.properties`](https://github.com/saros-project/saros/blob/master/gradle.properties) file. (We would then have to make sure to overwrite this parameter for actual release builds.)

tobous:
Added `dev` as the default qualifier for local Gradle builds. This should somewhat alleviate the problem. I would have liked to use the same format the development build does (i.e. use the commit hash of the current commit, but this would rely on a local Git installation which does not necessarily have to be available; furthermore, the local state might be dirty, so it wouldn't be that helpful anyway).

srossbach (changes requested):
The version class is a POJO.

The compatibility should not be determined in this class but in the version manager.

Also introducing some new flags for compatibility results which will be replaced somewhere in the near future is not very helpful for now.

If your only intention is to fix "non release builds" for now just return UNKNOWN.

This will still not fix the problem for now, as you can just switch the inviter side.

tobous:
Also, I found out why we did not set a `dev` qualifier by default. It causes the version including the qualifier to be written to the Eclipse manifest files on build. Furthermore, once these qualifiers have been written to the file, removing the `versionQualifier` property does not remove the qualifier from the manifest files. In its current form, the logic is only supposed to be used for CI builds. Will remove the commit.

srossbach:
> 
> 
> Also, I found out why we did not set a `dev` qualifier by default. It causes the version including the qualifier to be written to the Eclipse manifest files on build. Furthermore, once these qualifiers have been written to the file, removing the `versionQualifier` property does not remove the qualifier from the manifest files. In its current form, the logic is only supposed to be used for CI builds (see discussion in #824). Will remove the commit.

I still do not understand the problem. Our jars are not signed (see issue tracker). You can simply extract the MANIFEST.MF file, do a simple `sed` to change the version of the plugin and insert the modified file into the jar file.

tobous:
> I still do not understand the problem. Our jars are not signed (see issue tracker). You can simply extract the MANIFEST.MF file, do a simple sed to change the version of the plugin and insert the modified file into the jar file.

This is not aimed at people actively trying to change the plugin to make whatever they are trying to do work. As you correctly pointed out before, people can always build the plugin themselves and do whatever they want. That is not what I am trying to prevent. (The default `dev` qualifier just would have made it easier to spot when a user reported something based on a self-built version.)

I explicitly added this case to avoid people downloading and using Saros developer builds in combination with different developer builds or official releases and then report bugs based on a session running on this invalid setup.

srossbach:
But you still trying to solve this issue in the Core although it is related to IDE specific details (version numbers).

The field DEV_XYZ stuff does not fit. This is NOT the version of  the (a development) core but the version of the IDE using the specific Core Version.

Of course you can chance the logic to: if a qualifier is present and the version match but the qualifier mismatch then the version is not compatible, but do you know the real reason ? Imagine somewhere in the future person X thinks it is a good behavior to add the qualifier (for what reason ever) to the Saros/E plugin. And now the check fails.

If you really want to apply additional checks then please do not hard code them in the Core. Let the plugin decide how to deal with additional data, i.e if the core decides if 2 versions are equals then let Saros/I or Saros/E do additional checks if they are not compatible at all.

Also listing up all the reasons in the enum is quite confusing.

We had OK, OLDER, NEWER

Now we have OK, OLDER, NEWER, DEV_XYZ

What about OK, OLDER, NEWER, DEV_XYZ and FILE_NOT_FOUND ?

To be more specific, the comparison result is binary, either it is compatible or not. Everything else are details but the current enum does not reflect that very well (mainly the remaining options are the reason why there is no compatibility)

srossbach:
And by the way, if we would had a person that would decide the release schedule plan (e.g we are now working on 16.0.1) and decide that the next release would be 16.1.0 or 17.0.0 this patch would be not necessary at all. But I guess this is pretty hard to accomplish in this project.

tobous:
> But you still trying to solve this issue in the Core although it is related to IDE specific details (version numbers).
> 
> The field DEV_XYZ stuff does not fit. This is NOT the version of the (a development) core but the version of the IDE using the specific Core Version.
>
>  [...]
>
> If you really want to apply additional checks then please do not hard code them in the Core. Let the plugin decide how to deal with additional data, i.e if the core decides if 2 versions are equals then let Saros/I or Saros/E do additional checks if they are not compatible at all.

With the current implementation, the compatibility is solely determined through logic in the core that is shared by all IDE implementations, meaning it is enforced for all implementations. As part of this logic, we have currently defined a specific set of rules that determine the compatibility:
The Saros version has the components `MAJOR.MINOR.MICRO.QUALIFIER` where `QUALIFIER` is the only optional field. For two Saros versions to be compatible, the `MAJOR` and `MINOR` version number has to match. The `MICRO` and  `QUALIFIER` values are ignored. Additionally, as the `QUALIFIER` is a String rather than an integer, it was mainly used to convey additional information (like the development status).

I don't see why this handling does anything else. With this handling, I explicitly specify that the `QUALIFIER` is only to be used for development builds and introduce a new version compatibility check acting accordingly. I also don't see what this has to do with the IDE-specific implementations. This is not some sort of "additional data" but rather just a part of the version number, whose usage is specified though the logic in the core.

Also, adding IDE-implementation-specific compatibility logic only worsens the issue with the asymmetric version check, potentially creating even more cases where it matter who invites whom. And, if we are still aiming for IDE cross-compatibility, splitting the compatibility logic across different IDE implementations will make compatibility very cumbersome as the implementations will inevitably drift apart.

> Of course you can chance the logic to: if a qualifier is present and the version match but the qualifier mismatch then the version is not compatible, but do you know the real reason ?

What do you mean with "the real reason"? The real reason I want to implement this restriction? (If so, see my previous messages. Development builds.)

> Imagine somewhere in the future person X thinks it is a good behavior to add the qualifier (for what reason ever) to the Saros/E plugin. And now the check fails.

If in the future a developer decides to include the `QUALIFIER` element as part of an official release, they would break this rule and would have to deal with the ramification (either live with the missing compatibility or change the logic).

> Also listing up all the reasons in the enum is quite confusing.
> 
> We had OK, OLDER, NEWER
> 
> Now we have OK, OLDER, NEWER, DEV_XYZ
> 
> What about OK, OLDER, NEWER, DEV_XYZ and FILE_NOT_FOUND ?
> 
> To be more specific, the comparison result is binary, either it is compatible or not. Everything else are details but the current enum does not reflect that very well (mainly the remaining options are the reason why there is no compatibility)

That is what the documentation is for. The enum in general and every case in particular has javadoc that describes its purpose/meaning. If you wanted to break this down to just the binary distinction of "compatible" and "not compatible", you wouldn't need the enum at all. The enum is used to convey additional information about the compatibility. And while this information currently isn't used, I don't see an issue with keeping it around. This makes implementing additional logic dealing with different kinds of incompatibilities a lot easier.

> And by the way, if we would had a person that would decide the release schedule plan (e.g we are now working on 16.0.1) and decide that the next release would be 16.1.0 or 17.0.0 this patch would be not necessary at all. But I guess this is pretty hard to accomplish in this project.

I disagree. Even if we had a specific release plan, this issue would still apply. We are still releasing development builds with every commit pushed to the master (e.g. here are the [development builds for the current top commit on the master](https://github.com/saros-project/saros/actions/runs/358729825)). And that won't change with the release frequency. Even if we were to immediately increase the version after each release, meaning the development builds would no longer be compatible with the previous official releases, this would still not cover the compatibility of different development releases.

tobous:
And regarding the version number for the current Eclipse release. The released jar contains the correct version, so my guess would be that the version change was never merged back into the master. I will adjust it in a different PR along with the Saros Server version.

srossbach:
Can we at least agree to rename `MISMATCHED_DEV_BUILD` to `QUALIFIER_MISMATCH` ? This would also fit the description of the field.

tobous:
Done. I renamed the entry to `QUALIFIER_MISMATCH`. (Sorry, forgot to rebase separately beforehand, so the force-push diff also contains the other changes pushed to the master in the meantime.)

core/src/saros/versioning/VersionManager.java
```diff
@@ -45,27 +47,19 @@ public VersionCompatibilityResult determineVersionCompatibility(XMPPContact cont
...
     }
 
     Version remoteVersion = Version.parseVersion(versionString);
+
     if (remoteVersion == Version.INVALID) {
       log.warn("contact: " + contact + ", remote version string is invalid: " + versionString);
     }
 
-    Compatibility compatibility = determineCompatibility(localVersion, remoteVersion);
+    Compatibility compatibility = localVersion.determineCompatibilityWith(remoteVersion);
```
srossbach, line 62: The result contains 2 version objects that now includes a `determineCompatibilityWith` method. I do not know if this is really a good design choice.

tobous: I can make the method package-protected. That would keep other classes from accessing it, restricting the access to basically only `VersionManager`.

tobous:

tobous:
Set `Version.determineCompatibilityWith(Version)` to package-private.

srossbach (approved):

PR #1112, opened 2020-12-14
Revert "[VSC] Initialize vscode plugin"

srossbach:
Reverts saros-project/saros#1111


Drakulix:
Sorry for the premature merge of #1111.
I guess we can skip the review process for this, as it is just a plain revert. Any objections?

srossbach:
Seems reopening the original pull request is not that trivial. I do not want to add a revert of the revert. Please stop merging stuff next time while there is a still a discussion pending. Of course we could request changes but the discussion was more about a "Machbarkeitsstudie" rather than regarding the changes made in the patch.

Drakulix:
> Seems reopening the original pull request is not that trivial. I do not want to add a revert of the revert. Please stop merging stuff next time while there is a still a discussion pending. Of course we could request changes but the discussion was more about a "Machbarkeitsstudie" rather than regarding the changes made in the patch.

I genuinely overlooked the discussion as I meant to merge this days ago (before the discussion), but just forgot. Today I remembered, everything was still green and I simply did not notice the new comments. I would suggest next time, if there are such big issues with the general concept, a negative review would still be helpful. Still I should have catch'ed up and definitely not merged it. Sorry about that.

tobous:
While the PR was merged prematurely, I don't think it is necessary to revert it. It only adds basic build scripting and documentation. None of these are affected by the discussed issues. And in case the discussed issues prove to be insurmountable, everything relating to the VSCode implementation (including the LSP stuff) will probably be deleted anyways, meaning just reverting this part would have been unnecessary.

Issue #1114, opened 2021-01-07
Upgrade to Java 11?


Drakulix:
Saros/E fails to build with JDK 8 as seen by [our pipeline](https://github.com/saros-project/saros/runs/1662644456?check_suite_focus=true#step:5:1080).

```
 > Task :saros.eclipse:compileJava FAILED
import org.eclipse.ui.plugin.AbstractUIPlugin;
                            ^
  bad class file: /home/runner/.gradle/caches/modules-2/files-2.1/org.eclipse.platform/org.eclipse.ui.workbench/3.122.0/60c038f66a106d3d437e66d6109bb42657eec6b4/org.eclipse.ui.workbench-3.122.0.jar(org/eclipse/ui/plugin/AbstractUIPlugin.class)
    class file has wrong version 55.0, should be 52.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.
Note: Some input files use or override a deprecated API.
```

1. Clone saros
2. If necessary clean you build folder / global gradle cache
3. Run `./gradlew prepareEclipse sarosEclipse`

**Environment:**
 - OS: Arch Linux
 - IDE: (What gradle downloads for compiling)
 - Saros Version: master

**Additional context**
I suspect a version / dependency mismatch of some sort. The pipeline did build with the same commit days ago, some students 
first noticed the issue and informed me about it.

Building with Java 11 (as indicated by `class file has wrong version 55.0, should be 52.0`) fixes the problem, maybe the downloaded eclipse platform now defaults to a new version? So far I was not able to pin down the problem.

tobous:
Uhhh, Eclipse is now also finally building with Java 11 since release 4.17 (2020-09). See [wiki](https://wiki.eclipse.org/Eclipse/Installation#Eclipse_4.18_.282020-12.29). So we can revive the discussion on whether we want to upgrade to Java 11 or not.

Updating to Java 11 would potentially reduce the usability of Saros as users not on the newest release would have to manually configure Eclipse to run using Java 11. But this is still worth it in my opinion, just to keep up with the times, now that we finally have the opportunity.

Migrating the GitHub Actions and documentation is no problem, I could create a PR for that.
[...]

PR #1115, opened 2021-01-08
Workaround for compiling Saros/E with Java 8

Drakulix:
This is a temporary fix for #1114, which can be used to compile Saros/E with JDK version 8.

This is not a proposed solution and the workaround will need to be updated for every `org.eclipse.platform`, that we use and gets an update at this point. Actually discussion about long-term fixes should be done in #1114.

### DIFF
--- a/eclipse/build.gradle.kts
+++ b/eclipse/build.gradle.kts
@@ -42,6 +42,9 @@ dependencies {
     implementation(project(":saros.core"))
     // This is a workaround for https://github.com/saros-project/saros/issues/1086
     implementation("org.eclipse.platform:org.eclipse.urischeme:1.1.0")
+    // This is a workaround for https://github.com/saros-project/saros/issues/1114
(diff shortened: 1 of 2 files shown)


m273d15:
This change is sufficient for the build process, but I think you to add this workaround also to the [stf build file](https://github.com/saros-project/saros/blob/master/stf/build.gradle.kts). Otherwise, the stf testing pipeline will fail

m273d15 (approved):
Sweeeet 🍬

tobous:
Since the proper resolution of #1114 will probably take a while, I don't see an issue with actually merging this PR to fix the master in the meantime. While we should not take it as a reason to delay resolving the underlying issue, having a buildable master makes working on the project easier, especially for newcomers interested in the project.

And even if the solution is implemented faster than expected, reverting the workaround is easy enough, especially since we already have the existing workaround for the previous issue that will have to be removed as well.

tobous (approved):

PR #1118, opened 2021-01-25
[BUILD] Extract dependency export from core into seperate bundle

FKHals:
by creating a new bundle called "saros.libratory" which serves as a
dependency-container and replaces the previously used Core-bundle ("saros.core")
in this role.

This was done to reduce unnecessary coupling between the Core- and the Eclipse-bundle
caused by the Core-bundle's dependency exports.
Now the Core-bundle serves exclusively as a central library for a Saros-
implementation without also serving as a dependency container.
Since other packages (Intellij, Server,..) also used the Core's dependencies
those now get it's dependencies from the new Libratory-bundle too.
The following bundle-dependency-relations have been identified and get exported
by the Libratory-bundle:
(scheme: [bundle(s)]: * dependency of that bundles)

[core, eclipse, stf, sever, intellij]:
* log4j-1.2-api-2.13.3.jar
* log4j-api-2.13.3.jar
* log4j-core-2.13.3.jar
* picocontainer-2.11.2-patched_relocated.jar

[core, eclipse]:
* weupnp.jar
* smackx-3.4.1.jar

[core, eclipse, stf, sever]:
* smack-3.4.1.jar

The motivation is also described [here](https://git.imp.fu-berlin.de/vbrekenfeld/swp-large-codebase/-/wikis/Themensammlung#core-als-dependency-container).


srossbach:
Fine and who ensure the correct versioning ?

FKHals:
> Fine and who ensure the correct versioning ?

The correct versioning of what exactly? It would be great if you could go into more detail about your question.

Nothing has changed regarding the versioning of the used packages. What has changed are the locations of the jar files and the bundles that export them.

Drakulix:
To me this looks like the proposed split between implementation and library-exports.
While we could bike shed over the Libratory name, I see nothing fundamentally wrong with this PR.

But I do not want to overshadow anyone's opinion, so @srossbach could you please elaborate?
Also @m273d15 could I ask you providing a quick comment regarding, if this does what you had in mind?
After all you were the initial author of the linked proposal.

m273d15:
@FKHals  Hi, great to see that someone is working on the build and dependency stuff 👍 

I think my main idea was that the core is a just library that is used by the Saros implementations and the Saros implementations (eclipse|intellij|vscode) relies on the transitive dependencies of the Core.
This might be useful if we are talking about dependencies that are defining the communication protocol between Core and implementation. However, it is still implicit and we should make the dependency explicit.
If we are talking about utility libraries (for example `org.apache.commons.lang3.time`) it makes no sense. I would not use the tuple implementation of my logging library, why should I use the tuple implementation of the Core?
-- I hope my initial german explanation was comprehensible ... I needed a few minutes to understand it 😅 

I will review and think about the PR as soon as possible.

tobous:
This breaks the STF, at least our automated setup. See [STF run](https://github.com/saros-project/saros/pull/1118/checks?check_run_id=1784300540) and [build scan](https://scans.gradle.com/s/3n7cjeikvepf2/tests) for more details.

Also, if possible, I would like to get rid of the `plain` core configuration as part of this change.

It was initially introduced to avoid duplicating libraries (see issue #830 and PR #835). The issue was that IntelliJ and Eclipse plugins handle dependencies differently. For Eclipse, the dependencies are included in the jar. For Intellij, they are packaged as part of the main plugin. This meant that for Saros/I, the core libraries were included twice. Once in the main plugin lib directory and once as part of the `Core` jar. To avoid this issue, a `plain` configuration was introduced for the `core` which does not include the libraries.

But introducing the plain jar also had the side-effect of breaking/deteriorating the IntelliJ code analysis. The `core` jar and the `core-plain` jar are seen as separate entities by IntelliJ (as far as I can tell), meaning the code analysis (e.g. showing the usage of a method) only ever shows the results for one of the two jars. This makes refactoring the `Core` much more cumbersome as it is much harder to find the usages of the exported `Core` interfaces for all Saros implementations.

@m273d15 Any comment on the feasibility of this request? If it is not feasible as part of this change, can you think of another way to avoid this problem? I find it quite bothersome.

m273d15:
@FKHals It seems like you are using the CI for executing the STF tests. You can also execute them on your local Linux machine with `./run_stf.sh` (needs docker and the corresponding docker group privileges)

stf.test/META-INF/MANIFEST.MF
```diff
@@ -7,6 +7,7 @@ Bundle-Vendor: Saros Project
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Require-Bundle: saros.eclipse,
  saros.core,
+ saros.libratory,
```
srossbach, line 10: This should not be necessary. The core already includes and exports them. The worst thing that can happen in an OSGi context is that this plugin will see a different "set" of the libraries than the core. i.e the code is loaded twice into the VM and every static variable will be seen different depending which class loader loaded which class.

FKHals:
> @FKHals It seems like you are using the CI for executing the STF tests. You can also execute them on your local Linux machine with `./run_stf.sh` (needs docker and the corresponding docker group privileges)

I'm actually aware of that but i am currently using a Fedora which does not support docker (because of cgroups v2) and i am not sure what would break when i change that so using the CI and its produced artifacts seemed good enough for now.

tobous:
> > @FKHals It seems like you are using the CI for executing the STF tests. You can also execute them on your local Linux machine with `./run_stf.sh` (needs docker and the corresponding docker group privileges)
> 
> I'm actually aware of that but i am currently using a Fedora which does not support docker (because of cgroups v2) and i am not sure what would break when i change that so using the CI and its produced artifacts seemed good enough for now.

If your comments regard the missing `cgroups v2` support of docker, this has been fixed with release [20.10.0](https://docs.docker.com/engine/release-notes/#runtime-1). Docker should be compatible with Fedora 32 and 33 (see [documentation](https://docs.docker.com/engine/install/fedora/)).

FKHals:
> If your comments regard the missing `cgroups v2` support of docker, this has been fixed with release [20.10.0](https://docs.docker.com/engine/release-notes/#runtime-1). Docker should be compatible with Fedora 32 and 33 (see [documentation](https://docs.docker.com/engine/install/fedora/)).

Ah thanks, i was not aware of that!

stf.test/META-INF/MANIFEST.MF
FKHals: But we did not include the saros.libratory package in the `Export-Package`-section of the core's MANIFEST.MF. Would it be exported anyway?
We thought it would be preferable (to reduce coupling) if the core would not export the libratory package transitively.

srossbach (commented):
> 
> 
> To me this looks like the proposed split between implementation and library-exports.
> While we could bike shed over the Libratory name, I see nothing fundamentally wrong with this PR.
> 
> But I do not want to overshadow anyone's opinion, so @srossbach could you please elaborate?

It is mainly an issue with the deployment. If you update the bundle you also have to ensure that Eclipse will install it (so basically update the version number along with a proper site configuration). Otherwise the worst thing that happens is that the newest Eclipse version will use an outdated bundle.

m273d15:
> It is mainly an issue with the deployment. If you update the bundle you also have to ensure that Eclipse will install it (so basically update the version number along with a proper site configuration). Otherwise the worst thing that happens is that the newest Eclipse version will use an outdated bundle.

@srossbach Thank you for mentioning the issue. I already forgot this problem.

I think this PR is an improvement in terms of separation, but adding an additional OSGi bundle  also increases the complexity of the dependency resolution.
Therefore, I would propose to solve this with the existing OSGi bundles and the Gradle infrastructure. 

The introduction of your libratory project has two effects:
1. You created a new OSGi bundle
2. You created a separation of the Gradle configurations of Core and libratory
and I think this is a good idea.

I think it is possible to implement the 2nd aspect without the 1th.

I created the `releaseDep` because I wanted to define a certain set of dependencies that is available during runtime if a project A and is not inherited by all projects that depend on A. This was before the [`java-library` plugin](https://docs.gradle.org/current/userguide/java_library_plugin.html) was released. This plugin provides different configurations that allow to define a more fine-grained inheritance of dependencies in multi-project gradle projects. The interesting configuration is `api` which allows you to define the dependencies that are inherited by other projects. 

I think you could (example: core):
- Create a configuration `bundle` (`implementation` should extend from `bundle`)
- Create a configuration `bundleApi` (`api` should extend from `bundleApi`)
- You modify the `JarConfigurator.java` to accept a list of configurations that are included in the bundle jar (instead of using `RELEASE_CONFIG_NAME`).
- Set `bundle` and `bundleApi` as these configurations

Why is this an improvement?
All dependencies defined with `bundle` are included in the jar but are not inherited by other projects (example: eclipse). Therefore, we do not use transitive dependencies of the core project. In contrast `bundleApi` dependencies are included in the jar AND are inherited by other projects. If we had a complete separation the configuration `bundleApi` would be useless.

I think if you completely understand the configuration inheritance in the java-library project you can re-work the configuration setup and clean the existing solution which was created before the release of java-library. Probably, you can also find a better solution than the `-plain` workaround.

If you like we can chat about it via Skype, Zoom, Teams, Two cups and a string or smoke signals.

PR #1119, opened 2021-02-01
[UI][E] Menu Migration from Eclipse3 to Eclipse4

antej95:
Migration of the Main-Menu, as well as all Commands and Handlers to the E4 Platform.

Explanation:
Since 2012 Saros/E uses the compatibility layer provided by Eclipse for the UI to function.
To drop the need for the compatibility layer all UI Elements have to be migrated and all references to org.eclipse.ui have to be removed.
Also described [here](https://git.imp.fu-berlin.de/vbrekenfeld/swp-large-codebase/-/wikis/Themensammlung#e4-migration).

### DIFF
--- a/eclipse/fragment.e4xmi
+++ b/eclipse/fragment.e4xmi
@@ -1,0 +1,73 @@
+<?xml version="1.0" encoding="ASCII"?>
+<fragment:ModelFragments xmi:version="2.0" xmlns:xmi="http://www.omg.org/XMI" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:application="http://www.eclipse.org/ui/2010/UIModel/application" xmlns:commands="http://www.eclipse.org/ui/2010/UIModel/application/commands" xmlns:fragment="http://www.eclipse.org/ui/2010/UIModel/fragment" xmlns:menu="http://www.eclipse.org/ui/2010/UIModel/application/ui/menu" xmlns:ui="http://www.eclipse.org/ui/2010/UIModel/application/ui" xmi:id="_9R0kQDYLEeuIF8EFjFT7JQ">
+  <fragments xsi:type="fragment:StringModelFragment" xmi:id="_AA2WUDYMEeuIF8EFjFT7JQ" featurename="commands" parentElementId="xpath:/">
+    <elements xsi:type="commands:Command" xmi:id="_FiO7QDYMEeuIF8EFjFT7JQ" elementId="saros.ui.commands.Empty" commandName="EMPTY COMMAND"/>
+    <elements xsi:type="commands:Command" xmi:id="_Ye-coDYNEeuIF8EFjFT7JQ" elementId="saros.ui.commands.GettingStarted" commandName="Getting Started..."/>
+    <elements xsi:type="commands:Command" xmi:id="_hvu2UDYNEeuIF8EFjFT7JQ" elementId="saros.ui.commands.StartSarosConfiguration" commandName="Start Saros Configuration..."/>
+    <elements xsi:type="commands:Command" xmi:id="_VbGKADYPEeuIF8EFjFT7JQ" elementId="saros.ui.commands.CreateXMPPAccount" commandName="Create Account..."/>
+    <elements xsi:type="commands:Command" xmi:id="_Y_2DIDYPEeuIF8EFjFT7JQ" elementId="saros.ui.commands.AddContact" commandName="Add Contact..."/>
+    <elements xsi:type="commands:Command" xmi:id="_h9fTkDYPEeuIF8EFjFT7JQ" elementId="saros.ui.commands.OpenSarosPreferences" commandName="Preferences..."/>
+    <elements xsi:type="commands:Command" xmi:id="_ji39ADYPEeuIF8EFjFT7JQ" elementId="saros.ui.commands.AddXMPPAccount" commandName="Add Account..."/>
+    <elements xsi:type="commands:Command" xmi:id="_pZa5oDYPEeuIF8EFjFT7JQ" elementId="saros.ui.commands.ShareResource" commandName="Share Resource(s)..."/>
+    <elements xsi:type="commands:Command" xmi:id="_rsWlkDYPEeuIF8EFjFT7JQ" elementId="saros.ui.commands.SessionLeave" commandName="Leave Session..."/>
+    <elements xsi:type="commands:Command" xmi:id="_wvkAwDYPEeuIF8EFjFT7JQ" elementId="saros.ui.commands.SessionAddResources" commandName="Add Resource(s)..."/>
+    <elements xsi:type="commands:Command" xmi:id="_0XH9UDYPEeuIF8EFjFT7JQ" elementId="saros.ui.commands.SessionAddContacts" commandName="Add Contacts..."/>
+    <elements xsi:type="commands:Command" xmi:id="_6kVtgDYPEeuIF8EFjFT7JQ" elementId="saros.ui.commands.SessionAddSelectedResources" commandName="Add to Saros Session"/>
+    <elements xsi:type="commands:Command" xmi:id="_8TldkDYPEeuIF8EFjFT7JQ" elementId="saros.ui.commands.SessionAddSelectedContacts" commandName="Add to Saros Session"/>
+  </fragments>
+  <fragments xsi:type="fragment:StringModelFragment" xmi:id="_B_wWMDYQEeuIF8EFjFT7JQ" featurename="handlers" parentElementId="xpath:/">
+    <elements xsi:type="commands:Handler" xmi:id="_Fft18DYQEeuIF8EFjFT7JQ" elementId="saros.eclipse.handler.0" contributionURI="bundleclass://saros.eclipse/saros.ui.command_handlers.GettingStartedHandler" command="_Ye-coDYNEeuIF8EFjFT7JQ"/>
+    <elements xsi:type="commands:Handler" xmi:id="_ymPZEDYQEeuIF8EFjFT7JQ" elementId="saros.eclipse.handler.1" contributionURI="bundleclass://saros.eclipse/saros.ui.command_handlers.StartSarosConfigurationHandler" command="_hvu2UDYNEeuIF8EFjFT7JQ"/>
+    <elements xsi:type="commands:Handler" xmi:id="_AIKtADYREeuIF8EFjFT7JQ" elementId="saros.eclipse.handler.2" contributionURI="bundleclass://saros.eclipse/saros.ui.command_handlers.CreateXMPPAccountHandler" command="_VbGKADYPEeuIF8EFjFT7JQ"/>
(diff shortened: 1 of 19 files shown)


eclipse/src/saros/ui/util/SWTUtils.java
```diff
@@ -187,7 +187,12 @@ public static void runSafeSWTSync(final Logger log, final Runnable runnable) {
    * @return the display of the current workbench
    */
   public static Display getDisplay() {
-    return PlatformUI.getWorkbench().getDisplay();
+    // return PlatformUI.getWorkbench().getDisplay();
```
srossbach, line 190: There is a reason we do not use` Display#getDefault`. 

When Eclipse shuts down it disposes the display. With this change you will creating a new one and this is something we do not want to happen.

eclipse/src/saros/ui/command_handlers/SessionAddSelectedResourcesHandler.java
```diff
@@ -31,4 +48,21 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
...
+  public boolean canExecute(
+      @Named(IServiceConstants.ACTIVE_SELECTION) @Optional IResource resource) {
+    if (!(connectionHandler != null
+        && connectionHandler.isConnected()
+        && sessionManager != null
+        && sessionManager.getSession() != null
+        && sessionManager.getSession().hasWriteAccess())) {
+      return false;
+    }
+    final ISarosSession session = sessionManager.getSession();
```
srossbach, line 62: The session could be null at this point.

srossbach (changes requested):

eclipse/src/saros/ui/command_handlers/SessionAddSelectedResourcesHandler.java
antej95: Is there a better way besides the null check in lines 57/58? 
The Code is adapted from the ResourcePropertyTester where it is done with a simple null check as well.

Drakulix: I think the problem is, that `getSession` might return a different object (possible null) on the second call, if the session is currently shutting down.

You need to first null-check the session manager, then call `getSession` and then null-check the session .

eclipse/fragment.e4xmi
```diff
@@ -0,0 +1,73 @@
...
+        <children xsi:type="menu:HandledMenuItem" xmi:id="_DNLZUDYVEeuIF8EFjFT7JQ" elementId="saros.eclipse.handledmenuitem.4" iconURI="platform:/plugin/saros.eclipse/icons/elcl16/session_add_contacts_tsk.png" command="_Y_2DIDYPEeuIF8EFjFT7JQ"/>
+        <children xsi:type="menu:MenuSeparator" xmi:id="_RkmK0DYVEeuIF8EFjFT7JQ" elementId="saros.eclipse.menuseparator.RosterEnd" accessibilityPhrase="RosterEnd"/>
+        <children xsi:type="menu:MenuSeparator" xmi:id="_2St4kDYVEeuIF8EFjFT7JQ" elementId="saros.eclipse.menuseparator.CollaborationStart" accessibilityPhrase="CollaborationStart"/>
+        <children xsi:type="menu:HandledMenuItem" xmi:id="_6984oDYVEeuIF8EFjFT7JQ" elementId="saros.eclipse.handledmenuitem.5" iconURI="platform:/plugin/saros.eclipse/icons/elcl16/session_tsk.png" command="_pZa5oDYPEeuIF8EFjFT7JQ">
+          <visibleWhen xsi:type="ui:CoreExpression" xmi:id="_N9R-gDYYEeuIF8EFjFT7JQ" coreExpressionId="saros.ui.definitions.notIsSarosSessionRunning"/>
```
srossbach, line 48: Isn't there a way to negate the coreExpressionId reference via xml ?

eclipse/src/saros/Saros.java
```diff
@@ -130,10 +135,17 @@ public void start(BundleContext context) throws Exception {
...
-    getWorkbench().addWorkbenchListener(workbenchShutdownListener);
     isInitialized = true;
   }
 
+  @Inject
+  @Optional
+  public void applicationStarted(
+      @UIEventTopic(UIEvents.UILifeCycle.APP_STARTUP_COMPLETE) Event event, Saros saros) {
+
+    saros.getWorkbench().addWorkbenchListener(saros.workbenchShutdownListener);
```
srossbach, line 146: I do not understand the benefit ?

eclipse/src/saros/ui/command_handlers/SessionAddSelectedContactsHandler.java
```diff
@@ -1,23 +1,49 @@
...
+  public Object execute() {
 
     List<JID> jids = SelectionRetrieverFactory.getSelectionRetriever(JID.class).getSelection();
 
     CollaborationUtils.addContactsToSession(jids);
     return null;
   }
+
+  @CanExecute
+  public boolean canExecute(@Named(IServiceConstants.ACTIVE_SELECTION) @Optional JID contact) {
```
srossbach, line 40: I do not think that this is the correct implementation. IIRC only the host can invite other contacts.

eclipse/src/saros/ui/command_handlers/ShareResourcesHandler.java
```diff
@@ -35,12 +38,27 @@
...
+
+  @Execute
+  public Object execute() {
     WizardUtils.openStartSessionWizard(
         SelectionRetrieverFactory.getSelectionRetriever(IResource.class).getSelection());
     return null;
   }
+
+  @CanExecute
+  public boolean canExecute() {
```
srossbach, line 58: I actually find this canExecute method really confusing. The visibleWhen checks are still in the plugin.xml file. It seems this is the enabledWhen check. However this method is different from SessionAddSelectedResourcesHandler class. So why is there no check if the resource is already shared ? And last but not least looking at the SessionAddSelectedResourcesHandler class, how does the method know how it has to convert the selections ?

srossbach (changes requested):

eclipse/src/saros/Saros.java
antej95: This Listener was previously added in Line 133.
But with the Addition of Eclipse 4 Fragments Saros is initialized before the Workbench and the Listener has to be added seperately after the Startup is complete and the Workbench exists.

eclipse/src/saros/ui/command_handlers/ShareResourcesHandler.java
antej95: The plugin.xml still contains the popup menus which have not yet been migrated, thats where the visibleWhen expression are from atm.
ShareResourceHandler and SessionAddSelectedResourcesHandler are fundamentally different, one starts a session the other one adds to an existing session.
I think youre right regarding the SessionAddSelectedResourcesHandler @CanExecute Method that it cant convert the Selection and the Button in the Popup Menu is not yet working properly.

eclipse/src/saros/Saros.java
srossbach: The listener is still removed in the stop method. This seems pretty confusing. Isn't there no new method that should be added ?

eclipse/src/saros/ui/command_handlers/SessionAddSelectedContactsHandler.java
srossbach: Nvm. The original code archive this by hiding the command instead of disabling the handler.

eclipse/src/saros/Saros.java
antej95: For us it seemed that the addition of E4 components moved the startup of saros up, before the initialization of a workbench.
Thats why we moved the addition of the workbench listener it to this extra function.
But the time whene the stop method is called didnt change so there was no need to move the remove call to an extra function.
We realized that there is some redundancy though, since the workbench listener was and still is called before the stop method, which both call stopLifeCycle() and we saw no change when removing the listener completely.
But we guessed that the listener serves a purpose, probably when the workbench isnt shutdown by ordinary means which would not stop all plugins

eclipse/src/saros/ui/command_handlers/SessionAddSelectedContactsHandler.java
antej95: I think was only ensured by the visibleWhen expression in the pop menu command and not in the handler.

https://github.com/saros-project/saros/blob/master/eclipse/plugin.xml#L664
And
https://github.com/saros-project/saros/blob/master/eclipse/plugin.xml#L1042

eclipse/src/saros/ui/command_handlers/ShareResourcesHandler.java
srossbach: I just looked at the old plugin.xml code and it seems this handler is a bit "misused" through the magic that happens on line 53 which makes it really hard to understand.

So is the design of e4 where the canExecute method gets the selection data but when the framework calls the execute method you have to query the selection service on your own ?

eclipse/src/saros/Saros.java
srossbach: https://github.com/saros-project/saros/commit/d602c26eef4185307d05270a803c7333357d9eab#diff-677f70b4646a5136f14fdb947bca53f9afedec5f58e09d5f9f201315fca8e4c3

eclipse/src/saros/ui/util/SWTUtils.java
srossbach: So after looking at the issue this is not the correct fix.

Either we ensure that the Platform is available when our context is started or we modify the affected classes.

eclipse/src/saros/Saros.java
srossbach: I subscribe to UIEvents.UILifeCycle.APP_SHUTDOWN_STARTED which seems to be the counterpart. The only issue I found that it was called twice.

eclipse/src/saros/ui/util/SWTUtils.java
antej95: I pushed a new Commit that contains a fix by keeping track of the display after its first creation.
But it should be seen as a temporary fix that we chose because of time constraints.
The better solution would probably be to change all affected classes since there doesnt seem to be an equivalent to  PlatformUI.getWorkbench().getDisplay() in E4.

srossbach: Looking at the issue, well I do not know. I could modify the affected class. Regarding your solution and looking at the current call stack I guess it would be sufficient just to call getCurrent (assuming the Display was already created and Eclipse call us on the creator thread). Of course your solution works (if the display was created, even if there is no workbench) but it requires good knowledge on SWT / Eclipse Startup.

srossbach:
Are there any plans by you to work further on the whole migration ? I do not really feel very comfortable pushing a partial migration patch into the current master ?

PR #1125, opened 2021-02-23
[FIX][I] Improve new directory input validation

tobous:
Improves the logic validating whether the entered new directory name is
valid for incoming resource negotiations. Specifically disallows names
that contain separator or path separator characters, references to the
current or parent directory, or that already exist in the chosen base
directory.


intellij/src/saros/intellij/ui/wizards/pages/referencepointselection/ReferencePointTab.java
```diff
@@ -580,21 +579,24 @@ private void updateInputValidity() {
    */
   // TODO check whether resource with same name already exists in chosen base directory
   private boolean hasValidNewDirectoryName() {
-    String enteredName = newDirectoryNameTextField.getText();
-    if (enteredName.isEmpty()) {
+    String enteredName = newDirectoryNameTextField.getText().trim();
+
+    if (enteredName.isEmpty()
+        || enteredName.contains(File.separator)
+        || enteredName.contains(File.pathSeparator)
```
srossbach, line 586: Have you tested what will happen on Windows systems if you use `/` ?

tobous: I knew there was a reason why the initial solution I had stashed away used the path utility to check the name count. Yes, Windows will also accept '/' as separator character, even though it is not specified as the Java File.separator character.

I will modify the logic. I wanted to explicitly check for the separator to also disallow including it in the name without adding a new path segment (e.g. giving `foo/` instead of `foo`). But I guess it doesn't matter that much.

tobous: Done. Moved to using the Path API to check for the name count and whether the path is absolute.

intellij/src/saros/intellij/ui/wizards/pages/referencepointselection/ReferencePointTab.java
```diff
@@ -578,23 +578,30 @@ private void updateInputValidity() {
    * TODO check for other separators? Intellij also sees '\' as a separator on unix but this is not
    *  detected by the Java Unix path implementation
    */
-  // TODO check whether resource with same name already exists in chosen base directory
   private boolean hasValidNewDirectoryName() {
-    String enteredName = newDirectoryNameTextField.getText();
-    if (enteredName.isEmpty()) {
+    String enteredName = newDirectoryNameTextField.getText().trim();
+
+    if (enteredName.isEmpty() || enteredName.contains(File.pathSeparator)) {
```
srossbach, line 584: Isn't this check also done in the following try/catch block ?

srossbach (approved):

intellij/src/saros/intellij/ui/wizards/pages/referencepointselection/ReferencePointTab.java
tobous: The `isEmpty` check is technically unnecessary as it is also covered by the path check. I just included it to exit as early as possible, skipping the construction of the `Path` object.

The check whether the name contains a path separator (`;` for Windows or `:` for Unix) is not contained in the try/catch block.

PR #1126, opened 2021-02-23
[FIX][I] #962 Adjust Saros annotation highlighter layer

### LINKED ISSUE #962: Saros annotations overshadow everything, including local user selection
Currently, Saros/I uses [`HighlighterLayer.LAST`](https://github.com/JetBrains/intellij-community/blob/master/platform/editor-ui-api/src/com/intellij/openapi/editor/markup/HighlighterLayer.java#L38) for all its range highlighters. In IntelliJ, the highlighters on the higher layers overwrite/overshadow highlighters on the lower layers. `HighlighterLayer.LAST` is the highest predefined layer and is used by IntelliJ for the local user selection. As a result, the selection annotations overshadow the local user selection. This seems like poor usability to me as the local user should still take precedence over the displayed awareness information.

Furthermore, due to selection and contribution annotations being on the same layer, the darker colors of the contribution annotations seem to overwrite the lighter colors of the selection annotations. This is the opposite behavior to Saros/E, where selection annotations overwrite contribution annotations.

#### Solution approach

Lower the layer of Saros annotations. Where exactly it goes still has to be discussed (i.e. what other highlights Saros annotations should overwrite). A list of all levels used by IntelliJ is given in [HighlighterLayer](https://github.com/JetBrains/intellij-community/blob/master/platform/editor-ui-api/src/com/intellij/openapi/editor/markup/HighlighterLayer.java#L38).

Independently of which layer is chosen, the Selection annotations should be on a layer above contribution annotations to match the behavior of Saros/E.

tobous:
Adjusts the Saros editor annotation highlighter layer. The annotations
are now displayed at a very low level, meaning they will be overshadowed
by almost all other IDE highlighters. This will ensure that the user
will still be able to develop normally using the default IDE features
while in a Saros session.

Separates the different annotations onto different highlighter layers to
create the correct hierarchy. This ensures that selection annotations
overshadow contribution annotations to match the behavior of Saros/E.

It looks a bit weird that the line selection highlighter now also
overshadows Saros annotations. But it provides an important visual aid
to keep track of the current caret position, so I wanted to avoid
overshadowing it. But this is something that can easily be fixed if user
feedback suggests something different.

Resolves #962.

### DIFF
--- a/intellij/src/saros/intellij/editor/annotations/AnnotationHighlighterLayers.java
+++ b/intellij/src/saros/intellij/editor/annotations/AnnotationHighlighterLayers.java
@@ -1,0 +1,45 @@
+package saros.intellij.editor.annotations;
+
+import com.intellij.openapi.editor.markup.HighlighterLayer;
+
+/**
+ * Constants defining the different highlighter layers used by Saros editor annotations.
+ *
+ * <p>Annotations on a layer with a higher number overshadow annotations on a layer with a lower
+ * number.
+ *
+ * @see HighlighterLayer
+ */
+class AnnotationHighlighterLayers {
+
+  private AnnotationHighlighterLayers() {}
+
+  /**
+   * Base highlighter layer used for Saros editor annotations. All Saros annotations must be
+   * displayed on or below this highlighter layer.
+   *
+   * @see HighlighterLayer
+   */
+  private static final int SAROS_BASE_HIGHLIGHTER_LAYER = HighlighterLayer.CARET_ROW - 1;
+
+  /**
+   * Highlighter layer for the caret annotations displayed as part of the selection annotations.
+   *
+   * @see SelectionAnnotation
+   */
+  static final int CARET_HIGHLIGHTER_LAYER = SAROS_BASE_HIGHLIGHTER_LAYER - 1;
+
+  /**
+   * Highlighter layer for the selection annotations.
+   *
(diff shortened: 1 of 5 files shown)


intellij/src/saros/intellij/editor/annotations/AnnotationHighlighterLayers.java
```diff
@@ -0,0 +1,45 @@
...
+   * @see HighlighterLayer
+   */
+  private static int SAROS_BASE_HIGHLIGHTER_LAYER = HighlighterLayer.CARET_ROW - 1;
+
+  /**
+   * Highlighter layer for the caret annotations displayed as part of the selection annotations.
+   *
+   * @see SelectionAnnotation
+   */
+  static int CARET_HIGHLIGHTER_LAYER = SAROS_BASE_HIGHLIGHTER_LAYER - 1;
```
srossbach, line 30: Is there a reason to not make the fields (and class) final ?

tobous: No, it's supposed to be final. I forgot. Will adjust it right away.

tobous: Done.

srossbach (approved):

PR #1127, opened 2021-02-25
Prune annotation store

### LINKED ISSUE #873: Zero-width annotations of removed text are not always removed from the model
In certain cases, annotations of text that was removed are not marked as disposed by IntelliJ. As a result, the editor can contain many invisible zero-width annotations.

This only really becomes an issue when trying to also display annotations in the error stripe (right side by the scroll bar), as the zero-width annotations are still shown there.

This issue seems to be caused by the annotation manager expecting IntelliJ to mark such zero-width annotations as disposed. However, this does not always seem to be the case. From experimenting, it seems like IntelliJ only removes zero-width range highlighters if they were created as part of a deletion of a selected text. If, however, the text is only deleted using backspace, the range highlighter is not correctly disposed.

The issue can be solved by manually pruning the annotation store for the modified file every time we detect a local or remote deletion activity.

tobous:
Adjusts the annotations handling to always prune annotations that have a zero width after a text deletion. This is necessary as such annotations are not always invalidated by IntelliJ IDEA automatically (see #873).

Also adjusts the annotation store to always dispose/tear down any highlighters still present when removing an annotation. This ensures that existing range highlighters will always be completely removed from their editor when an annotation is removed from the annotation store.

I would suggest reviewing this PR commit by commit. The different commit messages give more details on their specific contents.

### Commits

<details><summary><b>[FIX][I] Ensure removed annotations are always correctly torn down</b></summary>
<br>

Adjusts the annotation logic to always properly dispose all range
highlighters when an annotation is removed from the annotation store.
Previously, it was the responsibility of the caller to do so. Including
the teardown in the removal ensures that all range highlighter are
properly removed.

Removes the manual tear down calls now no longer necessary.

Adjusts AnnotationManagerTest for removing invalidated annotations to
also mock the removal call for the caret annotation. This is now
necessary as the new logic ensures that the caret annotation is
correctly removed when the rest of the selection annotation becomes
invalid.

</details>

<details><summary><b>[FIX][I] #873 Prune annotation store after text removal</b></summary>
<br>

Adjusts the logic to always prune the annotation store when text was
removed from a document that is open in an editor. This is necessary to
remove zero-width annotations not automatically invalidated by IntelliJ
IDEA. This is the case when an annotation is removed character by
character instead of removing the entire range at once. In such cases
the annotation is preserved instead of being invalidated.

Such an annotation is not invalidated by IntelliJ IDEA as it could be
configured to be "greedy", meaning it would extend to any new additions
made at its edges. This would allow a zero-width annotation to re-grow
to a visible size.

Resolves #873.

</details>

<details><summary><b>[REFACTOR][I] Rename document modification handlers</b></summary>
<br>

Renames the document modification handler implementations to better
match their purposes.

Also adjusts the documentation of the
LocalDocumentModificationAnnotationUpdater to match its new purpose.

</details>


intellij/src/saros/intellij/editor/annotations/AbstractEditorAnnotation.java
```diff
@@ -15,6 +15,7 @@
 import org.jetbrains.annotations.Nullable;
 import saros.filesystem.IFile;
 import saros.intellij.runtime.EDTExecutor;
+import saros.repackaged.picocontainer.Disposable;
```
srossbach, line 18: Is this really the Disposable interface you want to use ?

tobous: I know what you mean. This isn't really the normal use case of the `Disposal` interface as it is never registered with a context.

I first wasn't sure whether the abstract parent class would have an implementation for the method, so I added the interface. It is probably better just to create a `protected` disposal method in the parent class and remove the interface as it is not really necessary.

tobous: Done.

srossbach (approved):

intellij/src/saros/intellij/editor/annotations/AbstractEditorAnnotation.java
```diff
@@ -205,9 +206,34 @@ void removeLocalRepresentation() {
...
+            return false;
+          }
+
+          if (!rangeHighlighter.isValid()) {
+            return true;
+          }
+
+          if (rangeHighlighter.getStartOffset() == rangeHighlighter.getEndOffset()) {
+            if (editor != null) {
+              removeRangeHighlighter(editor, rangeHighlighter);
```
srossbach, line 222: Maybe a comment why you use `removeIf` even the lambda has side-effects. I guess you want the O(n) performance instead of O(n²).

PR #1130, opened 2021-02-26
Fix Saros view update behavior

### LINKED ISSUE #888: Session and Contacts tab is collapsed when non-host participant leaves session
The session and contacts tab in the Saros view is collapsed when a non-host participant leaves the session.

#### Steps to Recreate
1. Alice (host) and Bob start a session
2. Alice has the session and contacts tab extended
3. Bob leaves the session

##### Expected Behavior
The session and contacts view remains open for Alice.

##### Actual Behavior
The session and contacts view is collapsed for Alice.

tobous:
#### [FIX][I] #888 Fix Saros view tree model update behavior

Introduces new methods to reload the entire tree model or a specific
node while trying to retain the current tree expansion. To do so, the
model saves all expanded paths before the reload and restores them after
the reload.

Replaces all calls to reloading the model with calls to the new utility
methods to ensure that the expansion state is preserved as part of the
reload.

Moves the responsibility of reloading the contacts node for session
related events (e.g. when hiding a contact from the contacts list when
they join the current session) to the contacts node class.

Replaces calls to expanding specific rows of the tree with calls to
expanding the path of the element that is actually supposed to be
expanded. This removes the guesswork whether the given row actually
points to the correct element and ensures that the logic still works
when the tree layout is changed.

Resolves #888.

#### [UI][I] Expand contacts node when first contact is added

Adds new logic explicitly expanding the contacts view if the first
contact is added. This can be the case when there were no contacts
before or when all contacts were part of the session the user just
left. This ensures that the contact node is expanded in such cases.

### DIFF
--- a/intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
+++ b/intellij/src/saros/intellij/ui/tree/ContactTreeRootNode.java
@@ -97,12 +97,18 @@ public class ContactTreeRootNode extends DefaultMutableTreeNode implements Dispo
     ContactInfo contactInfo = getContact(contact);
     if (contactInfo != null) return;
 
+    boolean noContactsToDisplayBefore = visibleContactsMap.isEmpty();
+
     contactInfo = new ContactInfo(contact);
     visibleContactsMap.put(contact, contactInfo);
 
     add(new DefaultMutableTreeNode(contactInfo));
 
     treeView.reloadModelNode(this);
+
+    if (noContactsToDisplayBefore) {
+      treeView.expandPath(new TreePath(getPath()));
+    }
   }
 
   private ContactInfo getContact(XMPPContact contact) {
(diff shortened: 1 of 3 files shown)


tobous:
The entire Saros/I Saros view implementation could use an overhaul, but I did not feel like rewriting at this point in time. So this is just a fix for the apparent issues.

srossbach:
Isnt there a better way to listen to model changes in the TreeViewer ?

tobous:
> Isnt there a better way to listen to model changes in the TreeViewer ?

As already mentioned, I see this more as a quick fix since I did not want to re-write the logic at this point in time (even tough it would need it). I adjusted the logic to now properly use the mechanisms that were already being used for the updating beforehand.

I don't have any real prior experience working with JTrees, so I can't comment on proper change handling. This way works and it should preserve any user changes made to the tree expansion state. Also I would like to mention that while the variable is named `treeView`, it has nothing to do with the Eclipse-specific `TreeViewer` class. This just deals with `JTree`, `DefaultTreeModel`, and `DefaultTreeNode` objects.

From the javadoc [`reload()`](https://docs.oracle.com/javase/7/docs/api/javax/swing/tree/DefaultTreeModel.html#reload()) should be the proper method to call after any nodes in the tree were changed. Also, we aren't really interested in changes to the model as it should not be able to be changed by any other source. Rather, we are interested in changes to the session and XMPP contacts state, according to which we then update the tree.

srossbach (approved):

PR #1135, opened 2021-03-06
[UI][E] UI Migration from E3 to E4

antej95:
This is the continuation of PR [1119](https://github.com/saros-project/saros/pull/1119).
We continued our work with the migration of the static popup menu and began with migrating the view.
Unfortunately we couldnt finish the migration in the scope of the university course.
We tried to get as close as possible to the previous implementation, but E4 doesnt offer the same capabilities as E3, especially regarding the use of Icons in the Menus. We were unable to change the look of the icons during runtime and had to resort to saving the previously dynamically created Icons.
This was not a realistic solution for all icons, which makes the changes incomplete.
Also the changes to the E4 View broke the UI STF Tests and in some places the cosmetics are still wrong.

The Idea behing this PR isnt to be merged but more to serve as a place of discussion about if we chose the right approach, where it can be improved and how it should be finished.

There are several areas that need to be worked on before the migration is complete. The most notable ones are:
- Finishing the View (Icons)
- Fix the STF Tests
- Remove the inheritance of Saros from AbstractUIPlugin
- Remove all dependencies on the package eclipse.org.ui and all its subpackages
- Wizards

In the future maybe this work can be built upon to finish the migration. 
There is a possibility that someone from our course is going to continue working on the migration. Whether it will build upon this work or not is not yet decided though.


eclipse/src/saros/ui/handlers/toolbar/ConsistencyHandler.java
```diff
@@ -368,32 +378,10 @@ public void run() {
               newDirection = FADE_UP;
             }
 
-            setImageDescriptor(new MemoryImageDescriptor(modifyAlphaChannel(OUT_SYNC, newValue)));
+            // TODO: create modified icon
```
srossbach, line 381: Nope, was already done. Fix it!

srossbach (changes requested):
This is not reviewable. The last commit used contains modified C&P and doing this in one step destroys every diff.

eclipse/src/saros/ui/handlers/toolbar/ConsistencyHandler.java
srossbach: Have not tested it but I guess it could work:

```
consistencyToolItem.setIconURI(null) // surpress the default renderer
((ToolItem)consistencyToolItem.getWidget()).setImage(....)
```

I do not know if the setImage call will make a copy of the image, otherwise you have to use 2 images and flip between them.

eclipse/src/saros/ui/util/SWTUtils.java
```diff
@@ -221,18 +221,6 @@ public static Shell getShell() {
...
-
-    if (activeWorkbenchWindow != null && !activeWorkbenchWindow.getShell().isDisposed())
-      return activeWorkbenchWindow.getShell();
-
-    final IWorkbenchWindow[] workbenchWindows = workbench.getWorkbenchWindows();
-
-    if (workbenchWindows.length > 0 && !workbenchWindows[0].getShell().isDisposed())
-      return workbenchWindows[0].getShell();
-
     return display.getActiveShell();
```
srossbach, line 224: Have you read the comment on line 213 (Hint: getActiveShell() may return null if the application is minimized which results in dialogs etc may pop up even if Eclipse is minimized).

Issue #1138, opened 2021-03-22
Build fails caused by missing transitive dep


FKHals:
A transitive eclipse dependency was changed which causes the build to fail (see [last JDK11 build action](https://github.com/saros-project/saros/actions/runs/675343408), but this also happens with Java 8 on my local machine).

The bundle [org.eclipse.ui.ide](https://mvnrepository.com/artifact/org.eclipse.platform/org.eclipse.ui.ide/3.17.200) requires the bundle [org.eclipse.e4.core.di](https://mvnrepository.com/artifact/org.eclipse.platform/org.eclipse.e4.core.di/1.7.700) which in turn has the actual problematic requirement `javax.annotation:javax.annotation-api:[1.3.5,2.0.0)` which does not seem to exist in maven central.
The newest available version of [javax.annotation-api](https://mvnrepository.com/artifact/javax.annotation/javax.annotation-api) would be [1.3.2](https://mvnrepository.com/artifact/javax.annotation/javax.annotation-api/1.3.2).

Edit: I just confirmed locally that a possible temporary workaround would be to use the [previous version of org.eclipse.e4.core.di](https://mvnrepository.com/artifact/org.eclipse.platform/org.eclipse.e4.core.di/1.7.600).

Edit: The reason seems to be that javax.annotation:javax.annotation-api is deprecated and has been replaced by jakarta.annotation-api which should have an identical API and also fits the version boundaries [1.3.5, 2.0.0) given by org.eclipse.e4.core.di. So it seems like the author may have overseen that also the package name should be changed but that is just an assumption.

Issue #1086, opened 2020-09-17
Build fails on master


m273d15:
A transitive eclipse dependency was changed.
The bundle [`org.eclipse.urischeme`](https://mvnrepository.com/artifact/org.eclipse.platform/org.eclipse.urischeme/1.1.100) requires the bundles `com.sun.jna:com.sun.jna:[4.5.1,6.0.0)` `com.sun.jna:com.sun.jna:[4.5.1,6.0.0)`. These bundles are not available in maven central (as other eclipse platform jars/bundles).

m273d15:
A solution could be using the [AsMavenPlugin](https://javadoc.io/doc/com.diffplug.gradle/goomph/latest/com/diffplug/gradle/p2/AsMavenPlugin.html) of the goomph plugin suite and convert the jna's p2 repository to maven and use the artifact.

m273d15:
We are currently using a [workaround](https://github.com/saros-project/saros/commit/ef46512e98211fca46e458f0a6f6806a6716a900#diff-58148c13ba0fce105864f07f8ea42b6d203926708aa85a8557de20187fb13365) to bypass the issue.

PR #1139, opened 2021-03-22
[BUILD] Workaround for build fail

FKHals:
This is a temporary fix for #1138.

This is not a proposed solution and the workaround will most probably need to be updated for coming versions of `org.eclipse.platform:org.eclipse.ui.ide` (the version of which is also part of another workaround, see [here](https://github.com/saros-project/saros/pull/1115)). Actually discussion about long-term fixes should be done in #1138.

### DIFF
--- a/eclipse/build.gradle.kts
+++ b/eclipse/build.gradle.kts
@@ -42,6 +42,8 @@ dependencies {
     implementation(project(":saros.core"))
     // This is a workaround for https://github.com/saros-project/saros/issues/1086
     implementation("org.eclipse.platform:org.eclipse.urischeme:1.1.0")
+    // This is a workaround for https://github.com/saros-project/saros/issues/1138
+    implementation("org.eclipse.platform:org.eclipse.e4.core.di:1.7.600")
     // This is a workaround for https://github.com/saros-project/saros/issues/1114
     implementation("org.eclipse.platform:org.eclipse.ui.ide:3.17.200")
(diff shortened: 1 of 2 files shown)


Drakulix:
So the same as in #1114 and #1086 again?

FKHals:
> So the same as in #1114 and #1086 again?

Yes, pretty much. The reason seems to be that `javax.annotation:javax.annotation-api` is deprecated and has been replaced by [jakarta.annotation-api](https://mvnrepository.com/artifact/jakarta.annotation/jakarta.annotation-api) which should have an identical API and also fits the version boundaries `[1.3.5, 2.0.0)` given by [org.eclipse.e4.core.di](https://mvnrepository.com/artifact/org.eclipse.platform/org.eclipse.e4.core.di/1.7.700).

Drakulix (dismissed):

FKHals:
Rebased on the master.

Unfortunately that requires the approval to be renewed (Is there a better solution to do rebases without requiring approval for the same changes?)

Drakulix:
> Rebased on the master.
> 
> Unfortunately that requires the approval to be renewed (Is there a better solution to do rebases without requiring approval for the same changes?)

You can use the "Merge"-Button from github, that keeps Reviews, but clutters the commit history.
But for simple one-commit changes like this one, that does not matter, because we can easily squash it all on merge.

Drakulix (approved):

Issue #1121, opened 2021-02-12
[Server] Project sharing, file sharing and chat do not work on master


jrtberlin:
**Describe the bug**

When Alice and Bob join Dave's Server, Alice and Bob can not chat, share projects or files with each other.

**To Reproduce**

1. Compile the server with `./gradlew cleanAll sarosServer`
2.  `java -Dsaros.server.interactive=yes -Dsaros.server.jid=dave@examplexmppserver.de -Dsaros.server.password=passwd -jar saros.server.jar`
3. `invite alice@examplexmppserver.de` & `invite bob@examplexmppserver.de`
4. try to share a project in Eclipse, chat with a participant or send a file
5. nothing happens and the server does not log any errors.

**Expected behavior**
Alice and Bob can chat and collaborate on code.

**Log files**

* Eclipse:
  * IDE logs - `<workspace>/.metadata/.log`
        There are none
  * Saros logs - `<workspace>/.metadata/.plugins/saros.eclipse/log/<date>/*.log`

[Saros_2021-02-12_17-03-51.log](https://github.com/saros-project/saros/files/5972967/Saros_multitest%40saros-con.imp.fu-berlin.de_2021-02-12_17-03-51.log)

**Environment (please complete the following information):**
 - OS: Arch Linux x86_64, Kernel: 5.10.13-arch1-1
 - IDE: Eclipse 2020-12
 - Saros Version: 16.0.1
 - Server Version: master

Drakulix:
Cannot confirm. After a debugging session with @jrtberlin this look like this issue happens, if the server.jid and any client.jid have no added themselves to their contact list.

As the server accepts any subscription request automatically, adding the server via gui is not a problem. The reverse in the server-cli however is not true.

The invite command should likely first add any previously unknown contacts.

srossbach:
Looking at the log file there seems to be configuration errors on the server side. The MUC channel is missing for some reason and so there will not be any session chat. The next strange issue is that there is no transport method defined.

FelixSelter:
@Drakulix @jrtberlin Would you like to see this implemented as another command or do you want it to work with the default invite command?

Drakulix:
> @Drakulix @jrtberlin Would you like to see this implemented as another command or do you want it to work with the default invite command?
> 
> I was trying to get started with contribution and thought this is an easy task. I could send a pull request just tell me which way you preferr

Hey thanks for reaching out.
[...]

PR #1140, opened 2021-04-04
[FIX] #1121 [S] New contact will be added through invite command if necessary

FelixSelter:

Please, check that:
- [x] your code complies with the google java format (see [here](https://www.saros-project.org/contribute/development-environment.html) for details)

### Description
In this Pull request issue, #1121 is fixed.
If someone is invited by the `invite` command of the Server it will check if this person is an actual contact and if not add him as one.

Also its my first time contributing so I hope everything is fine. Else tell me. I'm a human and got the ability learn

Thank you for the awesome work and to the reviewers!

### DIFF
--- a/server/src/saros/server/console/InviteCommand.java
+++ b/server/src/saros/server/console/InviteCommand.java
@@ -5,9 +5,14 @@ import static java.util.stream.Collectors.partitioningBy;
 import java.io.PrintStream;
 import java.util.List;
 import java.util.Map;
+import java.util.function.BiPredicate;
 import org.apache.log4j.Logger;
+import saros.exceptions.OperationCanceledException;
 import saros.net.util.XMPPUtils;
 import saros.net.xmpp.JID;
+import saros.net.xmpp.contact.XMPPContact;
+import saros.net.xmpp.contact.XMPPContactsService;
+import saros.repackaged.picocontainer.annotations.Inject;
 import saros.session.ISarosSessionManager;
 
 public class InviteCommand extends ConsoleCommand {
@@ -18,6 +23,8 @@ public class InviteCommand extends ConsoleCommand {
     this.sessionManager = sessionManager;
     console.registerCommand(this);
   }
+
+  @Inject private XMPPContactsService contactsService;
 
   @Override
   public String identifier() {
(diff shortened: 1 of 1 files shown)


srossbach:
I still have issues figuring out what is going wrong ? Neither the commit description or the the description of the issue mention what is going wrong. And before try to respond I do not accept the answer: You have to add the contact. I want to know what part of Saros is not working as expected and WHY.

srossbach:
Ok after digging through the code, Stefan changed how Saros support is determined. This now requires presence information which you will not obtain as long as there is no subscription status. So the minimum requirement is `subscription='to':` .

This does not fix anything unless the user somehow (depending on Saros/E or Saros/I) accepts the invitation. To be more specific, the user has to accept the invitation ASAP which it simply cannot.

FelixSelter:
As far as I understood the issue is that, users are confused if they try to invite someone but its not working because they are not in the contact list of the server. My approach to fix this was to send an contact request in this case. It would may make mor sense if I would print the warning and then return so they wont be covered by the error traces. The user then has time to accept and the admin will reinvite him. I also suggested to add this contact invitation as an external command but this wont fix the issue.

To my understanding the issue is fixed because the admin get notified and knows that the user has to accept the contact invite.

I will try to take a look into this subscription system but as I started im not really familiar with saros.

Any other suggestions? What would fix the issue in your opinion. Is it possible to override the rights of the other user so that theres no need to accept the contact invite (I dont think that should be done but seems to be the only other solution)?

FelixSelter:
Example session till I fix the issue with the subscription:
```
# Welcome to Saros Server (type 'help' for available commands)
> invite name@saros-con.imp.fu-berlin.de
Adding name@saros-con.imp.fu-berlin.de as new contact. You will have to reinvite him after he accepted being added to the contacts
10:47:17.751 [main] ERROR saros.server.console.InviteCommand - Unknown ErrorAn unknown error has occured:

subscription-required(407) Not subscribed

It will be continued anyway
The contact was added successfully. Please invite him again after he accepted
> invite name@saros-con.imp.fu-berlin.de
> [-] Starting session negotiation...
[\] Checking Saros support...
[|] Checking version compatibility...
[/] Sending negotiation request...
[-] Waiting for negotiation acknowledgement...
[\] Waiting for name@saros-con.imp.fu-berlin.de to accept invitation...
[|] Waiting for remote session configuration data...
[/] Sending final session configuration data...
[-] Waiting for name@saros-con.imp.fu-berlin.de to connect...
[\] Establishing connection and performing final initialization...
[|] Synchronizing user list...
```

server/src/saros/server/console/InviteCommand.java
```diff
@@ -44,9 +51,58 @@ public void execute(List<String> args, PrintStream out) {
         log.warn("Invalid JID skipped: " + jid);
       }
 
-      sessionManager.invite(jids.get(true), "Invitation by server command");
       for (JID jid : jids.get(true)) {
-        sessionManager.startSharingReferencePoints(jid);
+
+        // --------------add contact if not already--------------
```
srossbach, line 56: The comments are pretty pointless unless you are not able to read code at all.

server/src/saros/server/console/InviteCommand.java
```diff
@@ -44,9 +51,58 @@ public void execute(List<String> args, PrintStream out) {
...
+
+        if (!contactFound) { // add contact
+          log.warn(
+              "Adding "
+                  + jid
+                  + " as new contact. You will have to reinvite him after he accepted being added to the contacts");
+          out.println(
+              "Adding "
+                  + jid
+                  + " as new contact. You will have to reinvite him after he accepted being added to the contacts"); // the previous warning wont be displayed in the console
```
srossbach, line 74: Do you really want to show these messages in the log file ? It is an interactive command at all.

server/src/saros/server/console/InviteCommand.java
```diff
@@ -44,9 +51,58 @@ public void execute(List<String> args, PrintStream out) {
...
+
+          final BiPredicate<String, String> questionDialogHandler = // error handler
```
srossbach, line 76: It is an automated process, either you are query the STDIN or simply return true, As a side node, this will executed mostly all the time because of privacy reasons, i.e the server will not tell you if such a JID exists.

server/src/saros/server/console/InviteCommand.java
```diff
@@ -44,9 +51,58 @@ public void execute(List<String> args, PrintStream out) {
...
+
+          try {
+            contactsService.addContact(jid, null, questionDialogHandler);
+            log.info(
+                "The contact was added successfully. Please invite him again after he accepted");
+            out.println(
+                "The contact was added successfully. Please invite him again after he accepted");
+            return; // exit because the user cant accept the invite in time
+          } catch (OperationCanceledException e) {
+            log.error(e);
```
srossbach, line 97: Is this your intention to continue with the code on line 102 ? I do not think so.

server/src/saros/server/console/InviteCommand.java
```diff
@@ -44,9 +51,58 @@ public void execute(List<String> args, PrintStream out) {
...
+          }
+        }
+        // --------------end add contact if not already--------------
+
+        sessionManager.invite(
```
srossbach, line 102: If you do not see the contact online (I do not know if XMPP supports invisibility, but that is another issue) then it makes no sense to continue. The invitation will simply fail.

